PR #43, opened 2023-06-12
Hotfix: Update Readme [ SpellingError ]

MANISH007700:
Instead of `When it comes to RAG pipelines`, it was written `When is comes to RAG pipelines`. 
Rectified ✅


shahules786:
Thank you @MANISH007700 🤗

PR #59, opened 2023-07-12
Context Relevancy v2

shahules786:
## What
Added support for new context relevancy metrics
## Why
The earlier used context_relevancy metric was found to have a poor correlation with actual results. 
## How
New paradigm leverages candidate sentence extraction with self-consistency checks to ensure far better results even with larger context lengths. 
## Screenshots 
Screenshot 2023-07-13 at 10 47 52 AM

### DIFF
--- a/src/ragas/metrics/context_relevance.py
+++ b/src/ragas/metrics/context_relevance.py
@@ -7,7 +7,6 @@ from typing import List
 
 import numpy as np
 from datasets import Dataset
-from pandas import DataFrame
 from sentence_transformers import CrossEncoder
 
 from ragas.metrics.base import Metric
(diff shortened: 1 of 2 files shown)


src/ragas/metrics/context_relevance.py
```diff
@@ -2,62 +2,137 @@
...
 
     def score(self: t.Self, dataset: Dataset) -> Dataset:
         """
         dataset: Dataset[question: list[str], contexts: list[list[str]]]
         """
+        prompts = []
+        questions, contexts = dataset["question"], dataset["contexts"]
+        for q, c in zip(questions, contexts):
+            prompt = CONTEXT_RELEVANCE.format(q, "\n".join(c))
+            prompts.append(prompt)
```
jjmachan, line 104: wouldn't this need batching. If there are more than 20 (questions, contexts) pairs openAI will return an error?

src/ragas/metrics/context_relevance.py
```diff
@@ -2,62 +2,137 @@
...
+                ]
+                overlap_scores.append(len(indices) / len(context_sents))
+            if self.strictness > 1:
+                agr_score = self.sent_agreement.evaluate(n_output)
+            else:
+                agr_score = 1
+            scores.append(agr_score * np.mean(overlap_scores))
 
-        return score_ds
+        return dataset.add_column(f"{self.name}", scores)  # type: ignore
```
jjmachan, line 135: why are we type ignoring here? just curious

src/ragas/metrics/context_relevance.py
```diff
@@ -2,62 +2,137 @@
...
 
 
 @dataclass
 class ContextRelevancy(Metric):
-    batch_size: int = 32
     name: str = "context_relavency"
-    model_name: str = "t5-base"
+    batch_size: int = 15
+    agreement_metric: str = "bert_score"
+    strictness: int = 2
```
jjmachan, line 88: what does the strictness param do? would be great to have doc string at this point I guess

src/ragas/metrics/context_relevance.py
shahules786: A warning for not passing the dataset fingerprint argument, The same is done in the Faithfulness class.

shahules786:
@jjmachan  Added the required changes.

jjmachan (approved):

PR #91, opened 2023-08-10
feat: column mapping to select columns

jjmachan:


jjmachan:
todos
- [ ] update docs string for evaluate

PR #94, opened 2023-08-16
feat: add support for langchain

jjmachan:


jjmachan:
things todo before merge
- [ ] write doc strings
- [ ] make sure the callbacks are used correctly
- [ ] finish out the notebook

todos before blog
- [ ] langsmith cookbook notebook
- [ ] the verbose version of that will be the blog

docs/integrations/langchain.ipynb
hinthornw, line 1: Looks good! Could use a bit of cleanup to rm the image path strings at the end and the %autoreload

And then could you add
```
import nest_asyncio
nest_asyncio.apply()
```

At the top of the notebook?

Some versions of the run_on_dataset function currently parallelize via asyncio, which can cause issues when running in jupyter since an event loop is already running.

PR #96, opened 2023-08-23
Context Recall

shahules786:
## What
Context recall estimation using annotated answers as ground truth

## Why
Context recall was a highly requested feature, as it is one of the main pain points where pipeline error occurs in RAG systems

## How
Introduced a simple paradigm similar to faithfulness

### DIFF
--- a/src/ragas/metrics/context_recall.py
+++ b/src/ragas/metrics/context_recall.py
@@ -32,6 +32,18 @@ classification:
 
 @dataclass
 class ContextRecall(MetricWithLLM):
+
+    """
+    Estimates context recall by estimating TP and FN using annotated answer and
+    retrieved context.
+
+    Attributes
+    ----------
+    name : str
+    batch_size : int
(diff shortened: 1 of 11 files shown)


src/ragas/metrics/context_recall.py
```diff
@@ -0,0 +1,96 @@
...
+        Batch size for openai completion.
+    """
+
+    name: str = "context_recall"
+    evaluation_mode: EvaluationMode = EvaluationMode.gc
+    batch_size: int = 15
+
+    def __post_init__(self: t.Self):
+        self.prompt_format = CONTEXT_RECALL_RA
+        self.verdict_token = "[Attributed]"
```
jjmachan, line 53: why do we need post init here? you can add it to the main class itself right?

jjmachan (changes requested):
can you also change the quickstart guide?

jjmachan (approved):

PR #124, opened 2023-09-13
Update README with explanation of Ragas metrics

Yongtae723:
Added to the readme which results each metric is calculated from.

Personally, this information can be found by reading deeper, but for those new to ragas, it will make it easier to get started!

If you think it too much, please just close it!

thanks in advance

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,15 +89,15 @@ If you want a more in-depth explanation of core components, check out our [quick
 
 Ragas measures your pipeline's performance against different dimensions
 
-1. **Faithfulness**: measures the information consistency of the generated answer against the given context. If any claims are made in the answer that cannot be deduced from context is penalized.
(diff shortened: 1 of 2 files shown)


jjmachan:
Thank you so much for taking the time to improve the docs for the other users as well, much appreciated @Yongtae723 ❤️

shahules786:
hi @Yongtae723 , thanks for the PR. I have suggested some changes. Two suggestions here
1. Can you also make similar changes to [metrics.md](https://github.com/explodinggradients/ragas/blob/main/docs/metrics.md)
2. It would be good if you can specify the column names using code formatting. for example, "it is calculated from `answer`"

Yongtae723:
OK I did @shahules786

shahules786:
@Yongtae723 Thanks, I can merge once the comment in answer relevancy part  is addressed.

shahules786 (approved):

Yongtae723:
@shahules786 Oh!

thank you for fix and update my PR!
also thank you for merging!

PR #125, opened 2023-09-13
update tutorial notebook

Yongtae723:
original note book contains `# incorrect answer` on eval_answer.

Considering actual operation, eval_answer must always be correct, and it is strange to put a wrong answer in eval_answer!


jjmachan (approved):
Thank you so much for improving the readme and making it much easier to understand @Yongtae723 🙌🏽

PR #139, opened 2023-09-15
Update README.md

emilesilvis:
Include metrics diagram from https://blog.langchain.dev/evaluating-rag-pipelines-with-ragas-langsmith/.


jjmachan:
yep does makes it really more clear, Thank you @emilesilvis for adding this :)

PR #146, opened 2023-09-19
Add note about when ground truth labels are required

tleyden:
To reduce some confusion, add a note mentioning that ground truth labels are only required for context_recall, since the tool can generate other metrics in a "ground truth" free manner.


jjmachan:
Thank you so much @tleyden for taking the time to help with the docs. Much appreciated 🙂

Issue #136, opened 2023-09-14
Testset Generation: bringing continual learning to RAG pipelines


jjmachan:
We started ragas with `ground-truth`free evaluations so that you didn't have to put significant upfront effort into building an ideal test set before running evaluations. **Creating a test set needs substantial upfront investment in time, money, human hours and expertise to get it right.** It is also a continuous process as your product and ML model evolve to cater to diverse use cases.  We are exploring the possibilities of synthetic test set generation because

1. As RAG users get more mature and go into production, having a solid test set and evaluation strategies becomes critical to give users a seamless experience. This means they have to put more time into building solid test sets and evaluation methodologies.
2. ground-truth free evaluation has its limitations. It is very effective in quantifying aspects like faithfulness but **ground-truth free evaluation cannot be used to ensure aspects like answer correctness which is also very important**. Here a synthetic test set with ground-truth can be of high utility.

The whole focus of the Ragas library is to help you build more reliable RAG applications which is why with the next leg of Ragas we'll be focusing a lot more on test set generation and continual learning of RAG pipelines. The goal is to leverage custom LLMs and Data-Centric AI techniques to

1. Build more robust paradigms for test set generation. 
    1. Many libraries already have some sort of test set generation but they have a few shortcomings. Ideally, the test set should have a good distribution of easy -> hard questions across different tasks/situations as seen in production.
2. Tools to scale up and reduce the cost of test set generation. 
    1. Works like self-instruct and evol-instruct have proven that LLMs can generate human-quality synthetic data. **We are working on paradigms to generate high-quality synthetic data generation specific to RAG**. Ref [[1]](https://arxiv.org/abs/2306.08568) [[2]](https://arxiv.org/abs/2212.10560)
3. Methodologies to continuously add to and improve the test set as your RAG pipelines evolve using other data points like logs and feedback.

there is a lot of work to be done but with the v0.1 release of Ragas, we'll be releasing features in this direction. In the meantime, we would love to hear your opinions, expectations, suggestions and ideas about this too :)

Team Ragas

PR #150, opened 2023-09-21
Test set generation paradigm

shahules786:
#136 
TODO

- [ ] Add README with flowchart demonstration of paradigm 
- [x] Add quickstart notebook
- [x]  Fix type hinting 
- [x] Reformat methods if needed
- [x] Add progress bar
- [ ] Async workflow

### DIFF
--- a/src/ragas/testset/testset_generator.py
+++ b/src/ragas/testset/testset_generator.py
@@ -291,7 +291,7 @@ class TestsetGenerator:
             node_idx = self.rng.choice(available_indices, size=1)[0]
             available_indices = self._remove_index(available_indices, [node_idx])
 
-            neighbor_nodes = doc_nodeidx[document_nodes[node_idx].node_id]
+            neighbor_nodes = doc_nodeidx[document_nodes[node_idx].node_id] # type: ignore
 
             # Append multiple nodes randomly to remove chunking bias
             node_indices = (
--- a/src/ragas/testset/testset_generator.py
+++ b/src/ragas/testset/testset_generator.py
@@ -1,220 +1,233 @@
-from attr import dataclass
-from llama_index import OpenAIEmbedding
-from numpy.random import default_rng
-import numpy as np
-import numpy.testing as npt
-
-import pandas as pd
 import typing as t
-from llama_index.readers.schema.base import Document
-from llama_index.node_parser.simple import SimpleNodeParser
-from llama_index.schema import TextNode
-from langchain.chat_models import ChatOpenAI
-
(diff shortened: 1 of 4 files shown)


src/ragas/testset/testset_generator.py
```diff
@@ -0,0 +1,309 @@
...
+        results = generate(prompts=[prompt], llm=self.ctitic_llm)
+        return bool(results.generations[0][0].strip().endswith("Yes."))
+
+    def _reasoning_question(self, question: str, context: str):
+        return self._qc_template(REASONING_QUESTION, question, context)
+
+    def _condition_question(self, question: str, context: str):
+        return self._qc_template(CONDITIONAL_QUESTION, question, context)
+
+    def _multicontext_question(self, question: str, context1: str, context2: str):
```
jjmachan, line 136: return types are not annotated anywhere actually

src/ragas/testset/testset_generator.py
```diff
@@ -0,0 +1,309 @@
...
+                """Maximum possible number of samples exceeded, 
+                             reduce test_size or add more documents"""
+            )
+
+        available_indices = np.arange(0, len(document_nodes)).tolist()
+        doc_nodeidx = self._generate_doc_node_map(document_nodes)
+        count = 0
+        Testdata_tuple = namedtuple(
+            "Testdata_tuple", ["question", "context", "answer", "question_type"]
+        )
```
jjmachan, line 215: we can declare this outside as `TestdataRow` or something so that it can be used as a type

jjmachan (commented):
basically 3 things
- type annotation for return types
- more comments: this would really help understand the code better
- there are some small functions that are only used once. should we remove those and add them in-place instead

jjmachan:
also should we keep adding to this PR or merge this and make smaller PRs?

src/ragas/testset/testset_generator.py
```diff
@@ -0,0 +1,309 @@
...
+
+        return embeddings
+
+    def generate(self, documents: t.List[Document], test_size: int):
+        node_parser = SimpleNodeParser.from_defaults(
+            chunk_size=self.chunk_size, chunk_overlap=0, include_metadata=True
+        )
+        document_nodes = node_parser.get_nodes_from_documents(documents=documents)
+
+        if test_size > len(document_nodes):
```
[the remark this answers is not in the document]
shahules786: For this version, I don't recommend generating more than one `seed question` per node. Moving on I have to find a way to ensure diversity even if the algorithms need to frame more questions from a single node.

jjmachan (approved):

Issue #159, opened 2023-09-28
Answer correctness


shahules786:
Answer correctness given (generated_answer, ground_truth) can be measured are the linear combination of semantic similarity and factual correctness.  
TODO

- [x] Implement Semantic similarity metrics - using cross encoder
- [x] Implement Answer correctness as  _x * semantic similarity + y * factual correctness._
- [x] Verify faithfulness works with answer<->GT

PR #161, opened 2023-09-28
Answer correctness metrics

shahules786:
fixes: #159

### DIFF
--- a/src/ragas/metrics/__init__.py
+++ b/src/ragas/metrics/__init__.py
@@ -1,4 +1,5 @@
 from ragas.metrics.answer_relevance import AnswerRelevancy, answer_relevancy
+from ragas.metrics.answer_similarity import AnswerSimilarity, answer_similarity
 from ragas.metrics.context_precision import (
     ContextPrecision,
     ContextRelevancy,
@@ -16,6 +17,8 @@ __all__ = [
     "faithfulness",
     "AnswerRelevancy",
     "answer_relevancy",
+    "AnswerSimilarity",
+    "answer_similarity",
     "ContextRelevancy",
     "context_precision",
     "ContextPrecision",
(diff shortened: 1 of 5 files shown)


src/ragas/metrics/__init__.py
```diff
@@ -1,4 +1,5 @@
 from ragas.metrics.answer_relevance import AnswerRelevancy, answer_relevancy
+from ragas.metrics.answer_similarity import AnswerSimilarity, answer_similarity
```
jjmachan, line 2: why are you exposing answer similarity? it should be `AnswerCorrectness` right?

src/ragas/metrics/answer_correctness.py
```diff
@@ -0,0 +1,78 @@
...
+            ds_faithfulness = dataset.remove_columns(["contexts"])
+        else:
+            ds_faithfulness = dataset
+
+        ds_faithfulness = ds_faithfulness.rename_columns({"ground_truths": "contexts"})
+        faith_scores = self.faithfulness._score_batch(ds_faithfulness)  # type: ignore
+        similarity_scores = self.answer_similarity._score_batch(dataset)  # type: ignore
+
+        scores = np.vstack([faith_scores, similarity_scores])
+        scores = np.average(
```
jjmachan, line 71: why is average the right call here?

jjmachan (changes requested):
I think you have to expose `AnswerCorrectness` not sure if your already doing it (ignore in that case)

src/ragas/metrics/__init__.py
shahules786: Answer similarity is an independent metric.  It can be used by the user if required.

src/ragas/metrics/answer_correctness.py
shahules786: It's a weighted average where users can pass the coefficients.

jjmachan (approved):

PR #180, opened 2023-10-12
Fix typos

pitmonticone:


shahules786:
Nice catch @pitmonticone. Thank you.

PR #204, opened 2023-10-20
feat: context average precision

shahules786:
- [x] Add AP from [here](https://amitness.com/2020/08/information-retrieval-evaluation/#2-average-precisionap)
- [ ] Add docs
- [x] Rename context precision (depreciate)


src/ragas/metrics/context_precision.py
```diff
@@ -177,9 +186,75 @@ def _score_batch(
...
+        Batch size for openai completion.
+    """
+
+    name: str = "context_precision"
+    evaluation_mode: EvaluationMode = EvaluationMode.qc
+    batch_size: int = 15
+
+    def init_model(self: t.Self):
+        pass
+
```
jjmachan, line 208: this is not needed because the base class for `MetricsWithLLM` implements this now. this will override that implementation

jjmachan (changes requested):

jjmachan (approved):

PR #207, opened 2023-10-23
Refactor testset_generator.py for newbies to understand easily

Yongtae723:
Hi

I totally understand original code is correct.
But I think my suggestion is much easier to understand.

If you feel it is too much, you can ignore


jjmachan:
yep this is easier to understand, thanks for putting the PR :)

PR #216, opened 2023-10-26
docs: add langfuse notebook

jjmachan:
How to use langfuse dashboard with ragas metrics


jjmachan:
this should be be merged on monday evening

shahules786:
Nice one! It would be valuable to include the newly added `context_precision` instead of more than one aspect critique, which I would limit to one.

jjmachan:
ohh okay, I'll update it like that

Issue #208, opened 2023-10-23
Expose prompts 


shahules786:
Since ragas supports OSS LLM that might need some prompt tuning for each metric, it will be a good idea to allow users to edit prompts easily.  Currently, this is not possible except `answer_relevancy` because files and metric names are the same for other metrics.  Renaming files will be an easy solution.

PR #243, opened 2023-10-31
expose metrics prompts by renaming metric modules

grauvictor:
fixes: #208

Add prefix "metric_" for each metric modules. Shouldn't break existing workflow as the last expose component with the module name was the class instance, except for answer_relevancy. 

TODO:
- [x] Check documentation


jjmachan:
thank you for putting the PR but just one change requests
instead of adding `metric_` for each module could you add just an underscore `_`

eg `fiathfullness.py` will become `_faithfullness.py`

grauvictor:
> thank you for putting the PR but just one change requests instead of adding `metric_` for each module could you add just an underscore `_`
> 
> eg `fiathfullness.py` will become `_faithfullness.py`

Yes! Should be ok now

jjmachan (approved):

jjmachan:
yep this is awesome!
thank you so much @grauvictor for putting this in ❤️

PR #244, opened 2023-11-01
Renamed module names and made changes in init to reflect the change

tinomaxthayil:
* Changed module names to _contexts, _context_recall, _faithfullness
* Changed faithfulness module name in answer_correctness.py
* Added the new names to init.py


shahules786:
- file names for `answer_correctness` and other metrics also have to change
- Add unit tests? @jjmachan

jjmachan:
yes check the documentation on how we are importing it and add tests that mimic that.
also not that `critic` is something that is now imported like `ragas.metric.critic` do think about if we should change it

tinomaxthayil:
* Added test cases and made name changes for the rest of the modules.
*  Critique name is not changed but test case is added for the expected import.

src/ragas/metrics/__init__.py
```diff
@@ -1,15 +1,15 @@
...
-from ragas.metrics.context_precision import (
+from ragas.metrics._answer_correctness import AnswerCorrectness, answer_correctness
+from ragas.metrics._answer_relevance import AnswerRelevancy, answer_relevancy
+from ragas.metrics._answer_similarity import AnswerSimilarity, answer_similarity
+from ragas.metrics._contexts import (
     ContextPrecision,
     ContextRelevancy,
     context_precision,
     context_relevancy,
 )
```
jjmachan, line 9: @tinomaxthayil and @shahules786 should we split this into 2 files?

tinomaxthayil: you meant Class and instance for _contexts?

shahules786: @jjmachan Not sure what is ideal here, since both are metrics to evaluate contexts, I would suggest to keep as it is.

shahules786 (approved):

src/ragas/metrics/_contexts.py
shahules786, line 1: This should now be `_context_precision`

shahules786 (changes requested):

shahules786 (approved):

PR #256, opened 2023-11-06
Docs user tests

tinomaxthayil:
* Added hint in synthetic data generation to point to user data page


shahules786 (changes requested):
There are other unnecessary file changes with this PR, could you remove that?

PR #257, opened 2023-11-06
Custom prompts

tinomaxthayil:
* Changed the default prompts into strings and added the HumanPrompt function conversion inside the class
* Added attributes in the class to pass string prompts into the class
* Added test to check if custom prompts are attribute values in the instance


shahules786 (changes requested):
There are unnecessary file changes with this PR, and file conflicts. Please fix those.

PR #261, opened 2023-11-07
feat: add native support for OpenAI and AzureOpenAI

jjmachan:
- Add support for OpenAI and AzureOpenai both embeddings and LLMs
- rework `RagasLLM` with an async version of generate
- checks for API_KEYs and tests that ensure it is working


shahules786 (approved):
LGTM

PR #277, opened 2023-11-10
Custom embeddings from HuggingFace (sentence transformer) do not pass the assertion in src/ragas/metrics/_answer_similarity.py  

alebondarenko:
When passing the embeddings for example in the answer relevancy, sentence transformers do not pass assert isinstance(embeddings, Tensor) in src/ragas/embeddings/base.py, so the solution is to add "convert_to_tensor": True as follows:
 
AnswerRelevancy(embeddings=HuggingfaceEmbeddings(
    model_name='intfloat/multilingual-e5-large', encode_kwargs={"convert_to_tensor": True}))

However, this does not work for the src/ragas/metrics/_answer_similarity.py because encode_kwargs are initialized with self.embeddings.encode_kwargs = {"batch_size": self.batch_size}
my proposed solution is to change line 56 to:
self.embeddings.encode_kwargs = {"batch_size": self.batch_size, "convert_to_tensor": True}

### DIFF
--- a/src/ragas/embeddings/base.py
+++ b/src/ragas/embeddings/base.py
@@ -66,7 +66,7 @@ class HuggingfaceEmbeddings(RagasEmbeddings):
             texts, normalize_embeddings=True, **self.encode_kwargs
         )
 
-        #assert isinstance(embeddings, Tensor)
+        assert isinstance(embeddings, Tensor)
         return embeddings.tolist()
 
     def predict(self, texts: List[List[str]]) -> List[List[float]]:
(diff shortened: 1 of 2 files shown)


shahules786:
Nice catch @alebondarenko , thank you :)

jjmachan:
Hey @alebondarenko 
Thank you so much for helping us improve ragas with your PR ❤️ 
Now since it's Christmas and all we wanted to send you a postcard and a couple of stickers as our way of saying thank you for your contribution. If you are interested could you shoot an email at jamesjithin97@gmail.com and I'll tell you more?

Cheers 🙂 
Jithin

Issue #231, opened 2023-10-29
make heavy  dependencies optional


jjmachan:
currently we have a few heavy dependencies like sentence-transformer and torch, we can reduce this for smaller installables

shahules786:
This will require

- [ ] Removing strictness &[ sentence agreements](https://github.com/explodinggradients/ragas/blob/1b2c79b58c632bbb407a72b3f0a376503b8a17ae/src/ragas/metrics/context_precision.py#L49) from context relevancy
- [ ] Adding bi-encoder methods in [answer similarity  ](https://github.com/explodinggradients/ragas/blob/1b2c79b58c632bbb407a72b3f0a376503b8a17ae/src/ragas/metrics/answer_similarity.py#L45)

sid8491:
currently size of library is very big, removing these dependencies will make it more usable.

jjmachan:
fixed now

PR #279, opened 2023-11-11
fix: make torch optional dependency in utils.py

gsittyz:
#231


jjmachan:
thanks for raising the PR @gsittyz but closing this because we have merged in another fix for the same issue

But thanks again for taking the time to make a PR and my sincere apologies for not being able to merge this in 🙁

PR #297, opened 2023-11-17
Case-insensitive and whitespace handling for context_precision.

Pauldevillers:
Hello,

Description : This PR modifies the response evaluation in the context_precision.py script. The changes include:

Making the response evaluation case-insensitive: 

- The code now converts each response to lowercase before checking if "yes" is in the response. This ensures that the evaluation is not affected by the case of the response.
- Handling whitespace in responses: The code now strips leading/trailing whitespace from each response before checking if "yes" is in the response. This ensures that the evaluation is not affected by whitespace in the response.

### DIFF
--- a/src/ragas/metrics/_context_precision.py
+++ b/src/ragas/metrics/_context_precision.py
@@ -75,7 +75,7 @@ class ContextPrecision(MetricWithLLM):
             scores = []
 
             for response in grouped_responses:
-                response = [int("Yes" in resp) for resp in response]
+                response = [int(any("yes" in r.lower() for r in resp)) for resp in response]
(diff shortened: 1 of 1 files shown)


shahules786:
Great, thank you @Pauldevillers

jjmachan:
Hey @Pauldevillers 
Thank you so much for helping us improve ragas with your PR ❤️ 
Now since it's Christmas and all we wanted to send you a postcard and a couple of stickers as our way of saying thank you for your contribution. If you are interested could you shoot an email at jamesjithin97@gmail.com and I'll tell you more?

Cheers 🙂 
Jithin

PR #302, opened 2023-11-19
Make testgenerator output compatible with evaluate

tinomaxthayil:
* Changed context from str to List[str] so that it is consistent with eval. Now output of TestDataset can be used for evaluation.
*  Changed typo in _generate_doc_nodes_map
*  Changed TestDataset class to reflect the changes in test set generation. Drawback is episode_done will be True in all cases as data is changed at the level above.

### DIFF
--- a/src/ragas/testset/testset_generator.py
+++ b/src/ragas/testset/testset_generator.py
@@ -50,7 +50,7 @@ question_deep_map = {
     "conditional": "_condition_question",
 }
 
-DataRow = namedtuple("DataRow", ["question", "ground_truth_context", "ground_truth", "question_type"])
+DataRow = namedtuple("DataRow", ["question", "ground_truth_context", "ground_truth", "question_type", "episode_done"])
 
 
 @dataclass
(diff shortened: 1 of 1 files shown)


src/ragas/testset/testset_generator.py
```diff
@@ -392,8 +390,9 @@ def generate(
             if is_valid_question:
                 context = self._generate_context(question, text_chunk)
                 answer = self._generate_answer(question, context)
-                samples.append(
-                    DataRow(question.split("\n"), context, answer, evolve_type)
+                samples.extend(
+                    DataRow(qstn, [ctx], ans, evolve_type)
```
shahules786, line 394: `ground_truth` should be `List[str]`  and not `str`

shahules786 (changes requested):

src/ragas/testset/testset_generator.py
jjmachan: shouldn't we change that to `str`?

tinomaxthayil: ground_truth is expecting List[str]

jjmachan (approved):

shahules786 (approved):

PR #307, opened 2023-11-20
Refactor hallucination score calculation in _faithfulness.py

Yongtae723:
The definition of score goes back and forth and confuses newcomer.

This issue related to [discussion on discord](https://discord.com/channels/1119637219561451644/1176098098037145641/1176108672187899965)


shahules786:
Hi @Yongtae723 , thanks for the changes. I would  prefer to add this to communicate the message via a short comment instead of changing the variable name. Does that make sense?

shahules786:
@Yongtae723 This has been addressed in #333 . Thanks a lot for bringing this issue to out notice :) Keep up the good work.

PR #322, opened 2023-11-21
convert context precision to json format

shahules786:
- [x] Manual inspection of results to ensure correctness 
- [x] Linting


src/ragas/utils.py
```diff
@@ -15,3 +17,17 @@ def get_debug_mode() -> bool:
...
+
+def load_as_json(text):
+    """
+    validate and return given text as json
+    """
+
+    try:
+        return json.loads(text)
+    except ValueError:
+        print(text)
```
jjmachan, line 30: remove print and warming message should have the error string from exception.

jjmachan (changes requested):

jjmachan (approved):

Issue #334, opened 2023-11-24
[Feature Request] Custom synthetic data generation prompts


faradri:
Hi!

I apologize if this feature already exists but I found no evidence of it neither in the documentation nor in the source code. 

I'm working with the synthetic data generation feature. I'm using context documents written in italian language but the output questions and the corresponding ground truth answers are in english and it would be great if they were in the same language of the input context documents. 

I've made some changes to the code to use custom prompts and put them here:
[https://github.com/faradri/ragas/tree/testset-prompts](https://github.com/faradri/ragas/tree/testset-prompts)

Thank you

jjmachan:
hey @faradri we are looking into how we can best solve for this and will keep you posted about an easier way to do this rather than having to modify the code yourself 

but welcome to the community :)

faradri:
Hi @jjmachan.

Thank you for your comment and thank you for your welcome! :)
I made a change to the code that seemed to me less "invasive" as possible.

PR #335, opened 2023-11-24
feat(prompts): make the synthetic data test generation prompts customizable

faradri:
- Moved the prompts in testset/prompts.py into an Enum
- Added the possibility to pass custom prompts in TestDataset class

Related issue #334


shahules786:
Hi @faradri , thanks for your contribution. We will look into this.

PR #364, opened 2023-12-07
Retry for failed json loading

shahules786:


src/ragas/utils.py
```diff
@@ -29,3 +37,118 @@ def load_as_json(text):
...
+                        break
+
+                if len(stack) == 0 and start_index != -1:
+                    # Found a valid outermost JSON
+                    return (
+                        start_index,
+                        i + 1,
+                    )  # Add 1 to include the closing brace/bracket in the range
+
+        return -1, -1  # No valid JSON found
```
jjmachan, line 151: write tests for functions like these

src/ragas/utils.py
```diff
@@ -29,3 +37,118 @@ def load_as_json(text):
...
+        return {}
+
+    def _fix_to_json(
+        self,
+        text,
+        llm,
+        callbacks: t.Optional[CallbackManager] = None,
+        callback_group_name: str = "batch",
+    ):
+        with trace_as_chain_group(
```
jjmachan, line 114:
```suggestion
       # TODO (executor)
        with trace_as_chain_group(
```

jjmachan (approved):
only the test part is the problem now but feel free to add it in a following PR

Issue #365, opened 2023-12-09
Definition of context precision


yuukidach:
Hi, in the official document, I found the context precision is defined as: 
> a metric that evaluates whether all of the **ground-truth** relevant items present in the contexts are ranked higher or not. 

But when calculating the score, its computed by **question**
> This metric is computed using the **question** and the contexts

If the precision is about the gts, shouldn't it be compluted by **gts** and the contexts? In addition, since ragas is evaluating RAG, it is likely that the model itself does not have the relevant knowledge and cannot answer this question. How can we use question to verify context?

shahules786:
Hi @yuukidach , nice observation. As of now, we are checking if the part/chunk of the given text can be used to answer the query or not for context precision. So I suppose we should redefine it that way, What do you think?

yuukidach:
Hi @shahules786 . Thanks for replying. 

I think giving a new calculation formula would be more meaningful than modifying the definition. 

In retrieval task, the precision is usually defined as: 
`precision = (number of relevant docs) / (number of total retrieved docs)`

The key here is how to determine whether a doc is relevant.  If we use questions to confirm whether the context is relevant, we need to assume that the evaluate model can fully answer this question. However, an important scenario in rag is that some private knowledge cannot be fully learned by pre-trained models. This can lead to inaccurate measurements.

I tried to construct a test that included 100 questions: 5 docs were retrieved each time, but only the first two were the ones that truly answered the question. In this case, Ragas gave the context precision of 56 questions a perfect score of 1

yuukidach:
Following the current definition of context precision, using gt to measure context precision is actually more intuitive. For example, when we are working on a problem, if we don't know the answer, even if we are given many possible methods, we might still don't know which one is right. But when we have a reference answer, we can filter out effective solutions from a bunch of problem-solving methods

shahules786:
@yuukidach Great, using GT will certainly improve the accuracy of the metric. I want to keep it flexible so that users without access to ground truth also can get some precision-related signals from retrieved chunks. What do you think of a way by which we will automatically switch b/w GT & reference free method depending on the passed dataset for calculating context precision?  

ie, if GT is available -> Use it to calculate CP 
else use the current method.

yuukidach:
@shahules786 The simplest way is to keep the current calculation method and only modify the prompt. I can help with a test to see if modifying the prompt is enough to achieve satisfactory results.

shahules786:
@yuukidach Sure, please give it a go :)

[...]

PR #391, opened 2023-12-18
Context precision with ground truth

shahules786:
#365 fixes


shahules786:
@yuukidach Hey, I would love to know your thoughts on my approach here.

yuukidach:
@shahules786 It's really not easy to merge two PRs. I think we can directly adopt yours. I'll closed my PR.

1. In this PR, there is another metric for dataset with out gt. This makes prompt tuning more convenient
2. Since the prompt are simpler ans shorter, the running time can also be faster.

And for the evaluation of context_relevancy, I agree with your idea. There is a high probability that using GT/Answer can also improve the effect.

BTW, when tuning the prompt, I found that a suitable prompt is still very important in the evaluation process. Will Ragas consider let users to use our own prompts or provide a module that automatically tuning prompt like https://arxiv.org/abs/2309.03409

src/ragas/metrics/_context_precision.py
```diff
@@ -51,26 +69,40 @@ class ContextPrecision(MetricWithLLM):
...
     name: str = "context_precision"  # type: ignore
-    evaluation_mode: EvaluationMode = EvaluationMode.qc  # type: ignore
+    evaluation_mode: EvaluationMode = EvaluationMode.qcg  # type: ignore
     batch_size: int = 15
 
+    def get_dataset_attributes(self, dataset: Dataset):
+        answer = "ground_truths"
+        if answer not in dataset.features.keys():
+            logging.warning(
+                "Using context precision without ground truth will be soon depreciated. Use context_utilization instead"
```
jjmachan, line 79:
```suggestion
                "Using 'context_precision' without ground truth will be soon depreciated. Use 'context_utilization' instead"
```

src/ragas/validation.py
```diff
@@ -60,8 +61,15 @@ def validate_evaluation_modes(ds: Dataset, metrics: list[Metric]):
         required_columns = set(EVALMODE_TO_COLUMNS[m.evaluation_mode])
         available_columns = set(ds.features.keys())
         if not required_columns.issubset(available_columns):
+            extra_msg = ""
+            if (
+                isinstance(m, ContextPrecision)
+                and "ground_truths" not in available_columns
+            ):
+                extra_msg = "Looks like you're trying to use context precision without ground_truths. Please use context_utilization"
```
jjmachan, line 69:
```suggestion
                extra_msg = "Looks like you're trying to use 'context_precision' without ground_truths. Please use consider using  `context_utilization' instead."
```

jjmachan (approved):
awesome! this looks good

PR #388, opened 2023-12-17
Added prompt class

tinomaxthayil:
* Added class RagasPrompt for creating prompt objects
         - RagasPrompt.format - Will take the input parameters and returns the ChatPromptTemplate
         - RagasPrompt.validate - Will check if the prompt object has the field in correct formay

* Added RagasPrompt to init
* Added test_prompt to test prompt object

PR #393, opened 2023-12-19
feat(metrics): update prompt of context precision

yuukidach:
Update prompt of context precision as disscussed in https://github.com/explodinggradients/ragas/issues/365

Here are the results of a test set consisting of 97 medical related Q&A questions (Use GPT4 as evaluator)
[context_precision_test.xlsx](https://github.com/explodinggradients/ragas/files/13711531/context_precision_test.xlsx)
- human_eval: evaluated by human
- ragas_before: evaluated by ragas with only question and context
- ragas_update: evaluated by rags with question, gt and context

Subtract the rating scores from the human_eval using ragas_before and ragas_update respectively and get absolute value
| | mean | std |  number of zeros |
| --- | --- | --- | --- |
| update - human | 0.0676 | 0.1687 |  74 |
| before - human | 0.1796 | 0.2884 | 7 |

## others
1. The use of<></> is because if the context/answer has multiple fragments, it will cause gpt4 to be unable to effectively determine the text range, reducing accuracy
2. Although the prompt looks much longer now, it actually only uses 750 tokens (measured by tiktoken), which is acceptable for most LLMs. However, the running time is indeed longer than before
3. The prompt only tested in the medical related question with gpt4. It might overfit the situation. Maybe we can add a reflection agent like the `envolver` or `filter` in the test_generator to generalize different scenarios

### DIFF
--- a/docs/concepts/metrics/context_precision.md
+++ b/docs/concepts/metrics/context_precision.md
@@ -1,6 +1,8 @@
 # Context Precision
 
-Context Precision is a metric that evaluates whether all of the ground-truth relevant items present in the `contexts` are ranked higher or not. Ideally all the relevant chunks must appear at the top ranks. This metric is computed using the `question` and the `contexts`, with values ranging between 0 and 1, where higher scores indicate better precision.
+Context Precision is a metric that evaluates whether all of the ground-truth relevant items present in the `contexts` are ranked higher or not.  Ideally all the relevant chunks must appear at the top ranks. This metric is computed using the `question`, `ground_truths` and the `contexts`, with values ranging between 0 and 1, where higher scores indicate better precision.
+
+In scenarios where ground truths are unavailable for retrieval tasks, the metric relies on `question` and `context` alone for computation. However, it's important to note that the resulting precision may not be  accurate comparing with the same task with ground truths.
 
 ```{math}
 \text{Context Precision@k} = {\sum {\text{precision@k}} \over \text{total number of relevant items in the top K results}}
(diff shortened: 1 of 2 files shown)


yuukidach:
@shahules786 I found you have already create a PR https://github.com/explodinggradients/ragas/pull/391 . Sorry for not handling this matter for too long. Please have a look to see if 2 PRs can merge into one

shahules786:
Thank you @yuukidach.  No worries, I'll take a look and see how can we merge both.

shahules786:
First of all, thanks a lot for the great work you have put in. I should have left a note before taking up the issue myself. Apologies. 

Overall both of us have solved the problem in a very similar fashion. Some of the differences I can see are that
1) In #391 I accept either answer/ground_truths. My thought was that for any sample with question and context, it is very easy to generate an answer and evaluate it. Most users will at least have these 3 attributed. Although usage of the answer will be depreciated in `context_precision` and can be used with the `context_utilization` metric
2) Changes in tests and Evaluation mode to handle exception situations

I couldn't think of a way to merge both. Do you have any ideas? Would you be open to pushing your documentation to #391 ?

shahules786:
Also, can we modify context_relevancy in similar way to improve performance? Ie it should use (question, context, answer) triplets to find out how much information (number of sentences/tokens/etc) from given context is used arrive at the answer. This can help users to optimise their chunk size. @yuukidach

shahules786:
Sure @yuukidach .  Regarding prompts, we are refactoring and introducing a new idea of prompt object #388 . Prompts being non-trivial parts of software deserve to be more than just strings.  My view of prompts is to write simple and generalizable instructions that are then followed up with good demonstrations rather than overfitting instructions to a particular model. 
Up next, we will also introduce automated prompt language adaptation to allow users to adapt ragas prompts to any target language.   I'll read the paper you shared, thank you.

PR #407, opened 2023-12-25
feat: Automatic Prompt adaptation [language]

shahules786:
Automatic prompt adaption for different languages.  

- [x] Add support for all metrics
- [x] Testing and experimentation
- [ ] Add docs (with explanation)

To be merged only after #401 

### Usage
```
from ragas.metrics import faithfulness
faithfulness.adapt("hindi")
faithfulness.save()
```


src/ragas/llms/prompt.py
```diff
@@ -109,3 +119,110 @@ def format(self, **kwargs: t.Any) -> ChatPromptTemplate:
...
+        cache_path = os.path.join(cache_dir, f"{self.name}.json")
+        with open(cache_path, "w") as file:
+            json.dump(self.to_json(), file, indent=4)
+
+    @classmethod
+    def _load(cls, language: str, name: str, cache_dir: str) -> None:
+        logging.log(logging.INFO, f"Loading {name} from {cache_dir}")
+        path = os.path.join(cache_dir, language, f"{name}.json")
+        cls(**json.load(open(path))["kwargs"])
+
```
jjmachan, line 192: could you explain how this works? this is a classmethod but is not returning the cls and `adapt()` is not getting any class, so how is this working?

src/ragas/llms/prompt.py
```diff
@@ -109,3 +119,110 @@ def format(self, **kwargs: t.Any) -> ChatPromptTemplate:
...
+        results = [result[0].text for result in llm.generate(prompts).generations]
+        per_example_items = len(self.input_keys) + 1
+        grouped_results = [
+            results[i : i + per_example_items]
+            for i in range(0, len(results), per_example_items)
+        ]
+        assert len(grouped_results) == len(
+            self.examples
+        ), "examples and adapted examples must be of equal length"
```
jjmachan, line 159: should this raise an exception or is it okay

jjmachan (changes requested):
- logging module instead of print
- that part of code in `prompt.py` that I didn't quite understand, it could be why cache is slow.

Also try to put in a bit more effort into variable names after the `llm.generate` step. more thoughtful var names can make a big difference in improving readability of the code esp after `llm.generate` because normally we have no idea what the output of that function is.

src/ragas/metrics/_faithfulness.py
```diff
@@ -120,6 +122,19 @@ class Faithfulness(MetricWithLLM):
     evaluation_mode: EvaluationMode = EvaluationMode.qac  # type: ignore
     batch_size: int = 15
 
+    def __post_init__(self: t.Self):
+        self.long_form_answer_prompt = LONG_FORM_ANSWER_PROMPT
+        self.nli_statements_message = NLI_STATEMENTS_MESSAGE
+
+    def adapt(self, language: str, cache_dir: t.Optional[str] = None) -> None:
+        print(f"Adapting Faithfulness metric to {language}")
```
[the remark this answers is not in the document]
shahules786: Will do that, Is there a special reason why we have used print [here](https://github.com/explodinggradients/ragas/blob/84cb885d4bb26073bc1bd43f563238699f5c6cd7/src/ragas/evaluation.py#L100)?

src/ragas/utils.py
```diff
@@ -17,6 +17,12 @@
 # constant to tell us that there is no key passed to the llm/embeddings
 NO_KEY = "no-key"
 
+# Cache location
+DEFAULT_XDG_CACHE_HOME = "~/.cache"
+XDG_CACHE_HOME = os.getenv("XDG_CACHE_HOME", DEFAULT_XDG_CACHE_HOME)
+DEFAULT_RAGAS_CACHE_HOME = os.path.join(XDG_CACHE_HOME, "ragas")
+RAGAS_CACHE_HOME = os.path.expanduser(os.getenv("RAGAS_HOME", DEFAULT_RAGAS_CACHE_HOME))
+
```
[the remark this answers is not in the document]
shahules786: I followed the analogy from HF as shown [here](https://github.com/huggingface/datasets/blob/e23a59ef7ba2b50d4e5588825c41212a3cfd1331/src/datasets/config.py#L148)
User here can 
- change the cache home
- change Ragas home fully
Should this be simplified?

src/ragas/metrics/_faithfulness.py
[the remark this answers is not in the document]
jjmachan: yep because it was a part of the tqdm logs - otherwise that wouldn't make sense (hack) - removed that with executor

src/ragas/metrics/critique.py
```diff
@@ -81,6 +83,14 @@ def __post_init__(self: t.Self):
         self.strictness = (
             self.strictness if self.strictness % 2 != 0 else self.strictness + 1
         )
+        self.critic_prompt = CRITIQUE_PROMPT
+
+    def adapt(self, language: str, cache_dir: str | None = None) -> None:
+        logging.info(f"Adapting Critic to {language}")
```
jjmachan, line 89: create a new logger from logging like this
this goes on top of the file after all imports. The inits a new logger for the module. this means users can set logging level for ragas separately and see where the logs came from
```py
# logging_example.py

import logging

# Create a custom logger
logger = logging.getLogger(__name__)
```

then use
```py
logger.info(f"Adapting Critic to {language}")
```

jjmachan: do this for everything

src/ragas/metrics/critique.py
```diff
@@ -81,6 +83,14 @@ def __post_init__(self: t.Self):
         self.strictness = (
             self.strictness if self.strictness % 2 != 0 else self.strictness + 1
         )
+        self.critic_prompt = CRITIQUE_PROMPT
```
jjmachan, line 86: self.critic_prompt is not defined in the class above right? 
please fix docs string as well

same for all metrics

jjmachan (changes requested):
I'll fix these things

src/ragas/metrics/_faithfulness.py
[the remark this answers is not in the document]
jjmachan: as a rule of thumb - if its for user - print, if it's for dev - log
if its for user - expose in the top most section eg `evaluate(), testset_generation()`
internals are logged

src/ragas/utils.py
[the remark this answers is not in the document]
jjmachan: I think it can but lets merge this for now

src/ragas/llms/prompt.py
jjmachan: leaving it as assert itself - with the assumption that there is no try - except method we can do to catch this.
the con is that the whole operation fails in this case. I think we should raise an Exception and catch it and throw a warning

jjmachan (approved):

PR #409, opened 2023-12-26
feat(metrics): skip embedding process when the corresponding weight is zero

yuukidach:
This PR avoids calculating the embedding when assigning an embedding weight of 0 to answer correctness

Currently, the correctness of the answer relies on both factuality and semantic similarity. However, for most embedding models, the similarity between any two embeddings will be larger than 0. This is equivalent to introducing a bias into the rating system. Therefore, in practical use, I tend to turn off similarity evaluation and only use factual evaluation.

Do you think it would be better to allow embedding to be turned off like this or to add an additional factuality metric?

### DIFF
--- a/src/ragas/metrics/_answer_correctness.py
+++ b/src/ragas/metrics/_answer_correctness.py
@@ -56,7 +56,7 @@ class AnswerCorrectness(MetricWithLLM):
 
     """
     Measures answer correctness compared to ground truth as a combination of
-    semantic similarity and factuality
+    factuality and semantic similarity.
 
     Attributes
     ----------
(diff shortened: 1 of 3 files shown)


shahules786:
> . However, for most embedding models, the similarity between any two embeddings will not be larger than 0
Can you explain @yuukidach

yuukidach:
> > . However, for most embedding models, the similarity between any two embeddings will not be larger than 0
> 
> Can you explain @yuukidach

@shahules786 My mistake, I mean "will be larger than 0"

shahules786:
Excellent work @yuukidach . I like this optimization. Thank you.

Issue #415, opened 2023-12-28
ragas_score removed from Result class in November 2023 but still in documentation


mwertheim-zco:
**Describe the bug**
Ragas documentation [here](https://docs.ragas.io/en/latest/references/evaluation.html#ragas.evaluation.evaluate) states the `ragas_score` can be retrieved directly from result of `evaluate`. However, this throws a `KeyError`
Is there an intended workaround or was this removed with the intention that we would calculate harmonic mean using output?

Ragas version:
Python version: 3.10.13, 3.12.1

**Code to Reproduce**
```
# OPENAI API key set here...

from datasets import load_dataset
from ragas.metrics import (
    answer_relevancy,
    faithfulness,
    context_recall,
    context_precision,
    context_relevancy
)
from ragas.metrics.critique import harmfulness
from ragas import evaluate

fiqa_eval = load_dataset("explodinggradients/fiqa", "ragas_eval")


result = evaluate(
    fiqa_eval["baseline"].select(range(3)),
    metrics=[
        answer_relevancy,
        faithfulness,
        context_recall,
        context_precision,
        context_relevancy,
        harmfulness
    ]
)

print(result['ragas_score'])
```

**Expected behavior**
The result object would have a key/value pair for `ragas_score` as described in documentation.

shahules786:
Hi @mwertheim-zco , you're right. We are letting the users decide on what aggregation they want to do on the score level. Once you have the raw scores you could write simple aggregate functions to do any of these calculations. 
We should remove `ragas_score` from docs. would you like to help?

shahules786:
Thank you. Assigned to you @prufrock349

atursunov:
@shahules786  how can we get ragas_score then?

shahules786:
Hi @atursunov , you can combine the scores in results dict using any aggregation function you prefer.

PR #417, opened 2024-01-02
Feature/doc update ragas score

prufrock349:
PR to resolve issue #415 
`ragas_score` was removed from `Result` class. This PR removes referenced to `ragas_score` in documentatin


shahules786:
This is excellent work @prufrock349 .  Thanks a lot :)

PR #421, opened 2024-01-03
fix(metrics): avoid dividing by zero in correctness

yuukidach:


jjmachan (approved):
thank you so much @yuukidach for the fix ❤️

PR #427, opened 2024-01-05
fix: Import PrompValue 

shahules786:


jjmachan (approved):

dbolotov:
Thank you for the quick turnaround. Encountered another error with the same script, opened issue here: https://github.com/explodinggradients/ragas/issues/444

PR #428, opened 2024-01-05
fix: add uuid to analytics 

shahules786:


src/ragas/_analytics.py
```diff
@@ -57,6 +61,19 @@ def wrapper(*args: P.args, **kwargs: P.kwargs) -> t.Any:
     return wrapper
 
 
+def add_userid(payload: dict[str, t.Any]) -> dict[str, t.Any]:
+    uuid_filepath = os.path.join(USERID_PATH, "uuid.json")
+    if os.path.exists(uuid_filepath):
+        user_id = json.load(open(uuid_filepath))["userid"]
+    else:
+        os.mkdir(USERID_PATH)
+        user_id = uuid.uuid4().hex
```
jjmachan, line 70: here add a `a-` string before it - in case we decide to use some other methods of identification later down the road

src/ragas/_analytics.py
```diff
@@ -75,6 +92,7 @@ def track(event_properties: BaseEvent):
         return
 
     payload = asdict(event_properties)
+    payload = add_userid(payload)
```
jjmachan, line 95: should be part of `BaseEvent` as a default factory probably

tests/unit/test_analytics.py
```diff
@@ -0,0 +1,9 @@
+from __future__ import annotations
+
+from ragas._analytics import add_userid
+
+
+def test_add_userid():
+    payload = {"a": 1, "b": 2}
+    payload = add_userid(payload)
```
jjmachan, line 8: create a new instance of `BaseEvent` and check if it has the same user_id as from appdirs

jjmachan (changes requested):

src/ragas/_analytics.py
```diff
@@ -57,12 +63,23 @@ def wrapper(*args: P.args, **kwargs: P.kwargs) -> t.Any:
...
+        os.mkdir(USERID_PATH)
+        user_id = uuid.uuid4().hex
+        with open(uuid_filepath, "w") as f:
+            json.dump({"userid": user_id}, f)
+    return user_id
+
+
+class BaseEvent(BaseModel):
     event_type: str
+    user_id: str = Field(default_factory=lambda: add_userid)
```
jjmachan, line 80:
```suggestion
    user_id: str = Field(default_factory=add_userid)
```
is enough right?

jjmachan (approved):

PR #430, opened 2024-01-06
feat: added async version for `json_load` 

jjmachan:
- added async version for `json_load.asafe_load`
- ported all metrics

note: in order to get optimal async operation we need to support async embeddings too. will make a followup PR for that

### DIFF
--- a/tests/benchmarks/benchmark_eval.py
+++ b/tests/benchmarks/benchmark_eval.py
@@ -1,5 +1,4 @@
 import time
-import logging
 
 from datasets import DatasetDict, load_dataset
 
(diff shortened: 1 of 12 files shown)


tests/benchmarks/benchmark_eval.py
```diff
@@ -16,9 +16,9 @@
 from ragas.metrics.critique import harmfulness
 
 # data
-ds = load_dataset("explodinggradients/fiqa", "ragas_eval")
+ds = load_dataset("explodinggradients/amnesty_qa", "english")
```
shahules786, line 19: Might want to add `.select(range(0,10))` to fast up the process.

shahules786 (approved):

PR #432, opened 2024-01-06
feat: global prompt adaptation

shahules786:
- [x] prompt adaptation
- [x] Minor fix in faithfulness 
- [x] Docs for prompt adaptation
- [x] Add test

```python
from ragas import adapt
from ragas.metrics import faithfulness, context_precision, context_recall,
adapt(metrics=[faithfulness,context_recall],language="hindi")
```


src/ragas/adaptation.py
```diff
@@ -0,0 +1,33 @@
...
+    llm_wraper = None
+
+    if llm is None:
+        llm_wraper = llm_factory()
+    elif isinstance(llm, BaseLanguageModel):
+        llm_wraper = LangchainLLMWrapper(llm)
+
+    for metric in metrics:
+        if metric.llm is None or llm is not None:
+            metric.llm = llm_wraper
```
jjmachan, line 29: I don't know if this if statement is correct.

if metric has an LLM and the user passes an LLM with adapt functionality then the LLM the user passed metric.llm get replaced right? is that the expected behaviour

jjmachan (commented):
just this one concern

src/ragas/adaptation.py
shahules786: Yes. Say user has used `gpt-3.5` for eval but need to use `gpt-4` for adapt.

Issue #396, opened 2023-12-20
[Bug Report] Duplicated questions are generated while calculating answer relevance


un-certainty:
**Describe the bug**
While cacluating the answer relevance, the temperature is set to 0 by default. So with `n=self.strictness`, it will return `n` same questions, which is unexpected.

https://github.com/explodinggradients/ragas/blob/9850f096fd289c68a7a0cc84063f43797ac0c79f/src/ragas/metrics/_answer_relevance.py#L123-L127

un-certainty:
Also, the answer relavence is now calculated in an indirect way (by embedding similarity). Is it possible that we decompose the question into a list of sub-queries and check the ratio of whether the answer contains related information?

NilAndNone:
I change the prompt to let llm only gen one question, it works for me with my local llm, I believe that GPT will be able to perform even better

shahules786:
@un-certainty thanks for sharing this with us. We have been refactoring code for async operations and hence wasn't able to fix this issue earlier. I have raised a PR for this.

PR #437, opened 2024-01-08
fix: vary temperature based on strictness

shahules786:
fixes: #396


src/ragas/llms/base.py
```diff
@@ -92,6 +92,7 @@ def generate_text(
         stop: t.Optional[t.List[str]] = None,
         callbacks: t.Optional[Callbacks] = None,
     ) -> LLMResult:
+        temperature = temperature if n == 1 else 0.3
```
jjmachan, line 95: make this a part of `BaseRagasLLM` as a new function. The provide documentation so that people implementing the `BaseRagasLLM` can use it

jjmachan (changes requested):

jjmachan (approved):

PR #451, opened 2024-01-10
Update faithfulness.md

francesco-defalco:
I added "in generated answer" since in my opinion is not clear if the number of claims are extracted by the answer or the context.

Thank you


shahules786:
Yes, thank you @francesco-defalco

Issue #429, opened 2024-01-05
AttributeError: 'str' object has no attribute 'get' for context_recall 


0ENZO:
Hello everyone, I am having this error "AttributeError: 'str' object has no attribute 'get'" while computing context_recall metric, 

Ragas version: 0.0.21
Python version: 3.10.13

**Code to Reproduce**
```
from langchain.embeddings import VertexAIEmbeddings
from langchain.chat_models import ChatVertexAI

from ragas.metrics import AnswerCorrectness, AnswerSimilarity
from ragas.llms import LangchainLLM
import ragas

import google.auth

project_id = ""

config = {
    "project_id": project_id,
}

creds, _ = google.auth.default(quota_project_id=project_id)

chat = ChatVertexAI(credentials=creds)
ragas_vertexai_llm = LangchainLLM(chat)

embeddings = HuggingFaceEmbeddings(model_name="BAAI/bge-small-en-v1.5") 

metrics = [
    ragas.metrics.faithfulness,
    ragas.metrics.context_relevancy,
    ragas.metrics.context_recall
]

for m in metrics:
    m.__setattr__("llm", ragas_vertexai_llm)

    if hasattr(m, "embeddings"):
        m.__setattr__("embeddings", embeddings)
        
        
answer_similarity = AnswerSimilarity(llm=ragas_vertexai_llm, embeddings=embeddings)
answer_correctness = AnswerCorrectness(llm=ragas_vertexai_llm, answer_similarity=answer_similarity)
metrics.insert(0, answer_correctness)

results = ragas.evaluate(
    ds_row, 
    metrics=metrics,
)
```

**Error trace**
```
evaluating with [answer_correctness]
100%|██████████| 1/1 [00:02<00:00,  2.25s/it]
evaluating with [faithfulness]
100%|██████████| 1/1 [00:08<00:00,  8.78s/it]
evaluating with [context_relevancy]
100%|██████████| 1/1 [00:01<00:00,  1.24s/it]
evaluating with [context_recall]
  0%|          | 0/1 [00:03<?, ?it/s]
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[135], line 42
     39 answer_correctness = AnswerCorrectness(llm=ragas_vertexai_llm, answer_similarity=answer_similarity)
     40 metrics.insert(0, answer_correctness)
---> 42 results = ragas.evaluate(
     43     ds_row, 
     44     metrics=metrics,
     45 )

[...]

PR #453, opened 2024-01-11
Fix: context_recall AttributeError

0ENZO:
Simple workaround to fix `AttributeError: 'str' object has no attribute 'get'` caused by non consistent LLM evaluation output. Resulting in the use of .get() on a str instead of a dict.  
More on this [issue ](https://github.com/explodinggradients/ragas/issues/429)


shahules786:
@0ENZO thank you. This looks good.

Issue #380, opened 2023-12-14
[RFC] Testset Generation: making it faster and easy to use


jjmachan:
# What is this about?
We have had [Synthetic Test Data generation](https://docs.ragas.io/en/latest/concepts/testset_generation.html) in beta for a while and many of you have given us valuable feedback on the same. Now we are reworking it to be faster and extensible for a wider use.

Ragas takes a novel approach to evaluation data generation. An ideal evaluation dataset should encompass various types of questions encountered in production, including questions of varying difficulty levels. LLMs by default are not good at creating diverse samples as it tends to follow common paths. Inspired by works like [Evol-Instruct](https://arxiv.org/abs/2304.12244), Ragas achieves this by employing an evolutionary generation paradigm, where questions with different characteristics such as reasoning, conditioning, multi-context, and more are systematically crafted from the provided set of documents. This approach ensures comprehensive coverage of the performance of various components within your pipeline, resulting in a more robust evaluation process.
![image](https://github.com/explodinggradients/ragas/assets/5261489/6edc3ca6-7f72-47fb-8b3e-223f8a3695c1)

## Core Components
1. `Evolutions` - this is the core and defines how to evolve the given (context, question) pair into more complex questions - adding more context if needed. 
2. `TestsetGenerator` - this takes the LLM, evolutions, Documents and other configurations and returns the generated testset. This class is also responsible for scheduling the different runs in parallel for max throughput.
3. `DocumentStore` and `Document` - `Document` is a extension of langchain_core's Document abstraction. `DocumentStore` is responsible for connecting with the available documents and giving `Evolutions` and interface to fetch documents (adjacent and similar) as needed.
4. `Filter` - filters critique the output from the evolutions and decides if it should be accepted or not. The `Evolution` decides how to evolve the (context, question) and `Filter` checks if it is acceptable or not.
# Usage

## High Level
User can use it by importing the evolutions, defining the distribution of the evolutions in the final testset and configuring `TestsetGenerator`.
```py
from ragas.testset.evolutions import simple, reasoning, multihop, BaseEvolution
from ragas import TestsetGenerator

# define evolutions you will need
evolutions = {
  simple: 0.4,
  reasoning: 0.4,
  multihop: 0.2
}

generator = TestsetGenerator(
    generator_llm: RagasLLM,
    critic_llm: RagasLLM,
    embeddings_model: Embeddings,	
) -> TestsetGenerator

testset = generator.generate(
		documents: Documents = docs,
[...]

PR #464, opened 2024-01-14
feat: MVP for the new TestsetGenerator - SimpleEvolution

jjmachan:
perform an evolution
```py
from ragas.testset.evolutions import SimpleEvolution, NodeFilter, QuestionFilter, logger

node_filter = NodeFilter(gpt4)
ques_filter = QuestionFilter(gpt4)
se = SimpleEvolution(node_filter, ques_filter)

await se.aevolve(llm, docstore)
```

run evolutions with executor
```py
from ragas.executor import Executor

exec = Executor(is_async=False)
qs = []
for i in range(10):
    se = SimpleEvolution(node_filter, ques_filter)
    exec.submit(se.evolve, llm, docstore, name=f"SimpleEvolution-{i}")

try:
    qs = exec.results()
except ValueError:
    se = SimpleEvolution(node_filter, ques
```

generates 300 samples in <6min, should be scalable enough

related to #380

### DIFF
--- a/src/ragas/testset/evolutions.py
+++ b/src/ragas/testset/evolutions.py
@@ -1,7 +1,9 @@
 from abc import ABC, abstractmethod
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 import logging
 import typing as t
+from fsspec.exceptions import asyncio
+from random import choice, choices
 
 from langchain.prompts import ChatPromptTemplate
 from numpy.random import default_rng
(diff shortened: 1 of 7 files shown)


src/ragas/testset/evolutions.py
```diff
@@ -1,87 +1,201 @@
...
+            return await self.aretry_evolve(llm, docstore)
+        else:
+            # if valid question
+            return seed_question
+
+
+async def simple_evolution(
+    llm: BaseRagasLLM, seed_doc: Document, is_table_present: bool = False
+):
+    if is_table_present:
```
shahules786, line 164: It's better to remove this TABLE_QA part, it was experimental.

shahules786 (approved):

src/ragas/testset/evolutions.py
jjmachan: alrighty - I will remove that logic in the next PR

PR #473, opened 2024-01-17
doc: fix azure_open embeddings

shahules786:


jjmachan:
we can close this right @shahules786

PR #477, opened 2024-01-17
fix: clean up embeddings for ragas and add docs for azure embeddings

jjmachan:
- cleanup `RagasBaseEmbeddings` classes
- added documentation for AzureOpenAI


shahules786 (changes requested):
Can you rerun this with https://huggingface.co/datasets/explodinggradients/amnesty_qa, we are trying to not use fiqa dataset anymore

shahules786 (approved):

PR #516, opened 2024-01-25
Update testset_generation.md

Uzarel:
Explicitly added distributions to dataset generation, otherwise it won't work


jjmachan:
thank you so much for the fix ❤️

jjmachan:
also would love to hear your thoughts on the new testset generator too, what do you think? is it working for you as expected? what would you like to see here?

PR #538, opened 2024-01-31
add normalization factor to answer similarity score calculation

enestydarealmc:
Hi Ragas maintainers,

I was calculating answer similarity and noticed that my score was way above 1. I found out it was because the ground truth and generated answer embeddings were not normalized before fed into cosine similarity.

Note that this does not happen to all embeddings, only those that doesn't have L2-norm equal 1 innately.

It'd be nice if I can get some feedback and have this PR merged.

Thanks a bunch!


shahules786 (approved):
Thanks buddy :)

PR #558, opened 2024-02-05
Added documentation page: Using Ragas with Athina for production performance metrics

shivsak:


jjmachan:
thank you so much for the contributing the integration :)

PR #577, opened 2024-02-08
style: remove additional comments in prompt.py

leehanchung:


shahules786:
Nice catch @leehanchung

PR #588, opened 2024-02-09
fix: replace string formating by dict

shahules786:


shabie:
What model is being used for Bedrock for JSON generation? Also what prompt?

PR #592, opened 2024-02-12
Chore: add language to event tracking

shahules786:


src/ragas/utils.py
```diff
@@ -57,3 +60,14 @@ def is_nan(x):
         return np.isnan(x)
     except TypeError:
         return False
+
+
+def get_feature_language(feature):
```
jjmachan, line 65: type annotate this

jjmachan (approved):
do add type annotation

PR #593, opened 2024-02-13
feat(llms.json_load): Recursively load json lists

pberger514:
Slightly broken json are protected against by the function `ragas.llms.json_load.JsonLoader._find_outermost_json`. However, I've found that for many metrics, gpt4 can often return slightly broken json lists, for which this function returns only the first valid json. Here we wrap `_find_outermost_json` with `_load_all_jsons` which calls it recursively to load the full json list.

I.e. expected output for `'{"1":"2"},  ,, {"3":"4"}]'` is `[{'1': '2'}, {'3': '4'}]`

### DIFF
--- a/src/ragas/llms/json_load.py
+++ b/src/ragas/llms/json_load.py
@@ -83,8 +83,8 @@ class JsonLoader:
         retry = 0
         while retry <= self.max_retries:
             try:
-                start, end = self._find_outermost_json(text)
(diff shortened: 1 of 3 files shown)


shahules786:
This is great catch man @pberger514 Thank you

shahules786 (approved):

jjmachan:
thanks a lot @pberger514 for catching this :)

PR #628, opened 2024-02-18
Made embeddings and LLMs dependent on metric in evaluate function

peanutshawny:
Since we already specify whether or not a metric requires an LLM/embeddings model via inheritence from `MetricWithLLM` and `MetricWithEmbeddings`, there isn't really a need to force the use of default LLMs/embeddings in `evaluate` if, for example, no metrics that need embeddings are specified in `metrics`. 

I believe that initiating an LLM/embedding model at the metric level will help clarify how to use `evaluate`, and will make things simpler in the future when more metrics are added as it decouples needing to initialize LLMs/embedding models for metrics that potentially don't need it. They are even optional arguments to the function itself. 

**Copilot Description**

This pull request mainly refactors the `evaluate` function in the `src/ragas/evaluation.py` file. The changes aim to optimize the import and usage of `llm_factory` and `embedding_factory`, and clarify the function comments. 

Here are the main changes:

* [`src/ragas/evaluation.py`](diffhunk://#diff-ae27b15b33603128d151769a7a1a11ed36bd8151ff2326f81e1478889f87c91fL13-R14): Two new imports were added to the top of the file: `embedding_factory` and `llm_factory` from `ragas.embeddings.base` and `ragas.llms` respectively. This change helps to avoid repetitive imports within the `evaluate` function.

Changes within the `evaluate` function:

* The comments for the `llm` and `embeddings` parameters were updated to specify that the default language model and embeddings are used for metrics which require an LLM or embeddings. This provides more clarity on the function's behavior.
* The conditional logic for setting `llm` and `embeddings` was simplified. The `llm_factory` and `embedding_factory` are now only called when `llm` and `embeddings` are `None` and the corresponding metric requires them. This change removes the need for importing `llm_factory` and `embedding_factory` inside the function.

### DIFF
--- a/src/ragas/evaluation.py
+++ b/src/ragas/evaluation.py
@@ -10,7 +10,8 @@ from langchain_core.language_models import BaseLanguageModel as LangchainLLM
 
 from ragas._analytics import EvaluationEvent, track
 from ragas.callbacks import new_group
-from ragas.embeddings.base import BaseRagasEmbeddings, LangchainEmbeddingsWrapper
+from ragas.embeddings.base import BaseRagasEmbeddings, LangchainEmbeddingsWrapper, embedding_factory
+from ragas.llms import llm_factory
 from ragas.exceptions import ExceptionInRunner
 from ragas.executor import Executor
 from ragas.llms.base import BaseRagasLLM, LangchainLLMWrapper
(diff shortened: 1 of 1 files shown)


jjmachan:
thanks a lot for the fix @peanutshawny 🙂 

btw what do you think of new AI PR helper?

PR #633, opened 2024-02-19
rework classification and recall prompts for mixtral, add json respon…

stepoibm:
Issue #608 

* chg CONTEXT_PRECISION prompt to be more specifc what is expected. Reduces the likelihood of the model generating `verification : {...}` instead of the expected json directly
* chg CONTEXT_RECALL_RA prompt. Reduces likelihood of the model generating int 0 or 1 instead of the expected text "0" or "1".
* chg CONTEXT_RECALL_RA examples to always expect an array instead of switching based on one or multiple statements. Prevents "classification: {...}" vs "classifications: {...}" being generated. Also checks whether "classifications: {...}" was generated as a key and extracts the payload.
* added json payload in log in case of error
* chg expected "Attributed" to "attributed" as the model randomly switches between those

### DIFF
--- a/src/ragas/metrics/_context_precision.py
+++ b/src/ragas/metrics/_context_precision.py
@@ -18,7 +18,7 @@ logger = logging.getLogger(__name__)
 
 CONTEXT_PRECISION = Prompt(
     name="context_precision",
-    instruction="""Given question, answer and context verify if the context was useful in arriving at the given answer. Give verdict as "1" if useful and "0" if not with json output. """,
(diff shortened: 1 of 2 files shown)


stepoibm:
Solves #608 for me, but I don't have access to GPT, so I can't validate that users of that model aren't negatively impacted

jjmachan:
thank you so much @stepoibm for fixing this. @shahules786 would help in testing it out and we'll merge this in 🙂 

really sorry for the delay

shahules786:
Hey @stepoibm thanks for the PR. I like the overall changes you have done and think it would be useful for people running mixtral but I doubt if it's generalizable beyond that. I'm thinking of a way in which I can take the generalizable parts from it and include it in ragas. 
In the mean time, our mixtral users could refer to your work :)

jjmachan:
closing this since with v0.2 this has been made much easier

Issue #642, opened 2024-02-20
Async Executor/Runner slows to a halt with jobs that auto-retry with default (high) `max_wait`


joy13975:
**Describe the bug**
Anything that a) has a throttle limit and b) uses `Executor` (and in turn, `Runner`), such as OpenAI's API, will slow to a halt if too many jobs are requested.

Specifically, in my case I was attempting to generate testset for ~1k documents.
```Python
testset = generator.generate_with_langchain_docs(documents, test_size=N, distributions={simple: 0.5, reasoning: 0.25, multi_context: 0.25})
```
This refuses to even start if either there are too many documents (roughly >300) or too high `test_size`. Example with 500 docs stuck at 0% after 20min:
![image](https://github.com/explodinggradients/ragas/assets/6729737/48e9d2e8-f759-4b1d-a99b-84b4a55460ab)

Reducing `test_size` didn't help with high document count because the place that got stuck was at [docstore.add_documents](https://github.com/explodinggradients/ragas/blob/c18c7f4c9a4bb0f761a0cac916178b349b5705e1/src/ragas/testset/generator.py#L150). However, reducing both document count and `test_size` to < 100 did get it going at reasonable speed.

Ragas version: `0.1.2.dev8+gc18c7f4`
Python version: `3.9.13`

**Code to Reproduce**
Below is just one example that uses `Executor` & `Runner` with higher job count (i.e. 1,000). Document contens averaged ~800 characters. OpenAI API is being used as LLM.
```Python
testset = generator.generate_with_langchain_docs(documents, test_size=N, distributions={simple: 0.5, reasoning: 0.25, multi_context: 0.25})
```

**Error trace**
No error, just gets stuck.

**Expected behavior**
Don't get stuck with high job count.

**Additional context**
Will add my findings in comments below. Related: #394

joy13975:
**Problem Investigation**
[...]

PR #643, opened 2024-02-20
Add support for optional max concurrency

joy13975:
**Added optional Semaphore-based concurrency control for #642** 
As for the default value for `max_concurrency`, I don't know the ratio of API users vs. local LLM users, so the proposed default is an opinionated value of `16`
* I *think* more people use OpenAI API for now vs. local LLMs, thus default is not `-1` (no limit)
* `16` seems to be reasonably fast and doesn't seem to hit throughput limit in my experience

**Tests**
Embedding for 1k documents finished in <2min and subsequent Testset generation for `test_size=1000` proceeding without getting stuck:
image

another 30s passes:
image


joy13975:
Note: currently have `test_size=500` done without problem.
1000 still takes over an hour but does properly proceed rather than getting stuck at any %.

wangyiran33:
Thanks! try to run but encounter a problem`As of 3.10, the *loop* parameter was removed from Semaphore() since it is no longer necessary`

jjmachan:
I love this PR - thank you so much @joy13975 ❤️ - really! thank you for sharing your fix.

I just have one commend though. So I was thinking about the `RunConfig` and wondering it would be a good place to add this one too. the goal of that abstraction is that anything a user wants to configure for running - they can configure it through that and it will be our responsibility to make sure it gets used in the right place.

so as a user would you prefer it to be in `RunConfig` is that an overkill?

That was the only question I had but otherwise, I'm merging this in!

joy13975:
@jjmachan 
Totally overlooked putting it in RunConfig, but I think that might be neater than this. I’ll sort that out and put up another PR.

@wangyiran33 thanks for the report and l’ll make it compatible with 3.10 as well in the next PR.

jjmachan:
makes perfect sense @joy13975 - there seems to be a small type error - if you fix that I'll merge this in

I can fix it too if you want but it will take some time for me to get to this sadly but the PR looks really good so would be awesome to merge it as soon as we can right.

jjmachan:
or not - because if we don't have support for 3.10 that is a deal breaker - thanks for reporting that @wangyiran33 
we might have to refer to the stuff [they did](https://github.com/python-websockets/websockets/issues/916)

joy13975:
@jjmachan yeah I’ll fix all the above points once I get time. Should be up this weekend.

jjmachan:
sounds good - thanks a lot again
lets get this bug fixed 🔥

jjmachan:
awesome - finished my round of testing and it is a massive improvement ❤️ 

thanks again for this @joy13975 for helping us all and contributing this fix 🙂

PR #646, opened 2024-02-21
doc: fix import in the example code: ContextRelevance -> ContextRelevancy

izikeros:


jjmachan:
thank you so much @izikeros 🙂

Issue #606, opened 2024-02-15
Add tests for `llm.json_load`


jjmachan:
**Describe the Feature**
bugs like the one @pberger514 caught in #593 might occur down the road too. We need tests for these sections of code. The catch here is that these are tests which need LLM calls so have to add a new Pytest mark for this

pberger514:
Happy to pick this one up.

jjmachan:
assigned - feel free to give it a go 🙌🏽

jjmachan:
we might need to colaborate to figure out the right pytest magic so that everything is easy to use from the CI. 

I think we will have to test this for a couple of models like OpenAI, Bedrock, Llama2 and Mistral, have it part of the CI

pberger514:
@jjmachan I put a first pass in a draft. Let me know your thoughts, I'm happy to expand to other models if you think this works.

PR #659, opened 2024-02-25
Add a mock for test_json_load

pberger514:
Here's a draft of https://github.com/explodinggradients/ragas/issues/606 for OpenAI. Happy to discuss on how to improve/expand this.

### DIFF
--- a/tests/unit/llms/test_json_load.py
+++ b/tests/unit/llms/test_json_load.py
@@ -1,0 +1,58 @@
+import types
+import os
+import unittest.mock
+
+from openai.types.chat import ChatCompletionMessage
+from openai.types.chat.chat_completion import ChatCompletion, Choice
+
+from ragas.llms.base import llm_factory
+from ragas.llms.json_load import json_loader
+
+
+class MockOpenAI:
+    class MockCompletions:
+        def create(self, *args, **kwargs):
+            return ChatCompletion(
+                id="123456",
+                model="model",
+                object="chat.completion",
+                choices=[
+                    Choice(
+                        finish_reason="stop",
+                        index=0,
+                        message=ChatCompletionMessage(
+                            content='{"one": "two"},,, {"three": "four"}]',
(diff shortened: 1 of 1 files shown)


tests/unit/llms/test_json_load.py
```diff
@@ -0,0 +1,58 @@
...
+                        finish_reason="stop",
+                        index=0,
+                        message=ChatCompletionMessage(
+                            content='{"one": "two"},,, {"three": "four"}]',
+                            role="assistant",
+                        ),
+                    )
+                ],
+                created=123,
+            )
```
jjmachan, line 30: should we mock openai or langchain_openai

pberger514: So I'm not sure why, but I can't seem to get the patch to catch a higher abstraction. Here's an example

```
class MockRagasLLM(BaseRagasLLM):
    def generate_text(
        self,
        prompt: PromptValue,
        n: int = 1,
        temperature: float = 1e-8,
    ) -> LLMResult:
        return LLMResult(
            generations=[
                ChatGeneration(message=AIMessage('{"one": "two"},,, {"three": "four"}]'))
            ]
        )
```

then later on

```
        self.patches = [
...
            unittest.mock.patch(
                "ragas.llms.LangchainLLMWrapper",
                new=MockRagasLLM,
            ),
        ]
```

But again I can't get it to catch. I.e. I get a bad API key error from OpenAI, which shouldn't be happening because the patch should be catching the call to generate_text before it even gets to OpenAI.

pberger514: Committed it to the branch here: [9a76d41](https://github.com/explodinggradients/ragas/pull/659/commits/9a76d4106e5b6e8ddd62f117d1f6dfa93cc3e9d3)

jjmachan:
hey with v0.2 we're using a different method but thanks again for this PR! 

PR #666, opened 2024-02-27
Update evolution

cuent:
- If the evolution filter is set to True, indicating that sentences are equivalent in meaning, should the compressed question not be returned?
- typo getting relevant contexts, the prompt is expected to output `relevant_contexts`

### DIFF
--- a/src/ragas/testset/evolutions.py
+++ b/src/ragas/testset/evolutions.py
@@ -193,7 +193,7 @@ class Evolution:
             results.generations[0][0].text.strip(), llm=self.generator_llm
         )
         relevant_context_indices = (
-            relevent_contexts_result.get("relevant_context", None)
+            relevent_contexts_result.get("relevant_contexts", None)
             if isinstance(relevent_contexts_result, dict)
             else None
         )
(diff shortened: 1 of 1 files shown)


shahules786:
Hey bro, thanks a lot for this @cuent that's a nice catch!

shahules786:
Can we make release today @jjmachan ? 
Hey @cuent there was a misunderstanding in the filter evolution part of your PR, I later corrected it. If you look closely you'll see that the prompt for filtering returns "1" if the questions are the same. Then in this [line](https://github.com/explodinggradients/ragas/blob/12ad1905f2cfba5a794ddef33bdee79413778d72/src/ragas/testset/filters.py#L124) it will return True if the output was "1". In which case we need to retry the evolution.

cuent:
Ah, I see. It seems logical to me for MultiContextEvolution, as one can anticipate that questions may vary across different contexts. However, for ComplexEvolution, is it necessary for questions to be different? If so, what is the rationale behind this requirement?

Thank you for your prompt response! 😄

Issue #482, opened 2024-01-18
Metric to account for entities


sky-2002:
**Note**: This is just an idea, open to discussions.

**Problem**: For use cases(like financial, legal, health-care) where the data might contain names of people, places, organizations etc, most of the times, these entities are important and need to be present in the context for the model to understand and use them or might need to be present in the answer. There should be a metric to account for the inclusion of these entities in the context or the answer.

**Proposed solution**: I think of an `entity_recall` metric. Now this can be thought of in two ways:
1. One is to calculate the fraction of entities in the context to the entities in question/query.
2. Other one is to calculate the fraction of entities in the answer to the entities in the context. 

Feel free to point out if such thing already exists or is already ruled out.

Otherwise, if you find this interesting, or worth adding, I am interested in working on this, already have an implementation plan in mind. 
@shahules786 @jjmachan

shahules786:
Hi @sky-2002 , this is certainly an interesting metric. Could you also help me understand how can someone use this metric to improve their RAG/ get more understanding about the performance?

sky-2002:
@shahules786 Okay, I will try to give an example.
Consider we have an LLM to answer factual questions related to stock market/finance,
Now we would be using a retrieval service to provide relevant context, and thus need some evaluation metric for this system.

**Note:** Just for example purposes

**Example 1**
```
Query - 
What is the latest financial status of Apple Inc.?

Relevant correct context -
Apple Inc. (AAPL) recently reported its financial results for the last quarter. The company's revenue increased by 20% compared to the previous quarter, reaching $120 billion. Net profit also saw a significant rise, standing at $25 billion. Apple's CEO, Tim Cook, expressed optimism about the company's performance and outlined key strategies for the upcoming fiscal year during the earnings call.

Retrieved context - 
In the world of technology, Apple is a major player. The latest iPhone release has been a huge success, contributing to the company's revenue growth. Additionally, the CEO, Tim Cook, has been actively involved in various philanthropic initiatives.
```
Query entities - `[Apple Inc]`
Relevant context entities - `[Apple Inc, APPL, $120 Billion, Tim Cook]`
Retrieved context entities - `[Apple, Tim Cook, Iphone]`
So the `entity_recall` considering contexts is `0.5` as 2 of 4 entities in correct context are also in retrieved context

Now I understand there some vagueness in what to compare to, we have 4 things during evaluation(correct me if I am wrong here)- the query, correct context, retrieved context, and the generated text
[...]

PR #667, opened 2024-02-27
feat: Add new "entity-based context recall" metric

sky-2002:
This PR adds a new metric, "context_entity_recall", based on "entities" present in ground-truth and context, to evaluate retrieval mechanisms. I had discussed this metric in #482.

**Description**: 
In use cases where entities like names, places, dates etc matter, it is important that our retrieval mechanism fetches the right context, which has to be fed to the model. An example is tourism industry, names of cities, tourist places, special-dishes, etc matter quite a lot. Other examples could be health or finance domains. 

**Formulation of  `context_entity_recall`**:
I propose this new metric to be the ratio of number of entities present in `ground_truths` and `contexts` both to the number of entities present in `ground_truths` alone. This is like a measure of coverage of `ground_truth entities` by the `context`. I have used a simple formula -  $|CE \cap GE| / |GE|$ - where $CE$ and $GE$ are set of entities in `contexts` and `ground_truths` respectively.

**Example usage in RAGAS**:
```python
from datasets import load_dataset
from ragas import evaluate
from ragas.metrics import context_entity_recall

amnesty_qa = load_dataset("explodinggradients/amnesty_qa")

amnesty_result = evaluate(
    amnesty_qa["eval"].select(range(2)), 
    metrics=[
        context_entity_recall
    ],
    llm = LangchainLLMWrapper(llm) # I used llm separately because I am using OpenRouter
)
amnesty_result.to_pandas()
```
Below is the output:
image

**LLM setting used to generate above example**:
```python
import os
from langchain.chat_models import ChatOpenAI

llm = ChatOpenAI(
    model = "google/gemma-7b-it:free",
    openai_api_key=os.getenv("OPENROUTER_API_KEY"),
    openai_api_base="https://openrouter.ai/api/v1"
)
```

**Discussions and further thoughts**:
As far as I am aware, no other such metric has been used, so this needs to be tested for usefulness and impact. Also, there can be versions of this. Initially I have considered evluating retrieval, so I used `ground_truths` and `contexts`. 
If we need to evaluate our LLM on how it handles entities in context, we can use this similar metric on `answer` and `contexts`.
Or in general, on any pair of the inputs.

I am open to and would encourage discussions, in the community and with mentors, regarding usefulness and additions to this metric. Looking forward for feebback/review.
@shahules786 @jjmachan

### DIFF
--- a/src/ragas/metrics/__init__.py
+++ b/src/ragas/metrics/__init__.py
@@ -12,6 +12,8 @@ from ragas.metrics._context_relevancy import ContextRelevancy, context_relevancy
 from ragas.metrics._faithfulness import Faithfulness, faithfulness
 from ragas.metrics.critique import AspectCritique
 
+from ragas.metrics._context_entities_recall import ContextEntityRecall, context_entity_recall
+
 __all__ = [
     "AnswerCorrectness",
     "answer_correctness",
@@ -30,4 +32,6 @@ __all__ = [
     "ContextRelevancy",
     "AnswerRelevancy",
     "answer_relevancy",
+    "ContextEntityRecall",
+    "context_entity_recall"
 ]
--- a/src/ragas/metrics/__init__.py
+++ b/src/ragas/metrics/__init__.py
@@ -1,6 +1,10 @@
 from ragas.metrics._answer_correctness import AnswerCorrectness, answer_correctness
 from ragas.metrics._answer_relevance import AnswerRelevancy, answer_relevancy
 from ragas.metrics._answer_similarity import AnswerSimilarity, answer_similarity
+from ragas.metrics._context_entities_recall import (
+    ContextEntityRecall,
+    context_entity_recall,
+)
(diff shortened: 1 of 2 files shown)


shahules786:
Hey bro @sky-2002 , this is totally cool. I'm sure that the community will love it. As you can see I have made some changes to your code. Some feedback that you can keep in mind for your next PR
- Always ensure you're using the latest version of ragas, some of the code was written as per older version
- Number of calls to llm vs accuracy and reproducibility tradeoff: as of you can see I have simplified the task and made 2 calls instead of one, this is to ensure higher quality and reproducibility of metrics. 


By the way it would be great if you can add docs for this, for reference like [here](https://github.com/explodinggradients/ragas/blob/main/docs/concepts/metrics/context_precision.md). This would help community members educate about your metrics and drive more usage 🚀

shahules786 (approved):

Issue #636, opened 2024-02-20
Testset generator not working using AzureOpenai key.


rahul1-995:
I am trying to generate synthetic data using azure openai api, taking long time to run and after that getting error.

Ragas version:  0.1.1
Python version: 3.10

**Code to Reproduce**
import os
os.environ["AZURE_OPENAI_API_KEY"] = "AZURE_OPENAI_API_KEY"

azure_configs_gen = {
    "base_url": "",
    "model_deployment": "gpt-35-turbo-16k",
    "model_name": "gpt-35-turbo-16k",
    "embedding_deployment": "text-embedding-ada-002",
    "embedding_name": "text-embedding-ada-002",
}

azure_configs_critic = {
    "base_url": "",
    "model_deployment": "gpt-4",
    "model_name": "gpt-4",
    "embedding_deployment": "text-embedding-ada-002",
    "embedding_name": "text-embedding-ada-002", 
}
generator_llm = AzureChatOpenAI(
    openai_api_version="2023-05-15",
    azure_endpoint=azure_configs_gen["base_url"],
    azure_deployment=azure_configs_gen["model_deployment"],
    model=azure_configs_gen["model_name"],
    validate_base_url=False,
)
generator_llm = LangchainLLMWrapper(generator_llm)

critic_llm = AzureChatOpenAI(
    openai_api_version="2023-05-15",
    azure_endpoint=azure_configs_critic["base_url"],
    azure_deployment=azure_configs_critic["model_deployment"],
    model=azure_configs_critic["model_name"],
    validate_base_url=False,
)

critic_llm = LangchainLLMWrapper(generator_llm)

embed_model = AzureOpenAIEmbeddings(
    openai_api_version="2023-05-15",
    azure_endpoint=azure_configs_gen["base_url"],
    azure_deployment=azure_configs_gen["embedding_deployment"],
    model=azure_configs_gen["embedding_name"],
)
embed_model = LangchainEmbeddingsWrapper(embed_model)

pdf_path = r"machinelearning-lecture01.pdf"
documents = SimpleDirectoryReader(input_files=[pdf_path]).load_data()
#type(documents)

splitter = TokenTextSplitter(chunk_size=2000, chunk_overlap=100)
keyphrase_extractor = KeyphraseExtractor(llm=generator_llm)
docstore = InMemoryDocumentStore(
splitter=splitter,
embeddings=embed_model,
extractor=keyphrase_extractor,
)
from ragas.testset import TestsetGenerator
[...]

Issue #635, opened 2024-02-20
TestsetGenerator with custom LLM : 'ChatOpenAI' object has no attribute 'set_run_config'


thomasbtnfr:
Hi, I'm trying to use the `TestsetGenerator` to generate a dataset from Langchain document. 

Here's my code:
```python
from langchain_openai import ChatOpenAI
from langchain_community.embeddings import HuggingFaceEmbeddings
from ragas.llms.base import LangchainLLMWrapper
from ragas.embeddings.base import LangchainEmbeddingsWrapper
from ragas.testset.generator import TestsetGenerator
from ragas.testset.evolutions import simple, reasoning, multi_context
from langchain.document_loaders import TextLoader

loader = TextLoader(doc_filename)
documents = loader.load()

llm_api = ChatOpenAI(
    model=model,
    openai_api_key=key,
    openai_api_base=base,
)

embeddings = HuggingFaceEmbeddings(
    model_name=model_name_embedding,
)

generator_llm = LangchainLLMWrapper(llm_api)
critic_llm = LangchainLLMWrapper(llm_api)
embeddings_model = LangchainEmbeddingsWrapper(embeddings)

splitter = RecursiveCharacterTextSplitter(chunk_size=250, chunk_overlap=20)
keyphrase_extractor = KeyphraseExtractor(llm=generator_llm)

docstore = InMemoryDocumentStore(
    splitter=splitter,
    embeddings=embeddings_model,
    extractor=keyphrase_extractor,
)

generator = TestsetGenerator(
    generator_llm=generator_llm,
    critic_llm=critic_llm,
    embeddings=embeddings_model,
    docstore=docstore
)

testset = generator.generate_with_langchain_docs(documents, test_size=10, distributions={simple: 0.5, reasoning: 0.25, multi_context: 0.25})
```

With this code I get the following error: `"AttributeError: 'ChatOpenAI' object has no attribute 'set_run_config'`.

It happens here:
```
in Evolution.set_run_config(self, run_config)
     91     self.docstore.set_run_config(run_config)
     92 if self.generator_llm:
---> 93     self.generator_llm.set_run_config(run_config)
     94 if self.node_filter:
     95     self.node_filter.set_run_config(run_config)
```

Is this an error on my part or a bug?
Thanks.

thomasbtnfr:
I tried again and now it seems to be working. 
[...]

Issue #342, opened 2023-11-27
TestsetGenerator with own LLM


weissenbacherpwc:
Hi,

I want to use the Ragas TestsetGenerator to create a synthetic dataset. The function is using OpenAI in the background, but I was wondering if you can set your own LLM to create the synthetic dataset? 

For example for evaluation it is possible to set an own LLM and not an OpenAI Model. Like here:
```
from ragas.llms import LangchainLLM
from src.llm import build_llm


llm = build_llm() # this is loading the Llama 2 model with LlamaCpp from Langchain
llm = LangchainLLM(llm=llm)
from ragas.metrics import (
    context_precision,
    answer_relevancy,
    faithfulness,
    context_recall,
)
from ragas.metrics.critique import harmfulness

# change the LLM

faithfulness.llm = llm
answer_relevancy.llm = llm
context_precision.llm = llm
context_recall.llm = llm
harmfulness.llm = llm
```

jjmachan:
hey @weissenbacherpwc welcome to the community :)

this is possible you can create a new testset generation object to achieve it but the issue is that it might change in the future because testset generation is a beta feature.

but check the code [here](https://github.com/explodinggradients/ragas/blob/fe343d6fdcd0e11096fc1b8fbf4f36d289f32eaa/src/ragas/testset/testset_generator.py#L87) . you can create a new `TestsetGenerator` with any `LangchainLLM`

Issue #230, opened 2023-10-29
Making custom LLMs compatible with TestsetGenerator?


Data-drone:
Hi Team,

I am trying to get the `TestsetGenerator` to work properly with a Langchain custom LLM.

It's `_call` method current expects a string prompt but it gets a `ChatPromptTemplate` from ragas. what is the best way to handle this?

Data-drone:
I got things working by just running `format()` against the prompt my LLM received if it was in `ChatPromptTemplate` format but not sure that is the best approach.

I have set the `chat_qa=0.0` so shouldn't all the prompts be coming in as string?

shahules786:
Hey @Data-drone , can you share the ragas & python version you're using?

Data-drone:
ragas==0.0.18 python 3.10.12

shahules786:
Hi @Data-drone , We will look into the langchain [custom](https://python.langchain.com/docs/modules/model_io/models/llms/custom_llm) llms issue. Regarding setting `chat_qa=0`, it is used to control if the test set should contain conversational questions or not.

shahules786:
Hi @Data-drone , which model are you using? Are you using an instruction-tuned model (that is not chat based)?

Data-drone:
Vicuna 1.5 13b

jjmachan:
hey @Data-drone , so we do have this snipped inside RagasLLMs that should handle this logic for you
```py
        if isinstance(self.llm, BaseLLM):
            ps = [p.format() for p in prompts]
            result = self.llm.generate(ps, callbacks=callbacks)
        else:  # if BaseChatModel
            ps = [p.format_messages() for p in prompts]
            result = self.llm.generate(ps, callbacks=callbacks)
```

if your custom model is a base class of either, it should do the conversation for you. but not sure why its not working here. Could you show us how you're calling the testsetgenerator and custom models code?

msunkarahend:
@shahules786 @jjmachan In the same context, TestsetGenerator uses cross-encoder/stsb-TinyBERT-L-4/ from hugging face. Is there a way,  i can avoid using that and just use gpt models from azure open ai alone and generate synthetic dataset?

jjmachan:
but @shahules786 probably we should change `from_defaults` class method to `with_openai` or something like that to make it easier?

msunkarahend:
@jjmachan If you can share the finished snippet for the TestsetGenerator, it would be great. thanks in advance.

hbj52:
[...]

PR #670, opened 2024-02-27
feat(generator): extend construction to any langchain LLM and Embeddings

mspronesti:
## **User description**
The current version of `with_openai` contains a hardcoded instantiation of `langchain_openai.chat_models.ChatOpenAI`, which makes `TestsetGenerator` very limited and not compatible with completion models, Azure OpenAI models, and open-source models. 

This PR extends `TestsetGenerator` to any `BaseLanguageModel` and `Embeddings` from langchain for versatility, addressing #230, #342, #635, and #636. 

Lastly, I've removed all the occurrences of mutable default arguments (bad antipattern, read [here](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments)).


mspronesti:
Just finalized the PR - It's ready for review now :)

shahules786:
Wow, thanks man @mspronesti Will review it ASAP

mspronesti:
Hi there @shahules786, any chance this PR gets reviewed? :)

shahules786:
Hi @mspronesti apologies for the late review. @jjmachan was testing this out. Will be reviewed soon.

shahules786:
Hi @mspronesti just finished testing and making some changes to avoid breaking changes. 
I'm not sure if you tested it on any instruct models, but the changes made don't allow the use of instruct models.
Also if you have time can you remove the documentation changes that were made inorder to tackle the breaking change (with_openai)? 

@jjmachan you can do the review now.

shahules786 (changes requested):
remove all changes made to tackle the removal of (with_openai) function

mspronesti:
@shahules786 When I split this PR into two (the second part being #684) I didn't realize that completion models relied on the fact that this [parameter](https://github.com/explodinggradients/ragas/blob/main/src/ragas/llms/base.py#L84) needs to default to `None` not to bump into an `IndexError` (besides the mutable defaults problem).   

Could you try again defaulting that to `None` as done in #684?

Regarding the docs, my suggestion would be to mark `with_openai`  for deprecation (with a decorator) and discontinue it in the next major release. As such, I would advise showcasing the new usage in the docs :)

jjmachan:
hey @mspronesti - I agree from_defaults might be ambiguous so lets keep it as `from_langchain` itself as you first proposed.

as for deprecation I've added the code in this PR but let me think we have to remove it in this version

and for the completion bug - nice catch seems like it fixed the issue ❤️

jjmachan (approved):

mspronesti:
Hi @jjmachan, thanks for the review :) 

I wouldn't remove `with_openai` in this version, but in the next major release (0.2). I'd decorate it as deprecated now, suggesting `from_lanchain` as alternative. What do you think? :)

mspronesti:
@jjmachan I've just added an extra commit to make the deprecation decorator more versatile and informative, and I've used it on `with_openai` to mark its deprecation as "pending" so that you can schedule that whenever convenient.

jjmachan:
this is awesome @mspronesti - I love the depreciation util func - handy!
Once @shahules786 unblocks us we should be good to go and do a release later today as well


btw would love to hop on a call with you sometime if you are interested @mspronesti 🙂

shahules786:
@mspronesti Yes, that actually makes sense. Let's do that. I'll also add some more docs to help community use non-open ai models with it.

shahules786 (approved):

Issue #653, opened 2024-02-23
`Metric.adapt()` throws `AssertionError: Adapted output keys do not match with the original output keys`


joy13975:
**Describe the bug**
The `adapt` function on metrics does not succeed for Japanese.
According to [Import and adapt evolutions](https://docs.ragas.io/en/latest/howtos/applications/use_prompt_adaptation.html) one needs to adapt evoluations for non-English languages, so I thought metrics need adaptation too. Is it not the case?

Ragas version: `0.1.2.dev8+gc18c7f4`
Python version: Python 3.9.13

**Code to Reproduce**
```Python
from ragas.metrics import (
    context_precision,
    context_recall,
    faithfulness,
    answer_relevancy,
)
from langchain_openai.chat_models import ChatOpenAI
from ragas.llms import LangchainLLMWrapper

metrics = [
    context_precision,
    context_recall,
    faithfulness,
    answer_relevancy,
]
eval_prompt_cache_dir='../../data/qag/ragas_prompts/'
print(f'Adapting prompts...')
metric_llm = LangchainLLMWrapper(ChatOpenAI(model='gpt-3.5-turbo'))
for m in metrics:
    m.llm = metric_llm
    _ = m.adapt(language='japanese', cache_dir=eval_prompt_cache_dir)
```

**Error trace**
```log
{'reason': '提供された文脈は、与えられた答えに到達するのに本当に役立ちました。文脈には、アルバート・アインシュタインの生涯と貢献に関する
重要な情報が含まれており、それが答えに反映されています。', 'verdict': '1'}
{'reason': '2020 आईसीसी विश्व कप के संदर्भ में जानकारी स्पष्ट करने में उपयोगी था और इसका संकेत देता था कि इंग्लैंड वह विजेता था जो 2020 में आयोजित किया जाना
था पर वास्तव में 2022 में हुआ।', 'verdict': '1'}
{'reason': '提供された文脈はアンデス山脈について話しており、印象的ですが、エベレスト山を含んでおらず、世界で最も高い山に関する質問と直接
関係していません。', 'verdict': '0'}
{'0': {'reason': 'आइंस्टीन का जन्म तिथि स्पष्ट रूप से संदर्भ में उल्लिखित है।', 'statement_1': '14 मार्च 1879 को जन्मे अल्बर्ट आइंस्टीन, एक जर्मन मूल के सिद्धांत
[...]

PR #676, opened 2024-02-28
Fix typos that cause fatal assertion errors in prompts

joy13975:
This seems to be the cause of #653 when LLM attempts correct the typo and end up with keys not strictly identical to "expected" keys.


shahules786 (approved):

shahules786:
Thank you @joy13975 this is great.

Issue #656, opened 2024-02-24
[R-273] 'temperature' parameter in LangchainLLMWrapper.generate_text causing issues


Kirushikesh:
**Describe the bug**

LangchainLLMWrapper has .generate_text() function which further calls .generate_prompt() of the underlying LLM. The LangchainLLMWrapper passes 'temperature' parameter in .generate_prompt() function which causes the following issues,

1. temperature parameter is not affecting the response when using HuggingFace LLM
2. Some Langchain Extensions like [IBM Generative AI](https://github.com/IBM/ibm-generative-ai/tree/main) doesn't support temperature parameter to be passed in .generate_prompt() function.

Since when initialising an LangChain LLM we can pass the temperature as a parameter, it is not needed to be supplied additionally in LangchainLLMWrapper. 

For ex in HuggingFacePipeline, you can specify the temperature when initialization using:
`pipe = pipeline("text-generation", model=model, tokenizer=tokenizer, temperature=1)`

Or when using IBM LLM you can specify the temperature by:
```
llm = LangChainInterface(
        model_id='google/flan-t5-xl',
        client=Client(credentials=Credentials.from_env()),
        parameters=TextGenerationParameters(
                  decoding_method=DecodingMethod.SAMPLE,
                  max_new_tokens=1000,
                  min_new_tokens=1,
                  temperature=0.2,
                  top_k=20,
                  top_p=1,
                  random_seed=42,
                  repetition_penalty = 1.1
              )
)
```

Ragas version: 0.1.1
Python version: 3.10.6

**Code to Reproduce**
The following code explains why 'temperature' parameter not affecting the response in HuggingFaceLLM
```
from langchain_community.llms.huggingface_pipeline import HuggingFacePipeline
from transformers import AutoModelForCausalLM, AutoTokenizer, pipeline

from ragas.llms.base import BaseRagasLLM, LangchainLLMWrapper
from ragas.run_config import RunConfig
from ragas.llms.prompt import PromptValue

model_id = "gpt2"
tokenizer = AutoTokenizer.from_pretrained(model_id)
model = AutoModelForCausalLM.from_pretrained(model_id)
pipe = pipeline("text-generation", model=model, tokenizer=tokenizer, max_new_tokens=100)
hf_llm = HuggingFacePipeline(pipeline=pipe)

pv = PromptValue(prompt_str='hi, how are you')
run_config = RunConfig()
[...]

PR #677, opened 2024-02-28
Support google gemini

joy13975:
This change allows one to use `langchain-google-genai` and simply

```Python
generator = TestsetGenerator.with_google(...)
```

just like one would with `with_openai()` and get on with life.
Initial tests show no errors in embedding and testset generation. Prompt adaptations work on condition of #676 being merged first.

---

CI/CD check fail: `langchain-google-genai` requires `Python >=3.9, <4.0`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,6 +8,7 @@ dependencies = [
     "langchain-core",
     "langchain-community",
     "langchain_openai",
+    "langchain-google-genai",
     "openai>1",
     "pysbd>=0.3.4",
     "nest-asyncio",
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -8,6 +8,7 @@ dependencies = [
     "langchain-core",
     "langchain-community",
     "langchain_openai",
+    "langchain-google-genai",
```
jjmachan, line 11: what do you think of making this an optional dependency?
something like `ragas[google]`.

I'm waiting for our approval to use GCP so will test these after that

joy13975:
I actually think #670 could be a better way of dealing with different vendors.
However, one thing that still needs to be fixed is `temperature` arg being a bit of a mess (having and arg but not actually using it). When passed into the Langchain LLM it sometimes causes errors if the Langchain LLM isn't expecting it as an arg such as in #656

jjmachan:
hey @joy13975 I agree that #670 is turning out to be the better implementation, so we'll merge that in. 
as for #656 yeah that needs work and will try to get it sorted

PR #684, opened 2024-02-29
fix(ragas): remove mutable defaults

mspronesti:
Minor PR to clean the whole codebase from the well-known gotcha of [mutable default arguments](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments) - which I encourage you guys to be more careful of.


shahules786 (approved):

shahules786:
Thanks for the PR @mspronesti nice catch

jjmachan:
thank you so much for the PR @mspronesti ❤️ 
will keep an eye out for these kinds of bugs in the future - thanks for bringing it up and fixing it 🙂

Issue #681, opened 2024-02-29
New executor throws `RuntimeError: ... got Future <..> attached to a differen t loop`


joy13975:
**Describe the bug**
Calling `TestsetGenerator.generate_with_langchain_docs()` throws the following error, then subsequent attempts to embed (using the same Langchain cache) would get stuck and never proceed past 0%. Suspecting cache db(sqlite3) corruption but then integration check reported no problem..

This requires deeper investigation but I'm raising it before I forget/lose the details.

Ragas version: `0.1.3`
Python version: `Python 3.9.13`

**Code to Reproduce**
```Python
lc_docs = ...
testset = generator.generate_with_langchain_docs(
        documents=lc_docs,
        test_size=20,
        distributions={simple: 0.5, reasoning: 0.25, conditional: 0.25},
    )
```

**Error trace**
```log
Exception in thread Thread-7299:                                                                          
Traceback (most recent call last):                                                                        
  File "/home/ec2-user/.pyenv/versions/3.9.13/lib/python3.9/threading.py", line 980, in _bootstrap_inner  
    self.run()
  File "/home/ec2-user/code/ragas/src/ragas/executor.py", line 93, in run                                 
    results = self.loop.run_until_complete(self._aresults())                                              
  File "/home/ec2-user/.pyenv/versions/3.9.13/lib/python3.9/asyncio/base_events.py", line 647, in run_unti
l_complete                                                                                                
    return future.result()                                                                                
  File "/home/ec2-user/code/ragas/src/ragas/executor.py", line 81, in _aresults                           
    raise e                                                                                               
  File "/home/ec2-user/code/ragas/src/ragas/executor.py", line 76, in _aresults                           
    r = await future                                                                                      
  File "/home/ec2-user/.pyenv/versions/3.9.13/lib/python3.9/asyncio/tasks.py", line 611, in _wait_for_one 
    return f.result()  # May raise f.exception().                                                         
  File "/home/ec2-user/code/ragas/src/ragas/executor.py", line 36, in sema_coro                           
    return await coro                                                                                     
  File "/home/ec2-user/code/ragas/src/ragas/executor.py", line 109, in wrapped_callable_async             
    return counter, await callable(*args, **kwargs)                                                       
  File "/home/ec2-user/code/ragas/src/ragas/testset/evolutions.py", line 145, in evolve                   
    ) = await self._aevolve(current_tries, current_nodes)                                                 
[...]

PR #689, opened 2024-03-01
Remove unncessary Runner being on a different thread creating new async loop

joy13975:
This seems to solve #681 as I can no longer reproduce the problem

### DIFF
--- a/src/ragas/executor.py
+++ b/src/ragas/executor.py
@@ -23,15 +23,15 @@ def as_completed(
     # support Python < 3.10 where loop argument is still required
     loop_arg_dict = {"loop": loop} if sys.version_info[:2] < (3, 10) else {}
     if max_workers == -1:
-        return asyncio.as_completed(coros, **loop_arg_dict)
+        return asyncio.as_completed(coros, **loop_arg_dict)  # type: ignore
     
     semaphore = asyncio.Semaphore(max_workers, **loop_arg_dict)
-    async def sema_coro(coro):
+    async def sema_coro(coro: asyncio.Future):
         async with semaphore:
             return await coro
     
     sema_coros = [sema_coro(c) for c in coros]
-    return asyncio.as_completed(sema_coros, **loop_arg_dict)
+    return asyncio.as_completed(sema_coros, **loop_arg_dict)  # type: ignore
 
 @dataclass
 class Executor:
--- a/src/ragas/executor.py
+++ b/src/ragas/executor.py
@@ -1,5 +1,4 @@
 from __future__ import annotations
-from concurrent.futures import Future
 import sys
 
 import asyncio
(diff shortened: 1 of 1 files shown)


src/ragas/executor.py
```diff
@@ -114,29 +48,53 @@ def submit(
...
+        self.jobs.append((callable_with_index, args, kwargs, name))
 
     def results(self) -> t.List[t.Any]:
-        executor_job = Runner(
-            jobs=self.jobs,
-            desc=self.desc,
-            keep_progress_bar=self.keep_progress_bar,
-            raise_exceptions=self.raise_exceptions,
-            run_config=self.run_config,
+        loop = asyncio.get_event_loop()
```
joy13975, line 54: [Use current loop if available](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_event_loop) so we're not creating unnecessary new loops that seems to cause tasks to be on different loops problem

rbellamy:
I tried the code here for `executor.py` and I still see:

```python
/Users/rbellamy/test/notebooks/notebook.py:361: RuntimeWarning: coroutine 'Executor._aresults' was never awaited
```

I believe Jupyter has an ever-running asyncio event loop, and that may be interfering.

src/ragas/executor.py
jjmachan: image
this is why I planned not to use `get_event_loop()` - would have been much easier if we could :(

what do you think?

joy13975: Oh I didn't know that.
With my shuffling of job creation vs. new loop order, I think theoretically `new_event_loop()` should also work.
Will test and report.

jjmachan: `new_event_loop()` works too and basically creates a new Thread I believe. running a new Thread/event loop is the cleanest implementation because if we reuse the same event loop in Jupiter notebook - it causes a lot of issues.

alternatively, what was the main problem with the current setup? if it's traces we can improve that part actually (traces and harder to debug were the points against this)

joy13975: Since I was getting #681 I thought the issue was that in the current setup futures were created before the the new loop was created, and later the semaphore refers to the new loop but futures (rather, ‘jobs’) were already attached to a old, different loop.

Furthermore, ‘Runner’ itself is a new thread, so doesn’t it mean the new loop exists on a new thread whereas the ‘jobs’ were created for on a loop in the main thread?

Please correct me if I’m wrong. The fact that sometimes the code works without this patch is a mystery to me.

jjmachan:
@joy13975 that is great to hear. Yes ragas is now model agnostic. you can checkout this docs for more info [Bring Your Own LLMs and Embeddings | Ragas](https://docs.ragas.io/en/stable/howtos/customisations/bring-your-own-llm-or-embs.html)

cheers 🙂

PR #714, opened 2024-03-06
Function _calculate_average_precision() modified to cover when LLM provides a bad JSON format

rodralez:
The modified version of the `_calculate_average_precision` function introduces several improvements and changes compared to the original version. Here's a detailed comparison highlighting these changes:

1. **Debugging Support with Processed Responses**: 
   - **Original**: Directly processes the `json_responses` list within the list comprehensions, with no provision for debugging or inspecting the processed inputs.
   - **Modified**: Introduces `processed_json_responses`, a separate list to store processed responses. This change aids in debugging by keeping a record of processed inputs, making it easier to understand how each input was handled.

2. **Enhanced Error Handling with Print Statements**:
   - **Original**: No explicit error messages or logging for specific cases like non-dict items or missing verdict keys within the input dictionaries.
   - **Modified**: Adds print statements to log warnings when non-dict items are encountered or when a verdict is missing from a response. These enhancements improve the function's usability by providing immediate feedback on what might be wrong with the input data.

3. **Flexibility with Nested Verification Key**:
   - **Original**: Assumes a flat structure for the input dictionaries, directly looking for a `verdict` key at the top level.
   - **Modified**: Adds flexibility to handle a potentially nested structure by checking for a `verification` key, which, if present and is a dictionary, is used to extract the verdict information. This adjustment allows the function to work with a broader range of input formats. This modification was done to deal with Azure OpenAI responses.

4. **Robustness in Handling Invalid Verdict Values**:
   - **Original**: Only considers responses with a `verdict` value explicitly equal to "1" as valid, implicitly treating all other cases as invalid (resulting in a `np.nan`).
   - **Modified**: Explicitly checks if the `verdict` string is either "0" or "1", ensuring that only these values are considered valid and improving the handling of invalid data by appending `np.nan` for verdicts that do not match these criteria.

5. **Use of `np.nansum` for Numerator and Denominator Calculation**:
   - **Original**: Uses Python's built-in `sum` function which could lead to incorrect results if `np.nan` values are present in the `verdict_list`.
   - **Modified**: Replaces `sum` with `np.nansum` for calculating both the numerator and the denominator. This choice correctly handles `np.nan` values by ignoring them in the sum, thus ensuring that the calculation of the average precision score is not affected by missing data.

6. **Improved Numerator Calculation Logic**:
   - **Original**: Uses a comprehension inside a `sum` call without directly addressing potential issues with NaN values in the computation.
   - **Modified**: The use of `np.nansum` in the numerator calculation makes the approach more robust, correctly handling `np.nan` values during the computation process.

These improvements enhance the function's flexibility, error handling, and robustness, particularly in dealing with varying input formats and missing or invalid data.

### DIFF
--- a/src/ragas/metrics/_context_precision.py
+++ b/src/ragas/metrics/_context_precision.py
@@ -92,23 +92,39 @@ class ContextPrecision(MetricWithLLM):
 
     def _calculate_average_precision(self, json_responses: t.List[t.Dict]) -> float:
         score = np.nan
-        json_responses = [
-            item if isinstance(item, dict) else {} for item in json_responses
-        ]
-        verdict_list = [
-            int("1" == resp.get("verdict", "").strip())
-            if resp.get("verdict")
-            else np.nan
-            for resp in json_responses
-        ]
-        denominator = sum(verdict_list) + 1e-10
-        numerator = sum(
+        processed_json_responses = []  # To store processed responses for debugging
+        for item in json_responses:
+            if isinstance(item, dict):
+                processed_json_responses.append(item)
+            else:
+                processed_json_responses.append({})
+                print("context_precision: Non-dict item found, replacing with empty dict.")  # Handle non-dict items
+
+        verdict_list = []
+        for resp in processed_json_responses:
+            # Adjusted logic to handle both formats
+            if "verification" in resp and isinstance(resp["verification"], dict):
+                verdict_info = resp["verification"]
+            else:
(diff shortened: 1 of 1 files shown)


shahules786:
Hey @rodralez thanks for the PR, will review it by weekend :)

jjmachan:
hey @rodralez with 0.2 we have made some improvements to this - have you checked it out?

PR #725, opened 2024-03-08
updated context_scoring_prompt and provide extra examples.

nofrish:
update context_scoring_prompt to better prompt and provide two extra examples.

### DIFF
--- a/src/ragas/testset/prompts.py
+++ b/src/ragas/testset/prompts.py
@@ -132,14 +132,25 @@ conversational_question_prompt = Prompt(
 
 context_scoring_prompt = Prompt(
     name="score_context",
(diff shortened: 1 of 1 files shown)


shahules786:
Hey @nofrish can you tell me why this is required?

nofrish:
hi, @shahules786 

First, the original prompt was inconsistent. It stated "complete the TWO following tasks" at the top, but there was actually only one task below. In the process of fixing this issue, I tried to standardize the prompt to make it clearer (this is not required, but I think it's more clearer).
I also added two examples to help the larger model make better judgments.

shahules786:
got it @nofrish thanks

PR #759, opened 2024-03-15
fix: Recursive Key Retrieval to Handle Unspecified JSON Schema Types and Fix for Int Value Returns about context_precision and context_recall

i-w-a:
### Summary
While utilizing the OpenAI API's GPT-4 turbo, I encountered an issue where JSON responses were returning with schema types that were not predefined. Additionally, despite specifying values as strings, some keys were returning integers. This pull request introduces a solution for both of these issues.

### Changes
- Implemented a recursive key retrieval function that can handle JSON objects of any structure, ensuring robust parsing regardless of the schema.
- Added a type-checking mechanism that converts integers back to strings if the original schema specified string types for key values.

### Rationale
The flexibility of the JSON response format from the GPT-4 turbo endpoint requires a more dynamic approach to parsing. The recursive method ensures that our application can seamlessly handle any JSON structure without the need for predefined schemas. Furthermore, the integrity of the data types is crucial for downstream processes, hence the necessity for a type-correction mechanism.

### Impact
This update will enhance the reliability of our system when interfacing with the OpenAI API and provide a more consistent data handling experience, especially in cases where the API's response structure is unpredictable.

Please review the changes and provide your feedback.

Thank you!

i-w-a

### DIFF
--- a/src/ragas/metrics/_context_precision.py
+++ b/src/ragas/metrics/_context_precision.py
@@ -90,21 +90,42 @@ class ContextPrecision(MetricWithLLM):
             for c in contexts
         ]
 
+    def _extract_verdict_value(self, item):
+        """get 'verdict' recursively"""
+        if isinstance(item, dict):
+            if "verdict" in item:
+                verdict_value = item["verdict"]
+                if isinstance(verdict_value, int):
+                    return 1 if verdict_value == 1 else 0
+                elif isinstance(verdict_value, str):
+                    verdict = verdict_value.strip()
+                    return 1 if verdict == "1" else 0
+            for key in item:
+                result = self._extract_verdict_value(item[key])
+                if result is not None:
+                    return result
+        elif isinstance(item, list):
+            for subitem in item:
+                result = self._extract_verdict_value(subitem)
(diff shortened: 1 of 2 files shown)


shahules786:
Thanks buddy @i-w-a lemme check this out.

shahules786:
Hey bro @i-w-a , your solution is interesting. I'll share my thoughts on it
the fundamental problem here is related to the parsing and structuring of the output format, which we think should be solved by as mentioned in this issue #761 

This is a good temporary solution but we would love to solve the fundamental issue and have a permanent solution. 
What do you think?

i-w-a:
Hi, @shahules786 
Thank you for your comments!

Agreed, issue #761 is crucial and needs a long-term solution. I support integrating Langchain's outputparser for a more sustainable fix. It should greatly improve our parsing process.

However, I'm concerned about the potential for increased API calls and costs due to the outputparser's schema validation. We should be mindful of this in our implementation.

Thoughts?

jjmachan:
hey @i-w-a we did make improvements, especially about the cost part, did you get a chance to try it out?

Issue #625, opened 2024-02-18
Testset generation. ValueError: a cannot be empty unless no samples are taken


Sansan4ez:
**Describe the bug**
ValueError: a cannot be empty unless no samples are taken

Ragas version: 0.1.1
Python version: 3.10.12

**Code to Reproduce**
```
from ragas.testset import TestsetGenerator  
from ragas.testset.evolutions import simple, reasoning, multi_context

distributions = {  
  simple: 0.5,  
  multi_context: 0.25,  
  reasoning: 0.25  
}

testsetgenerator = TestsetGenerator.with_openai(
  generator_llm = 'gpt-3.5-turbo-0125',  
  critic_llm = 'gpt-3.5-turbo-0125',
  embeddings = "text-embedding-3-small",
  chunk_size = 512,
)

testsetgenerator.adapt(language='russian', evolutions=distributions)

testset = testsetgenerator.generate_with_langchain_docs(test_docs, test_size=40, distributions=distributions, raise_exceptions=False)
```

**Error trace**
```
ERROR:ragas.executor:Runner in Executor raised an exception
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/dist-packages/ragas/executor.py", line 58, in _aresults
    r = await future
  File "/usr/lib/python3.10/asyncio/tasks.py", line 571, in _wait_for_one
    return f.result()  # May raise f.exception().
  File "/usr/local/lib/python3.10/dist-packages/ragas/executor.py", line 91, in wrapped_callable_async
    return counter, await callable(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/ragas/testset/evolutions.py", line 161, in evolve
    ) = await self._aevolve(current_tries, current_nodes)
  File "/usr/local/lib/python3.10/dist-packages/ragas/testset/evolutions.py", line 272, in _aevolve
    nodes = self.docstore.get_random_nodes(k=1)
  File "/usr/local/lib/python3.10/dist-packages/ragas/testset/docstore.py", line 274, in get_random_nodes
    return rng.choice(np.array(self.nodes), size=k).tolist()
  File "_generator.pyx", line 803, in numpy.random._generator.Generator.choice
ValueError: a cannot be empty unless no samples are taken
```

**Expected behavior**
generation of test set from documents in russian language.

**Additional context**
--

shahules786:
Hey @Sansan4ez can you share how many documents did you pass?

shahules786:
[...]

PR #766, opened 2024-03-17
Fix adapt issue for keyphrase extractor

smodlich:
Hi,
This fixes [issue #625](https://github.com/explodinggradients/ragas/issues/625). This issue is happening when after adaption testset generation is called. This issue is caused by the fact, that after adaption the keyword extractor returns nothing. This leads to failure later.  However the keywords are indeed extracted correctly (as can be seen when debugging) but not returned properly from the function. The key "topics" doesn't exist (after adaption) in the dict, instead it's "keyphrases". This PR works for both with/without adaption.

### DIFF
--- a/src/ragas/testset/evolutions.py
+++ b/src/ragas/testset/evolutions.py
@@ -296,7 +296,7 @@ class SimpleEvolution(Evolution):
         results = await self.generator_llm.generate(
             prompt=self.seed_question_prompt.format(
                 context=merged_node.page_content,
-                topic=rng.choice(np.array(merged_node.keyphrases), size=1)[0],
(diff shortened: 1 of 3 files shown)


shahules786:
Hey @smodlich bro, thanks for the PR. I just did look into this issue. The issue happened because I updated the prompt key from `keyphrases` to `topics` if you can see. Now your adapt is loading the older version of the prompt from your cache, to fix this just delete the prompt saved in the cache location `/.cache/ragas/<your-langugae>/keyphrase_extraction.json` and redo adapt.  For a quick fix, I will change the name of the prompt too.

PR #771, opened 2024-03-18
auto fix some json decoder errors

ZeyuTeng96:
Hi there, 

When I used the 'context_recall' metric, I constantly got json decoder error of 'Invalid \escape'. And the llm is fail to generate a propriate json. So, I just think can we auto fix some types of decoder error without calling llm. Therefore, I provide two fail cases and fix it by simply replacing some chars. It can also extend more decoder errors in the future.


jjmachan:
this has been fixed with v0.2

PR #776, opened 2024-03-18
Fix: json format inconsistency in context recall (prompt for statemen…

jokokojote:
There is a small inconsistency in the prompt for statement creation / attribution for the context recall metric. The output key `classification` is expected to be an array of attributed statements (one or more statements). This is defined correctly in the first example, but in the two following examples the array brackets are missing. 

This inconsistency may not lead to any problems when working with powerful llms (e.g. with OpenAI) most of the time, but I noticed some confusion about the expected output when running the original code with some other smaller models. This way it is more clear and consistent, that one **or more** statements are expected to be returned.


shahules786:
Makes sense @jokokojote

PR #782, opened 2024-03-19
Added metadata field to datarows

rnbokade:
This would help have a better picture of the documents that were fetched. Users can choose to drop this column if not required. But in most cases, this would be a good addition at least in the initial stages for verifying the quality of the dataset and understanding the otherwise black-box nature of the dataset generation here.


shahules786:
Nice touch @rnbokade

PR #796, opened 2024-03-22
fix: Eventloop unclosed

yunfwe:
When I call ragas in the fastapi, every time I call it, python will have an open event loop (/proc/self/fd).


jjmachan (approved):
thanks a lot for catching this and fixing it ❤️

PR #801, opened 2024-03-24
fix: Explanaitions typo on index.md

dcamejo1:
Noticed this typo and figured I'd offer to fix it

image
image


shahules786:
Nice catch @dcamejo1

Issue #761, opened 2024-03-15
More robust JSON prompting and parsing


mrtj:
**Describe the Feature**

This feature request is about integrating LangChain's `PydanticOutputParser` and pydantic models into ragas prompting.

**Why is the feature important for you?**

The current prompting of the validation metrics uses a somewhat vague definition of the answer format, specifying only examples about the expected format. The LLM then should deduct the expected format and generate an answer that corresponds to the schema that is only implicitly specified in the single metrics implementation. Especially when using different LLMs compared to the default GPT-3.5, this leads to potential parsing errors and the metrics can not be calculated correctly.

For example:

- Claude and other models tends to return the binary verdict as a JSON number opposed to the expected string: [#752](#752) (this issue is about testset generation, but I saw very similar issues also during context precision/recall calculation) and also [#715](#715)
- Sometimes the "verdict" envelop is omitted from the response: [#733](#733)
- Sometimes the response is embedded in a superfluous envelope: [#668](#668)
- It seems different models have issues with the "Attributed" keys: [#619](#619)
- ...and numerous other bugs might be related to the weak JSON parsing.

**Additional context**

LangChain has a [robust implementation](https://python.langchain.com/docs/modules/model_io/output_parsers/types/pydantic) of instructing the models to return a JSON response conforming to a specific schema. The output format can be specified with pydantic data classes, the expected JSON schema is injected into the prompt, and the response is automatically parsed by pydantic. Additionally, a [retry mechanism](https://python.langchain.com/docs/modules/model_io/output_parsers/types/retry) can be included for even more robust JSON parsing.

Considering that Ragas is already using LangChain, implementing this feature would not create addition dependencies to the project.

mrtj:
As implementing this enhancement would have a certain impact on the codebase, I opened this issue to start a discussion about this idea. If you give green light to this feature, I would be happy to contribute to the implementation.

shahules786:
Hey @mrtj this is certainly something we have in the roadmap. We welcome any help on this.

shahules786:
Lemme know if you're working on it and I'll assign it to you, we can also help you with it.

mrtj:
[...]

PR #807, opened 2024-03-26
More robust JSON prompting and parsing

mrtj:
This PR implements a langchain-style, hopefully more robust output parsing as discussed in #761 .


mrtj:
Currently I added a `use_langchain_parser` configuration attribute to each supported metrics. Settings this to `true` will use the JSON output parser implementation directly from LangChain, but in this case the retry mechanism is not (yet) implemented. Setting this to `false` will use the original JSON parsing from ragas, and verify the extracted JSON object with pydantic. 

In both cases the output instructions and the expected JSON schema is injected in the prompts, and the models tend to follow correctly these instructions and generate a JSON output conforming to the schema.

I've tested the behavior with gpt-3.5-turbo, gpt-4-turbo, Claude v2, v2.1, v3 Sonnet and Claude Instant v1 (via Amazon Bedrock), and each models usually return a correct JSON format.

shahules786:
Hey @mrtj perfect timing. I was collecting issues related to the same and thinking of your work + how I can help. 
I'll take a look at this today. Thanks a lot :)

shahules786:
Just took a detailed look at it. Great work, exactly what we needed. Some todo's for me

- [ ] Implement RetryWithErrorOutputParser
- [ ] Remove `json_load` if not required
- [ ] Convert all prompt (testset) outputs to same pydantic model
- [ ] Make sure prompt adaptation is working 
- [ ] Minor formating changes

shahules786:
Hey @mrtj I have made some changes. would love to have your opinion on it as well if any.

shahules786 (approved):

mrtj:
Refactoring the try..catch blocks around the output parsing is an idea that I was hesitating how to implement the best: having a custom subclass of pydantic output parser is certainly the most elegant way, I like it very much.

The other thing I wanted to ask: could you explain me why do you think that asking a binary integer (0/1) response from the model is better than using a boolean directly? My experience with LLMs suggest that usually it is a good idea to restrict the grade of liberty of the models about the response format. This is the reason I've converted these variables to boolean. But maybe you have some other reasons?

shahules786:
Hey @mrtj thank you.

The reason for converting verdicts to boolean is to make it language agnostic - ie, during prompt adaptation we convert a few shot examples to the target language. So in this case the verdict should stay the same irrespective of the language.
https://docs.ragas.io/en/stable/howtos/applications/use_prompt_adaptation.html

Any better idea to do the same?

src/ragas/llms/prompt.py
```diff
@@ -136,7 +147,7 @@ def get_example_str(self, example_no: int) -> str:
                 else value
             )
             example_str += f"\n{key}: {value}"
-        return example_str
+        return "```" + example_str + "```"
```
jjmachan, line 150: why do we use the triple backticks ?
other models seem to mess this up by adding extra tokens like ```bash etc

jjmachan (approved):
hey @mrtj thanks a lot for this PR ❤️ 

I tested this will a few other OSS models and Claude and there are improvements there as well. We will have to write some tests for JSON parsing across OSS models I think so that we keep track of model changes and their support for JSON. I'll work on those


But this improves the major issue the community are facing today and Thanks a lot for contributing this 🙂

larunach:
> Currently I added a `use_langchain_parser` configuration attribute to each supported metrics. Setting this to `true` will use the JSON output parser implementation directly from LangChain, but in this case the retry mechanism is not (yet) implemented. Setting this to `false` will use the original JSON parsing from ragas, and verify the extracted JSON object with pydantic.
> 
> In both cases the output instructions and the expected JSON schema are injected in the prompts, and the models tend to follow correctly these instructions and generate a JSON output conforming to the schema.
> 
> I've tested the behavior with gpt-3.5-turbo, gpt-4-turbo, Claude v2, v2.1, v3 Sonnet and Claude Instant v1 (via Amazon Bedrock), and each models usually return a correct JSON format.

I am using mistral model via hugging face and it returns output in this format, still it is not able to parse the output
 ```
{
    "reason": "The provided context was useful in arriving at the given answer. The context explicitly states that the capital of India is Delhi, which matches the answer.",
    "verdict": 1
}
```
What is the issue with this  ?

File: src/ragas/llms/prompt.py (hunk 1 above)
Comment (mrtj) [reply src/ragas/llms/prompt.py:150]: if I am not mistaken, the json/pydantic response parser in langchain search for the triple backsticks in the response to identify the json structure in the sometimes more verbose response of the model. Some models love to add some free text preamble before the json block, or conclusion/explanation of their work done after the json.

File: docs/getstarted/evaluation.md [hunk 2]
```diff
@@ -3,10 +3,6 @@
 
 Once your test set is ready (whether you've created your own or used the [synthetic test set generation module](get-started-testset-generation)), it's time to evaluate your RAG pipeline. This guide assists you in setting up Ragas as quickly as possible, enabling you to focus on enhancing your Retrieval Augmented Generation pipelines while this library ensures that your modifications are improving the entire pipeline.
 
-<p align="left">
```
mtharrison, line 6: Just curious, what was the reason for removing the diagram?

PR #837, opened 2024-04-03
fix: change documentation to reflect accurate method name

dbtreasure:
Updates the docs to reflect the correct method name.


jjmachan:
thanks a lot for catching @dbtreasure

dbtreasure:
There are a few more that I've caught since and I can make some more PRs. Thanks for merging!

PR #847, opened 2024-04-05
minor: fix llm expansion in docs

glenacota:
Updates the docs to fix the LLM acronym expansion.


shahules786:
Thanks buddy @glenacota nice catch.

PR #849, opened 2024-04-08
Update llama_index imports

sestinj:
Going through the tutorial I found that llama_index was restructured recently (in the last month?) so these imports had to be changed to work correctly


shahules786:
Thanks buddy @sestinj this is helpful :)

Issue #286, opened 2023-11-15
datasets.features.features.Sequence


samedovzaur:
**Describe the bug**
I'm getting this error either with my own dataset or your dataset in the documentation, it keeps giving this error:
ValueError: Dataset feature "contexts" should be of type Sequence[string], got <class 'datasets.features.features.Sequence'>

Ragas version:
Python version:

**Code to Reproduce**
from ragas import evaluate

result = evaluate(
    fiqa_eval["baseline"],
    metrics=metrics,
)

result
Copy code

**Error trace**

**Expected behavior**

**Additional context**

shahules786:
Hey, can you share the Ragas and Python versions?

samedovzaur:
Ragas: Version: 0.0.20
Openai: Version: 1.3.0

samedovzaur:
> Hey, can you share the Ragas and Python versions? and make sure it's the latest

Now I'm encountering this error:
NotFoundError: Error code: 404 - {'error': {'code': '404', 'message': 'Resource not found'}}

While evaluating relevancy score:

evaluating with [faithfulness]
100%|██████████| 7/7 [09:02<00:00, 77.46s/it]
evaluating with [answer_relevancy]
  0%|          | 0/7 [00:07<?, ?it/s]
---------------------------------------------------------------------------
NotFoundError                             Traceback (most recent call last)
[<ipython-input-149-c87895f9a55b>](https://localhost:8080/#) in <cell line: 3>()
      1 from ragas import evaluate
      2 
----> 3 result = evaluate(fiqa_eval['train'], metrics=metrics)

11 frames
[/usr/local/lib/python3.10/dist-packages/openai/_base_client.py](https://localhost:8080/#) in _request(self, cast_to, options, remaining_retries, stream, stream_cls)
    875             # to completion before attempting to access the response text.
    876             err.response.read()
--> 877             raise self._make_status_error_from_response(err.response) from None
    878         except httpx.TimeoutException as err:
    879             if retries > 0:

NotFoundError: Error code: 404 - {'error': {'code': '404', 'message': 'Resource not found'}}

guankaisi:
the same problem, so how to fix it?

samedovzaur:
> the same problem, so how to fix it?

which problem is it u are considering? last one or the one at the top? (second one is embedding problem, I used opensource embedding(jina-ai) it worked but for prod I need openai embedding to work out.
if it 1st one let me know, I have a solution.

[...]

PR #879, opened 2024-04-19
Update validation to allow for empty context lists

LogicalShark:
When no strings are provided in the contexts during evaluation, the context feature's type in the dataset is the generic Sequence type rather than the expected Sequence[string], thus it fails validation with a ValueError, despite being a valid dataset.

For example, if retrieval uses an unrealistically high similarity score threshold, no contexts will be found, but the error message will be the unhelpful and irrelevant `ValueError: Dataset feature "contexts" should be of type Sequence[string], got <class 'datasets.features.features.Sequence'>`.

See also https://github.com/explodinggradients/ragas/issues/286

### DIFF
--- a/src/ragas/validation.py
+++ b/src/ragas/validation.py
@@ -51,7 +51,7 @@ def validate_column_dtypes(ds: Dataset):
         if column_names in ds.features:
             if not (
                 isinstance(ds.features[column_names], Sequence)
-                and ds.features[column_names].feature.dtype == "string"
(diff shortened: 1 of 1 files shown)


shahules786:
Hey @LogicalShark thanks for the PR, looks good.

jjmachan:
hey @LogicalShark this has been fixed with v0.2, do check that out and I'm really sorry we couldn't merge your PR but thanks a lot for taking the time ❤️

PR #884, opened 2024-04-20
fix: replace assert with ValueError for llm validation

iw4p:
It's absolutely better to use raise ValueError instead of assert. (or maybe someone has a better idea)
If you agree for the next refactors I can implement this method
```
def check_llm(self):
    if self.llm is None:
        raise ValueError("LLM must be set before using this method")
```
So by using this method whenever we need to use llm, we can easily call it before that line like
`
self.check_llm()
`


jjmachan:
thanks a lot for the suggestion and the PR 🙂 ❤️. hopefully with #1558 we will be able to fix this but keeping it for v0.3 so as to not break anything

Issue #888, opened 2024-04-21
[R-230] Adapt prompts "in place" also for cached prompts


HerrIvan:
**Describe the Feature**
Adapt Prompts "in place", also in case they are retrieved from the cache.

**Why is the feature important for you?**
If you want to generate a testset in a non-english language, the right approach is tofirst adapt all the generation/filtering prompts and then run the generation. 
However, this works only if you adapt the prompts the first time. In that case, the prompt instances get changed "in place", that means that when the `testset.generate` accesses the prompts, these prompts are the translated ones. However, if the prompts are retrieved from the cache, they are not changed "in place", which means that when TestsetGenerator.generate accesses the prompts, it uses the original ones in English.

**Additional context**
This is half a feature request, half a bug, since the behavior of the "adapt" function is not consistent: Prompts in the cache are not modified "in place", while instances of prompts not in the cache are modified.

See below in `src/ragas/llms/prompt.py`, when a prompt is "translated/adapted" the translated values are stored in the prompt instance. On top of that, the instance is returned.

```python
        for i, example in enumerate(grouped_results):
            [...]
            self.examples[i] = example_dict

        self.language = language

        # TODO:Validate the prompt after adaptation

        return self
```

However, in the same function, whenever there is a hit in the `cache` the current prompt instance isn't modified (ie, nothing is assigned to the `self` var:

```python
        if os.path.exists(os.path.join(cache_dir, language, f"{self.name}.json")):
            return self._load(language, self.name, cache_dir)
```

(The function `self._load` is a class method which doesn't modify the instance).

I will open a PR to address this issue.

<sub>[R-230](https://linear.app/exploding-gradients/issue/R-230/adapt-prompts-in-place-also-for-cached-prompts)</sub>

shahules786:
Hey @HerrIvan thanks for pointing this out. Seems like a bug to me.

jjmachan:
also tagging #890

jjmachan:
This has been fixed with v0.2 - I know finally 😅 🎉

do checkout the docs here: https://docs.ragas.io/en/stable/howtos/customizations/metrics/_metrics_language_adaptation/
reference here: https://docs.ragas.io/en/stable/references/prompt/#ragas.prompt.PromptMixin

and if you're migrating from v0.1 check out the migration docs here: https://docs.ragas.io/en/stable/howtos/migrations/migrate_from_v01_to_v02

[...]

PR #889, opened 2024-04-21
Fix: `adapt` function to translate cached `prompts` in place.

HerrIvan:
Address issue #888

It enables the following functionality in order to generate a testset in a language other than English.

```python

from ragas.testset.prompts import testset_prompts
from ragas.testset.generator import TestsetGenerator

...
# some other imports 
...

# adapts all prompts to a given language, here german
for p in testset_prompts:
    p.adapt("german", translator_llm)

# run the test set generation
generator = TestsetGenerator(
        docstore=docstore,
        generator_llm=generator_llm,
        critic_llm=critic_llm,
        embeddings=embeddings
    )

dataset = generator.generate(
        test_size=TEST_SIZE,
        distributions=DISTRIBUTION
)
```

Without this fix, this would only work the first time. The first time, when the prompts are translated the "adapt" method changes the prompt instances (that is, in place). Therefore, the generate below uses the translated prompts. However once the prompts are in the cache, running adapt on all prompts doesn't change them "in place" and then `generate` will access the original prompts completely in English. This PR makes this approach to work also when the translated prompts are already cached.

I also added the `testset_prompts` variable for convenience.


jjmachan:
thanks alot for the PR @HerrIvan ❤️ 

I know you pushed this a while back, really sorry for not merging it in earlier. but would love to catchup with you sometime if you are interested. Just to meet you 🙂

Issue #804, opened 2024-03-26
AssertionError: Adapted output keys set(output.keys())=set() do not match with the original output keys: output_keys[i]={'statements'}


Ailce8862:
[ ] I have checked the [documentation](https://docs.ragas.io/) and related resources and couldn't resolve my bug.

**Describe the bug**
AssertionError: Adapted output keys set(output.keys())=set() do not match with the original output keys: output_keys[i]={'statements'}

Ragas version:0.1.5
Python version:3.10.13 

**Code to Reproduce**
from ragas import adapt
    from ragas.metrics import (
        faithfulness,
        answer_relevancy,
        context_recall,
    )
    from langchain_community.embeddings import QianfanEmbeddingsEndpoint
    from langchain_community.chat_models import QianfanChatEndpoint

    embed_llm = QianfanEmbeddingsEndpoint(
        qianfan_ak='XXX',
        qianfan_sk='XXX'
    )
    chat_llm = QianfanChatEndpoint(streaming=True)

    adapt(metrics=[faithfulness, answer_relevancy,context_recall], language="Chinese", llm=chat_llm)
    print(faithfulness.long_form_answer_prompt.to_string())

**Error trace**
Traceback (most recent call last):
  File "/Users/liangpan/工作/新点/coding/RAG_langchain/step3/RAGAS_eval_separation.py", line 133, in <module>
    adapt(metrics=[faithfulness, answer_relevancy,context_recall], language="Chinese", llm=chat_llm)
  File "/Users/liangpan/miniforge3/envs/langchain/lib/python3.10/site-packages/ragas/adaptation.py", line 36, in adapt
    metric.adapt(language, cache_dir=cache_dir)
  File "/Users/liangpan/miniforge3/envs/langchain/lib/python3.10/site-packages/ragas/metrics/_faithfulness.py", line 203, in adapt
    self.long_form_answer_prompt = self.long_form_answer_prompt.adapt(
  File "/Users/liangpan/miniforge3/envs/langchain/lib/python3.10/site-packages/ragas/llms/prompt.py", line 231, in adapt
    set(output.keys()) == output_keys[i]
AssertionError: Adapted output keys set(output.keys())=set() do not match with the original output keys: output_keys[i]={'statements'}

Process finished with exit code 1

LukaszDejneka:
Same issue for trying to adapt to Polish language.

jjmachan:
tagging #890 as a meta issue to cleanup everything

[...]

Issue #774, opened 2024-03-18
Automatic language adaption TestSet generation error: "Adapted output keys do not match with the original output keys"


baswenneker:
**Describe the bug**
Can't get the automatic language adaption going for testset generation. I retried this about 10 times.

Ragas version: 0.1.4
Python version: 3.11

**Code to Reproduce**
Share code to reproduce the issue
```
from ragas.testset.generator import TestsetGenerator
from ragas.testset.evolutions import simple, reasoning, multi_context,conditional
from langchain_openai import ChatOpenAI, OpenAIEmbeddings

# generator with openai models
generator_llm = azure_llm()
critic_llm = azure_llm()
embeddings = azure_embeddings()

generator = TestsetGenerator.from_langchain(
    generator_llm,
    critic_llm,
    embeddings
)

# adapt to language
language = "Dutch"
cache_dir = ".cache"

generator.adapt(language, evolutions=[simple], cache_dir=cache_dir)
generator.save(evolutions=[simple, reasoning, multi_context, conditional], cache_dir=cache_dir)
```

This is the output until it errors out:
```
{'keyphrases': ['Zwart gat', 'Regio van ruimtetijd', 'Sterke zwaartekracht', 'Licht en elektromagnetische golven', 'Theorie van algemene relativiteit']}
{'keyphrases': ['Chinese Muur', 'Oude vestingwerken', 'Noord-China']}
{'answer': 'Menselijke activiteiten dragen voornamelijk bij aan klimaatverandering door de uitstoot van broeikasgassen bij het verbranden van fossiele brandstoffen. Deze uitstoot verhoogt de concentratie van broeikasgassen in de atmosfeer, wat meer warmte vasthoudt en leidt tot opwarming van de aarde en veranderde weerspatronen.', 'verdict': '1'}
{'answer': 'Kunstmatige intelligentie is ontworpen om menselijke cognitieve functies na te bootsen, met belangrijke capaciteiten zoals leren, redeneren, waarnemen en reageren op de omgeving op een manier die vergelijkbaar is met mensen. Deze capaciteiten maken AI cruciaal in verschillende velden, inclusief gezondheidszorg en autonoom rijden.', 'verdict': '1'}
{'answer': 'Het antwoord op de gegeven vraag is niet aanwezig in de context', 'verdict': '-1'}
{'relevant_contexts': [1, 2]}
[[1, 2], {'relevant_contexts': [1, 2]}]
{'score': 6.0}
[...]

PR #891, opened 2024-04-22
fix: Adapted output keys do not match with the original output keys

FeU-aKlos:
this should fix the error mentioned [here](https://github.com/explodinggradients/ragas/issues/774) and [here](https://github.com/explodinggradients/ragas/issues/804) 

the error is thrown due to an example with an empty list of statements.

### DIFF
--- a/src/ragas/llms/base.py
+++ b/src/ragas/llms/base.py
@@ -7,7 +7,7 @@ from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from functools import partial
 
-from langchain_community.chat_models import ChatVertexAI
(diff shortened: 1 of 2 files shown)


shahules786:
Hey @FeU-aKlos thanks for the PR, will check it out soon.

LukaszDejneka:
Hey @shahules786 any idea when you could take a look? Without adapting the prompts to specific language, prompts for metrics written in default EN are unreliable for languages such as Polish.

jjmachan:
we fixing this in v0.2 - could you check it out

Thanks a lot for putting in the PR and putting in the time to fix it but closing this now. 

do try out the latest version and if there is anything just comment down here 🙂

PR #894, opened 2024-04-23
Fix a few typos in the docs

amirma:


jjmachan:
hey @amirma thanks for taking the time to fix these, should be addressed in v0.2. closing this for now

PR #921, opened 2024-04-30
Cleaned installation step in README

andreped:
I saw that the installation step from source was unnecessarily complicated in the README.

I simplified it, following patterns using in most other large python package frameworks.

I also changed the text around to make it clearer to the user what each step does.


jjmachan (commented):
thank you for fixing this 🙂

PR #924, opened 2024-04-30
reproducibility PR1 : introduce ensembler 

shahules786:
Changes

- [x] An ensemble to combine multiple outputs for the same query (binary)
- [x] introduce the concept of reproducibility to 3 metrics (self consistency)
- [x] Modify and break answer-correctness into 2 step process to improve stability and decrease errors.

#925

### DIFF
--- a/src/ragas/metrics/_context_precision.py
+++ b/src/ragas/metrics/_context_precision.py
@@ -10,7 +10,7 @@ from langchain.pydantic_v1 import BaseModel, Field
 
 from ragas.llms.output_parser import RagasoutputParser, get_json_format_instructions
 from ragas.llms.prompt import Prompt, PromptValue
-from ragas.metrics.base import EvaluationMode, MetricWithLLM
+from ragas.metrics.base import EvaluationMode, MetricWithLLM, ensembler
 
 if t.TYPE_CHECKING:
     from langchain_core.callbacks import Callbacks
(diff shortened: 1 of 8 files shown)


src/ragas/metrics/_context_precision.py
```diff
@@ -90,6 +90,18 @@ class ContextPrecision(MetricWithLLM):
     evaluation_mode: EvaluationMode = EvaluationMode.qcg  # type: ignore
     context_precision_prompt: Prompt = field(default_factory=lambda: CONTEXT_PRECISION)
     max_retries: int = 1
+    _reproducibility: int = 1
+
+    @property
+    def reproducibility(self):
+        return self._reproducibility
```
jjmachan, line 97: shouldn't this be part of `Metrics`?

src/ragas/metrics/base.py
```diff
@@ -156,3 +157,30 @@ def init(self, run_config: RunConfig):
...
+
+        verdict_agg = []
+        for i in range(len(inputs[0])):
+            item = inputs[0][i]
+            verdicts = [inputs[k][i][attribute] for k in range(len(inputs))]
+            verdict_counts = dict(Counter(verdicts).most_common())
+            item[attribute] = list(verdict_counts.keys())[0]
+            verdict_agg.append(item)
+
+        return verdict_agg
```
jjmachan, line 183: how does this work?

src/ragas/metrics/_context_precision.py
jjmachan: The question you have to ask is - does every `Metric` or `MetricWithLLM` or `MetricWithEmbedding` need this?

and you only have to ask the question if you have to do this 2+ times - which is this case you did. so what was the consideration here?

jjmachan (changes requested):

src/ragas/metrics/_context_precision.py
shahules786: I did think about this. For now, only 3/6 of the metrics can use this technique. I tried multiple ways to combine generative o/p too for improved reproducibility but I didn't get satisfactory results to raise a PR.  So, given that only 3/6 metrics have this wouldn't it be okay to keep it here now?

src/ragas/metrics/base.py
shahules786: I just ensembles multiple outputs. 
For example if results is [0,1,1] over 3 llm calls for the query, output will be 1 which is the majority vote. I will add a doc for this.

jjmachan (approved):

PR #937, opened 2024-05-07
Fix testset generator issue on context selection

Jovillios:
This pull request addresses an issue in the TestsetGenerator where the chosen contexts for evolutions were not being sampled correctly because only the very first contexts were selected. Now, the selection can potentially choose any context


shahules786:
Nice catch @Jovillios thanks for the PR buddy.

PR #958, opened 2024-05-15
fix - typo & prompt

omkar-334:
1. Function typo 
Original name - set_node_relataionships()
Fixed name - set_node_relationships()

2. In the seed_question_prompt, a keyphrase is given.   
But in the instruction, it is mentioned that `The question should be formed using topic`.   
image  
`The question should be formed using the given keyphrase.` would be better


jjmachan:
hey @omkar-334, we have shifted to a new approach of storing documents with v0.2, do check it out

PR #961, opened 2024-05-16
fix typo in f1 score

robuno:
In documents, the absolute value sign was missing in the equation of F1 score.


omkar-334 (approved):

robuno (commented):
An usual typo correction.

Issue #963, opened 2024-05-16
TestsetGenerator -> RuntimeError: ... got Future <..> attached to a differen t loop


abetatos:
This is related to: https://github.com/explodinggradients/ragas/issues/681

**Describe the bug**
TestGeneration launches an error with default example of https://docs.ragas.io/en/stable/concepts/testset_generation.html with VertexAI models. 

Ragas version: 0.1.7
Python version: Python 3.10.13

**Code to Reproduce**
```python
generator_llm = VertexAI(model_name="text-bison", temperature=0, response_validation=False)
critic_llm = VertexAI(model_name="gemini-pro", temperature=0, response_validation=False)
vertexai_embeddings = VertexAIEmbeddings("textembedding-gecko@001")

generator = TestsetGenerator.from_langchain(
    generator_llm,
    critic_llm,
    vertexai_embeddings
)

distributions = {
    simple: 0.5,
    multi_context: 0.4,
    reasoning: 0.1
}

loader = PubMedLoader("liver", load_max_docs=10)
documents = loader.load()

testset = generator.generate_with_langchain_docs(documents, 10, distributions, with_debugging_logs=True) 
testset.to_pandas()
```
**Error trace**

```python
Filename and doc_id are the same for all nodes.                                                                                                                        
Generating:   0%|                                                                                                                               | 0[/10](http://localhost:8889/10) [00:00<?, ?it[/s](http://localhost:8889/s)][ragas.testset.filters.DEBUG] context scoring: {'clarity': 3, 'depth': 3, 'structure': 3, 'relevance': 3, 'score': 3.0}
[ragas.testset.evolutions.DEBUG] keyphrases in merged node: ['Hyperuricemia', 'Protein peptides', 'Anti-hyperuricemia properties', 'Uric acid-lowering peptides', 'Functional foods']
[ragas.testset.filters.DEBUG] context scoring: {'clarity': 3, 'depth': 3, 'structure': 3, 'relevance': 3, 'score': 3.0}
[ragas.testset.evolutions.DEBUG] keyphrases in merged node: ['Liver surgery', 'Liver anatomy and physiology', 'Liver resection', 'Vascular control', 'Liver transplantation', 'Residual liver function']
[ragas.testset.filters.DEBUG] context scoring: {'clarity': 2, 'depth': 2, 'structure': 2, 'relevance': 3, 'score': 2.25}
[...]

PR #979, opened 2024-05-21
Chore: Fix RuntimeError: ... got Future <..> attached to a different loop

abetatos:
In issue #963 I commented about the RuntimeError when trying to use TestsetGenerator. 

In issue  #681 you discussed that using get_event_loop() was the optimal way to solve the issue but it had a Deprecation warning. 

image

In newer python versions get_event_loop() raises a RuntimeError when no event loop is active. Thus we can avoid creating unnecessary new loops that seem to be the problem that was mentioned in both issues. 

I have modified the code the least possible to make it clear what I intended to, although in this setup runner could be avoided as was proposed in #689 

Setup: 
Ragas version: 0.1.7
Python version: Python 3.10.13

### DIFF
--- a/src/ragas/executor.py
+++ b/src/ragas/executor.py
@@ -58,7 +58,10 @@ class Runner(threading.Thread):
         self.run_config = run_config or RunConfig()
 
         # create task
-        self.loop = asyncio.new_event_loop()
+        try:
+            self.loop = asyncio.get_event_loop()
+        except RuntimeError:
+            self.loop = asyncio.new_event_loop()
         self.futures = as_completed(
             loop=self.loop,
             coros=[coro for coro, _ in self.jobs],
@@ -96,7 +99,6 @@ class Runner(threading.Thread):
             results = self.loop.run_until_complete(self._aresults())
         finally:
             self.results = results
-            self.loop.stop()
 
 
 @dataclass


src/ragas/executor.py
```diff
@@ -96,7 +99,6 @@ def run(self):
             results = self.loop.run_until_complete(self._aresults())
         finally:
             self.results = results
-            self.loop.stop()
```
jjmachan, line 99: why did you remove this?

jjmachan: Thanks a lot for putting the effort to fix this issue ❤️ 🙂

abetatos: Because if you use get_event_loop from a jupyter notebook you obtain the loop of the own jupyter, thus closing it might raise some problems. There might be another solution but I thought it was the most straightforward one.

abetatos: Thanks to you for creating such an awesome library!

rbellamy: > Because if you use get_event_loop from a jupyter notebook you obtain the loop of the own jupyter, thus closing it might raise some problems.

Thank you so much for this! 🙇‍♂️

jjmachan (approved):

src/ragas/executor.py
charlie-ang-collibra: Hi @abetatos .  Thanks for submitting the PR.  🙏  .   I tried to follow this issue but my knowledge of `asyncio` isn't very good.

PR #982, opened 2024-05-21
Improve JSON format prompt for large chunks & Handle ZeroDivisionError

Manav916:
## Description
In this PR there are three different changes
## Changes

1. Fixed a Typo in the filter_question_prompt Instruction from `Asses` to `Assess`
2. Added a try-except block for handling ZeroDivisionError for the filter method in the NodeFilter class
3. Improved the JSON_FORMAT_INSTRUCTIONS for better output generation
It seems like the LLM sometimes loses track of the imposed instruction for the output, especially for long prompts. So although the output generated for a chunk_size of 512 is perfect, the output for a chunk_size of 1024 has an extra newline. Here parsing fails when using PydanticOutputParser even though the llm has generated an output and there is only an extra `'\n'` as shown in the instances below.
Screenshot 2024-05-22 093242
But by tuning the prompt and adding `Please output your response in the demanded json format.` at the end of the instruction we get output without `'\n'`. This output can then be parsed and the context can be considered.
Screenshot 2024-05-22 093757


jjmachan:
@shahules786 could you check if we can merge this in?

src/ragas/testset/filters.py
```diff
@@ -57,7 +57,10 @@ async def filter(self, node: Node) -> t.Dict:
         output = results.generations[0][0].text.strip()
         output = await context_scoring_parser.aparse(output, prompt, self.llm)
         output = output.dict() if output is not None else {}
-        output["score"] = sum(output.values()) / len(output.values())
+        try:
+            output["score"] = sum(output.values()) / len(output.values())
+        except ZeroDivisionError:
+            output["score"] = 0  # or some default value or handling
```
mspronesti, line 63: Relying on exceptions for control flow is usually not efficient. Just use a third operator:

```suggestion
        vals = output.values()
        output["score"] = sum(vals) / n_vals if (n_vals := len(vals)) > 0 else 0
```

mspronesti (changes requested):

jjmachan:
hey @Manav916 we did make some improvements with the v0.2 release - do check it out

PR #990, opened 2024-05-26
Add initial version of summarization metric

sky-2002:
This PR adds an initial version of a metric for summarization use cases.

**Description**
For evaluating summaries generated by LLMs, we need a metric to consider overall informativeness and conciseness of the summary. This metric has been formulated to do the same.

**Formulation**
To determine if a summary is a good one or not, it is essential that it contains all important information that is present in the original text. We begin with generating `n` number of `'yes'/'no'` questions, for which the original text strictly has the answer `yes`.
We then ask those questions to the generated summary, and compute the `qa_score` as the fraction of correctly answered questions.

We then also take into account the length of the summary, because otherwise this metric would be biased towards summaries that are same as the original text, thus answering all questions correctly and obtaining a score of 1. So we provide an option to add a penalty based on the relative length of the summary.

For example, if the summary could answer `yes` to 3 out of 5 questions(`['yes', 'yes', 'no', 'yes', 'no']`), `qa_score` would be 0.6.
And if the summary length is 400 and original text length was 1000, `relative_length` of summary would be 0.4.
Consider two summaries `S1` and `S2` for a text `T`, if `len(S1)=450` and `len(S2)=600` and both have the same `qa_score`, we would prefer the shorter summary, so I used `conciseness_score=1-(len(S)/len(T))` , shorter the summary, larger the `conciseness_score`.

Finally, I take the average of the two scores to return the `summary_score`.

**Usage in RAGAS**(as per current code in the PR, subject to change)
```python
import datasets
import pandas as pd
from ragas.metrics import summary_score
from ragas import evaluate
from langchain.chat_models import ChatOpenAI
from ragas.llms.base import LangchainLLMWrapper

data = pd.DataFrame(columns=['text', 'summary'])
data['text'] = <TEXTS>
data['summary'] = <SUMMARIES-FOR-ABOVE-TEXTS>

llm = ChatOpenAI(
    model=<YOUR-MODEL>,
    openai_api_key=<OPENAI_API_KEY>,
)

summary_eval_response = evaluate(
    datasets.Dataset.from_pandas(data), 
    metrics=[summary_score], llm=LangchainLLMWrapper(llm))
```

**Discussions and Further thoughts**
- As far as I have checked, no other summarization metric codebase takes relative length of the summary into account, so this could be useful to promote shorter summaries.
- I am reading some papers and have some idea in mind to add more logic to this metric, considering diversity(based on original text) in summary, usefulness of summary etc. Will add to this PR or create new ones.

@shahules786 @jjmachan Looking forward for review and discussions.

### DIFF
--- a/src/ragas/metrics/_summarization_metric.py
+++ b/src/ragas/metrics/_summarization_metric.py
@@ -1,0 +1,228 @@
+from __future__ import annotations
+
+import logging
+import typing as t
+from dataclasses import dataclass, field
+from typing import Dict
+
+import json
+import numpy as np
+from langchain.pydantic_v1 import BaseModel
+
+from ragas.llms.output_parser import RagasoutputParser, get_json_format_instructions
+from ragas.llms.prompt import Prompt, PromptValue
+from ragas.metrics.base import EvaluationMode, MetricWithLLM
+
+if t.TYPE_CHECKING:
+    from langchain.callbacks.base import Callbacks
+
+logger = logging.getLogger(__name__)
+
+class GenerateQuestionsResponse(BaseModel):
+    questions: t.List[str]
+
+class GenerateAnswersResponse(BaseModel):
+    answers: t.List[str]
+
+
+_output_instructions_question_generation = get_json_format_instructions(
+    pydantic_object=GenerateQuestionsResponse
+)
+_output_instructions_answer_generation = get_json_format_instructions(
+    pydantic_object=GenerateAnswersResponse
+)
+
+_output_parser_question_generation = RagasoutputParser(pydantic_object=GenerateQuestionsResponse)
+_output_parser_answer_generation = RagasoutputParser(pydantic_object=GenerateAnswersResponse)
+
+GENERATE_QUESTION_INSTRUCTION = """\
+Based on the given text, generate "n" closed-ended questions that can be answered with either a 'yes' or 'no'. The questions generated should ALWAYS result in a 'yes' based on the given text.    
+** IMPORTANT
+Only return a JSON with a 'questions' key, which is a list of strings. The questions have to be STRICTLY closed ended. The given text should be able to answer 'yes' for each question.
+**
+"""
+TEXT_GENERATE_QUESTIONS = Prompt(
+    name="text_generate_questions",
+    instruction=GENERATE_QUESTION_INSTRUCTION,
+    output_format_instruction=_output_instructions_question_generation,
+    input_keys=["text", "n"],
+    output_key="questions",
+    output_type="json",
+    examples=[
+        {
+            "text": """JPMorgan Chase & Co. is an American multinational finance company headquartered in New York City and incorporated in Delaware. It is the largest bank in the United States and the world's largest bank by market capitalization as of 2023. As the largest of Big Four banks, the firm is considered systemically important by the Financial Stability Board. Its size and scale have often led to enhanced regulatory oversight as well as the maintenance of an internal "Fortress Balance Sheet". The firm is headquartered at 383 Madison Avenue in Midtown Manhattan and is set to move into the under-construction JPMorgan Chase Building at 270 Park Avenue in 2025.
+                    The firm's early history can be traced to 1799, with the founding of what became the Chase Manhattan Company. In 1871, J.P. Morgan & Co. was founded by J. P. Morgan who launched the House of Morgan on 23 Wall Street as a national purveyor of commercial, investment, and private banking services. The present company was formed after the two predecessor firms merged in 2000, creating a diversified holding entity. It is a major provider of investment banking services, through corporate advisory, mergers and acquisitions, sales and trading, and public offerings. Their private banking franchise and asset management division are among the world's largest in terms of total assets. Its retail banking and credit card offerings are provided via the Chase brand in the U.S. and United Kingdom.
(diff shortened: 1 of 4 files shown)


shahules786:
Hey @sky-2002 Thanks for the PR and great work.  Loved the way you added both factual & text length penalizing dimensions. 

A couple of reviews from my side: 

1. While framing questions, I think it is important to make sure that questions are framed from parts of the text that are non-trivial. For example, questions should be framed from named entities or keyphrases could be a solution to this
2. Regarding The output instead of yes/no, I would recommend keeping it as 1/0 to make it more parsable and language agnostic so that even if the prompt is converted to other languages using r[agas prompt adaptation](https://docs.ragas.io/en/stable/concepts/prompt_adaptation.html) it would remain working.

shahules786 (changes requested):
Hey @sky-2002 Thanks for the PR and great work.  Loved the way you added both factual & text length penalizing dimensions. 

A couple of reviews from my side: 

1. While framing questions, I think it is important to make sure that questions are framed from parts of the text that are non-trivial. For example, questions should be framed from named entities or keyphrases could be a solution to this
2. Regarding The output instead of yes/no, I would recommend keeping it as 1/0 to make it more parsable and language agnostic so that even if the prompt is converted to other languages using r[agas prompt adaptation](https://docs.ragas.io/en/stable/concepts/prompt_adaptation.html) it would remain working.

sky-2002:
Hey @shahules786 @jjmachan 
- Made changes to frame questions on important keywords and events
- I have also added a new dimension of **topic distribution similarity**(let me know if this sounds good) to summary evaluation, this would:
    - Promote diverse summaries
    - Include contributions from all topics in the text in proportion of their length share in whole text.
 
Would create a new detailed PR for docs.

src/ragas/metrics/_summarization_metric.py
```diff
@@ -0,0 +1,412 @@
...
+_output_parser_topics_extraction = RagasoutputParser(pydantic_object=ExtractTopicsResponse)
+_output_parser_link_summary_topics = RagasoutputParser(pydantic_object=LinkSummaryTopicsResponse)
+
+EXTRACT_TOPICS_INSTRUCTION = """\
+Based on the given text, extract some broad opics, events, concepts or other important broader areas. Then for each of those extracted topic, associate chunks of text that are related to it and make topic-chunk pairs.
+
+** Important
+Return in JSON format with key as 'topics' and a dictionary of topic-chunk pairs as the value. Only return the JSON.
+"""
+TEXT_EXTRACTION_TOPICS = Prompt(
```
shahules786, line 58: * The idea of topic distribution is interesting but I have some concerns about it 
    * The assumption here is that each topic in the text has to be equally important, which would not be true for many texts
    * The workflow for calculating it is very expensive in terms of tokens. 
    * Paradigms for metrics should be very simple to understand, this concept seems little jargon for users coming from non-data science backgrounds.

src/ragas/metrics/_summarization_metric.py
```diff
@@ -0,0 +1,412 @@
...
+                "Chemosynthetic bacteria and their role": "Only chemosynthetic bacteria, which use chemical energy from inorganic compounds, would survive.",
+                "Energy production from photosynthesis and fossil fuels": "Photosynthesis from millions of years ago created fossil fuels, essential for modern society.",
+                "Formation of fossil fuels and their significance": "Photosynthesis from millions of years ago created fossil fuels, essential for modern society.",
+                "Environmental implications of rapid carbon dioxide rise": "Their rapid consumption is increasing carbon dioxide levels at an unprecedented rate, significantly impacting Earth's climate."
+            }
+        }
+    ]
+)
+
+GENERATE_QUESTION_INSTRUCTION = """\
```
shahules786, line 159: The entity/keyword extraction should be separate tasks and not part of the same prompt that creates the question. I would create two prompts
1) To extract entities and or keyphrases from given text
2) Generate questions from text given text and keyphrases and or entities

Although mering this to one improves efficiency, it lacks robustness.

shahules786:
Hey @sky-2002 thanks for the updates. I have shared my reviews
An ideal workflow for me would be

Entity/text extraction -> Question generation -> Verification 
+
Length penalty

Let me know if something is not clear.

shahules786 (changes requested):

src/ragas/metrics/_summarization_metric.py
sky-2002: Agreed. Removed it for now. Would come up with a lighter version sometime later.

src/ragas/metrics/_summarization_metric.py
sky-2002: @shahules786 Done, made the changes. 
Though I have a concern that how would we regulate the number of questions generated for a text?

shahules786 (approved):

aakashthatte:
@shahules786 Looks good to me

sky-2002:
@shahules786 Looks good to me

src/ragas/validation.py
```diff
@@ -67,6 +67,7 @@ def validate_column_dtypes(ds: Dataset):
     EvaluationMode.ga: ["ground_truth", "answer"],
     EvaluationMode.qga: ["question", "ground_truth", "answer"],
     EvaluationMode.qcg: ["question", "contexts", "ground_truth"],
+    EvaluationMode.ts: ["text", "summary"],
```
jjmachan, line 70: I have one question here. would it make sense to make it ["contexts"] and summary?

the reasoning was that 'text' and 'contexts' object seems similar and I was wondering if we could join them together

what do you think @sky-2002 @shahules786 ?

shahules786: Yes.

sky-2002: @jjmachan @shahules786 Done

jjmachan:
awesome, let merge this in @sky-2002 🎉

PR #994, opened 2024-05-28
Handle embeddings for empty strings in AnswerSimilarity class

baptiste-pasquier:
Fixes: #996 

Some embedding models do not work with empty strings (for example Gemini) and return an error.
The error does not appear with OpenAI which returns an embeddings even for an empty string.

Proposed resolution: replace empty strings with `" "`


jjmachan:
merging this in - sorry it took a while do checkout this form https://docs.google.com/forms/d/e/1FAIpQLSdM9FrrZrnpByG4XxuTbcAB-zn-Z7i_a7CsMkgBVOWQjRJckg/viewform - our way of saying thank you 🙂

PR #887, opened 2024-04-21
fix: temperature parameter in generate_text not ignored.

HerrIvan:
Addresses #886. 

Temperature parameter is not anymore overwritten by calling to `self.get_temperature(n=n)`. Now it will only call that method if no parameter was given.

Issue #886, opened 2024-04-21
[R-231] `generate_text` in `LangchainLLMWrapper` ignores the value of the temperature parameter


HerrIvan:
[x] I have checked the [documentation](https://docs.ragas.io/) and related resources and couldn't resolve my bug.

**Describe the bug**
The method `generate_text` in `LangchainLLMWrapper` doesn't take into account the temperature parameter.

Ragas version: '0.1.dev368+g8a5ff9f'
Python version: indifferent

**Code to Reproduce**
```python
    async def agenerate_text(
        self,
        prompt: PromptValue,
        n: int = 1,
        temperature: float = 1e-8,
        stop: t.Optional[t.List[str]] = None,
        callbacks: Callbacks = None,
    ) -> LLMResult:
        temperature = self.get_temperature(n=n)  # <- this function sets a temperature value regardless of the one passed as a parameter
```

**Error trace**
This "bug" won't throw any exception but the code won't work as expected.

**Expected behavior**
Take the given temperature value into account if explicitly provided. Now, since the temperature has a default value already in the signature, it is not straightforward to know whether the param has been defined, or it implicitly has taken the default value. Therefore the signature should be slightly modified (at the "typing" level).

**Additional context**
Since `agenerate_text` is a concrete implementation of an abstract method in `BaseRagasLLM`, probably the abstract class signature would need to be updated too. I don't see in the repo any place where removing the default value in the signature would cause any problem.

I will open myself a PR with a solution proposal.

<sub>[R-231](https://linear.app/exploding-gradients/issue/R-231/generate-text-in-langchainllmwrapper-ignores-the-value-of-the)</sub>

jjmachan:
closed with PR, thanks again @HerrIvan 🙂 ❤️

PR #998, opened 2024-05-30
consider langchain object temperature attr in generation

fealbuqu:
Addresses https://github.com/explodinggradients/ragas/issues/886.

Taking into account solution made by this [PR](https://github.com/explodinggradients/ragas/pull/887) but also checks if temperature was set under langchain object.

This is critical because on TGI endpoints no generation is returned with such low value as `1e-8`. And often langchain classes has the temperature attribute set during initialization.


jjmachan (changes requested):
there is a lot of different things in this PR so I'm merging in the original PR. One issue with taking the default from the object is that the default for that is very high and we want to over ride that. 

Ideally we need to come up with some other way one can interface with temperature I think.

PR #1025, opened 2024-06-12
adding `get_available_metrics` method

immortal3:
Based on databaset column, it tries to return all available metrics which are suitable for evalution

### DIFF
--- a/src/ragas/metrics/__init__.py
+++ b/src/ragas/metrics/__init__.py
@@ -1,3 +1,5 @@
+from datasets import Dataset
+
 from ragas.metrics._answer_correctness import AnswerCorrectness, answer_correctness
 from ragas.metrics._answer_relevance import AnswerRelevancy, answer_relevancy
 from ragas.metrics._answer_similarity import AnswerSimilarity, answer_similarity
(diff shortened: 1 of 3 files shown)


src/ragas/metrics/__init__.py
```diff
@@ -15,6 +17,37 @@
...
+    context_precision,
+    context_utilization,
+    context_recall,
+    context_relevancy,
+    answer_relevancy,
+    context_entity_recall,
+]
+
+
+def get_available_metrics(ds: Dataset) -> list[Metric]:
```
jjmachan, line 36: could you move this to a new file called `metrics/utils` and import it here?
that will be much cleaner

jjmachan (changes requested):
thanks again for the PR, just one tiny change and we should be good to go

src/ragas/metrics/__init__.py
immortal3: yeah, my plan was to do in utils.py only, but it was having circular dependency. Raised new changes which brings this function to `metrics/utils`

jjmachan (approved):
thanks a lot!

PR #1033, opened 2024-06-19
fix(typo): missing 'to' in prompt instruction

desaxce:
Looking into the underlying prompt to compute context relevancy, the missing "to" annoyed me.


desaxce:
Just saw that `context_relevancy` got deprecated and I should switch to `context_precision`. Closing.

PR #1050, opened 2024-06-27
Feature/updating prompt

louky123:
Editing prompt.py so that adaptation of testset_generator to dutch is possible;
Added a output_parser for the question_answer_prompt, so that there are less nans with testset_generation

### DIFF
--- a/src/ragas/llms/prompt.py
+++ b/src/ragas/llms/prompt.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import json
 import logging
 import os
+import ast
 import typing as t
 
 from langchain_core.messages import BaseMessage, HumanMessage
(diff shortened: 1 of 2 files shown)


shahules786 (approved):

shahules786:
The PR seems good to me. But could you also pin the issue that caused issue without this modification? 
@louky123

louky123:
> The PR seems good to me. But could you also pin the issue that caused issue without this modification? @louky123

@shahules786  There were two seperate things, maybe I should have split them up in two seperate PRs.
1. [#774 ](https://github.com/explodinggradients/ragas/issues/774#issue-2192005926). When changing the Testsetgenerator language to Dutch (and possibly some other languages) there is an error because  the adapt function also tries to translate the dict : {'relevant_contexts': [1, 2]} (in prompt.py). It fails to return the expected format because the LLM adds some text with the translation and doesnt return a dict. I saw others worked around it by adding more prompts for specific languages. 
2. I use AzureOpenAI and Azure OpenAI embeddings instead of OpenAI for the Testsetgenerator, and we get a lot of nans in the generated testset regardless of the language used. When debugging I found out it is because the LLM does not generate a ground-truth in the prescribed format of a dict with the keys 'answer' and 'verdict'. I added the Ragas Output Parser for the question_answer_prompt with the idea that the output structure will be more strict.

shahules786:
Thanks buddy @louky123

PR #1051, opened 2024-06-27
fix: langchain EvaluatorChain OpenAI validation error

danielfornarini:
# This mr solve the following problem:
- When you are using a custom llm or embeddings with langchain EvaluatorChain and you have not set OpenAI's ENVs, a validation error is raised for missing ENVs


jjmachan (approved):
Thank you for the fix!

PR #1083, opened 2024-07-10
[Fix bug] utils.py

jrredondo:
Error when retrieving values from dictionary


jjmachan:
@jrredondo Thank you for catching that 🙂 ❤️ 
and welcome to the contributor list!

PR #1111, opened 2024-07-18
chore: remove context_relevancy from library

jjmachan:


shahules786 (approved):

xiaokening:
why do you remove context_relevancy from library? can you give me a explanation? thanks!

zhiwei2017:
Hey,
just want to comment on the best practises of semantic versioning, usually people remove functions in at least minor version, not patch version, even though there were deprecation warning for the removed functions. In patch versions, only the bugs should be fixed. Please check the documentation of [semantic versioning](https://semver.org/#spec-item-6)

jjmachan:
@zhiwei2017 that is a good point, thanks for bringing it up. Will keep this in mind moving forward! 🙂 

sorry for the inconvenience

mongodben:
since this has been removed, it'd be great if you could have some other reference-free evaluation metric that looks at how relevant retrieved context is to answer an input. 

Could you also expand on why context relevancy "turned out to be not a useful metric" meriting its removal?

jjmachan:
@shahules786 could you help @mongodben out?

shahules786:
Hi @mongodben thanks for the suggestions.
I just added #1174 to the documentation. We had it for a while but it wasn't displayed in doc making it almost invisible to users. 
Try it out and let us know :)

Issue #1016, opened 2024-06-05
[R-270] Synthetic test data generation - v3


jjmachan:
Highly programmable knowledge graph-based synthetic data generation grounded on documents.

PR raised #1024

<sub>From [SyncLinear.com](https://synclinear.com) | [R-270](https://linear.app/exploding-gradients/issue/R-270/synthetic-test-data-generation)</sub>

PR #1138, opened 2024-07-31
Empty generation, change None checking to is_nan call over list elements

### LINKED ISSUE #1137: Empty generation does not raise a `ragas.exceptions.ExceptionInRunner`

**Describe the bug**
Hi! I'm currently working with `ragas` to test different RAG architectures, so I'm using `Ollama`, `HuggingFace` and `LangChain` framework on top of `ragas` and I'm facing an issue when I'm trying to implement an unit test around the synthetic generation : the generated `ragas.testset.generator.TestsetGenerator` object has empty rows.

I think it comes from a specific parametrization of each frameworks but after having looked over the repository, I think you've tried to avoid such a situation but your checking does not raise any error.

Ragas version: 0.1.11
Python version: 3.10.12

**Code to Reproduce**
The `example.pdf` file used here can be found at: https://css4.pub/2015/usenix/example.pdf

```python
from langchain_community.chat_models import ChatOllama
from ragas.testset import TestsetGenerator
from rag_sandbox.embedding.huggingface import HuggingFaceEmbeddings
from langchain_community.document_loaders.pdf import UnstructuredPDFLoader

if __name__ == "__main__":
    llm = ChatOllama(base_url="http://localhost:11434", model="qwen2:7b")

    generator = TestsetGenerator.from_langchain(
        generator_llm=llm,
        critic_llm=llm,
        embeddings=HuggingFaceEmbeddings()
    )

    documents = UnstructuredPDFLoader(
        "./tests/data/pdf/example.pdf",
        mode="elements",
        strategy="hi_res",
        infer_table_structure=True,
        hi_res_model_name="yolox",
        extract_images_in_pdf=True,
        extract_image_block_output_dir="./results/data_extraction/images"
        ).load()

    dataset = generator.generate_with_langchain_docs(
        documents=documents,
        test_size=1
    )
```

**Error trace**
No error, but that is the problem.

**Expected behavior**
According to your code, one can expect to get a `ragas.exceptions.ExceptionInRunner` in such a situation.

**Additional context**
I'll offer a PR to fix this issue, but I don't know if it won't conflict with another part of the code.

Gwenn-LR:
# Description
Should fix #1137.

# Fonctionnality
An empty generation from `TestsetGenerator.generate` should raise a `ragas.exceptions.ExceptionInRunner`.

# Issue
Currently the behaviour is expected as written in the code but it is not actived by the comparison `if not test_data_rows`.

# Solution proposed
Check if all generated rows are empty and raise if it is the case (another approach could be to raise an error only if at least one element is nan).

### DIFF
--- a/src/ragas/testset/generator.py
+++ b/src/ragas/testset/generator.py
@@ -301,7 +301,7 @@ class TestsetGenerator:
 
         try:
             test_data_rows = exec.results()
-            if not test_data_rows:
+            if all(is_nan(row) for row in test_data_rows):
                 raise ExceptionInRunner()
 
         except ValueError as e:
--- a/src/ragas/testset/generator.py
+++ b/src/ragas/testset/generator.py
@@ -301,7 +301,7 @@ class TestsetGenerator:
 
         try:
             test_data_rows = exec.results()
-            if all(is_nan(row) for row in test_data_rows):
+            if all([is_nan(row) for row in test_data_rows]):
                 raise ExceptionInRunner()
 
         except ValueError as e:


src/ragas/testset/generator.py
```diff
@@ -301,7 +301,7 @@ def generate(
 
         try:
             test_data_rows = exec.results()
-            if not test_data_rows:
+            if all([is_nan(row) for row in test_data_rows]):
                 raise ExceptionInRunner()
```
jjmachan, line 305: actually I don't know if this change is accurate. Originally if exec.result is `None` it signified an error in executor. Since this is not the case now that could be one change here

Gwenn-LR: I agree for your statement, it would raise an error if `exec.result` is None, but currently it returns a list of results:
https://github.com/explodinggradients/ragas/blob/ffebf4f064ba995cbf9652a56183ece7800170d8/src/ragas/executor.py#L109
Python does not handle those values the same way, you can highlight it with:
```python
>>> print("OK" if not None else "Not Ok")
OK
>>> print("OK" if not [None] else "Not Ok")
Not Ok
```

Furthermore, it is not a None as the list element since it is the result of a `ragas.testset.evolutions.Evolution` which returns a `np.nan`:
```python
>>> print("OK" if not np.nan else "Not Ok")
Not Ok
>>> print("OK" if not [np.nan] else "Not Ok")
Not Ok
```

Gwenn-LR: In any case, thanks again for your amazing work that helps me and a lot of the research community! I'm just trying to make my contribution to this wonderful project so I don't fix it only on my side but also on everybody side!

shahules786:
Hey @Gwenn-LR on a related note, since you're into synthetic data generation we are preparing to release a[ v3 version](https://github.com/explodinggradients/ragas/issues/1016) of this (10x improvement in quality and 10x reduction in token usage). I would love to talk and make sure that your requirements are also included in the next release. If you're available https://cal.com/shahul-ragas/30min

Gwenn-LR:
@shahules786, I've booked a meeting with you, I'm eager to talk with you!

src/ragas/testset/generator.py
jjmachan: hey @Gwenn-LR thanks for the kind words 🙂

jjmachan:
@Gwenn-LR I'm closing this in favour of removing the `ExceptionInRunner` altogether. with #1093 there was no need for that

Issue #1140, opened 2024-07-31
Random seed fixed.


Gwenn-LR:
- [x] I have checked the [documentation](https://docs.ragas.io/) and related resources and couldn't resolve my bug.

**Describe the bug**
While looking for the answer to the issue #1138, I've noticed my generation issue was coming from the generation of an empty row. Across many tests, I've always got this issue and when I've tried to generate a second row, I've noted it was always the same generated row. Looking over our repository, I've finally found that you were fixing the `numpy.random`'s seed, 

Ragas version: 0.1.11
Python version: 3.10.12

**Code to Reproduce**
Here is a script that takes your implementation and illustrates the random choice behaviour when we relaunch an experiment:

```python
import numpy as np

if __name__ == "__main__":
    test = [0, 1, 2, 3]

    for exp_num in range(3):
        print(f"Experiment n°{exp_num}")
        rng = np.random.default_rng(seed=42)
        for choice_num in range(10):
            print(f"\tChoice n°{choice_num}: {rng.choice(test)}")
```

**Error trace**
| Experiment | Choices |
| :-: | :-: |
| 1 |0, 3, 2, 1, 1, 3, 0, 2, 0, 0 |
| 2 |0, 3, 2, 1, 1, 3 ,0 ,2 ,0, 0 |
| 3 |0, 3, 2, 1, 1, 3 ,0 ,2 ,0, 0 |

**Expected behavior**
If a user does not give any seed, the generation should be totally random.

**Additional context**
I'll submit a PR integrating this seed and `numpy.random.Generator` along with the `RunConfig` and use it accordingly wherever this `rng` was used.

PR #1142, opened 2024-07-31
Random seed fixed.

Gwenn-LR:
# Description
Should fix https://github.com/explodinggradients/ragas/issues/1140

# Fonctionnality
Allow a user to fix the seed him/herself and use a random seed otherwise.

# Solution proposed
Define the seed as an initial attribute and the rng as a post_initialization attribute that a user can override and adapt its use in ragas.testset.docstore.InMemoryDocumentStore and ragas.testset.evolutions.Evolution

### DIFF
--- a/src/ragas/run_config.py
+++ b/src/ragas/run_config.py
@@ -2,6 +2,7 @@ import logging
 import typing as t
 from dataclasses import dataclass
 
+import numpy as np
 from tenacity import (
     AsyncRetrying,
     Retrying,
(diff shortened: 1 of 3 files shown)


jjmachan:
@Gwenn-LR thanks for the PR, will let @shahules786 take a call on this 🙂

shahules786:
Hey, thanks for the PR @Gwenn-LR . LGTM

shahules786 (approved):

Gwenn-LR:
I've just noted that those changes don't keep your setup, it would be better to have `seed: t.Optional[int] = 42` so you would have the same behaviour as before !

Gwenn-LR:
There is also an issue with the definition of the __post_init__ method where the double underscores at the end are missing... Sorry for the flaky PR, I'll do better next time!

jjmachan:
@Gwenn-LR, yeah no worries about that but could you make the PR to fix it? I see that the CI was broken too, fixing that now

Gwenn-LR:
Ahah I've just done it in another PR (#1152) exactly when you've asked me to do so ^^

PR #1147, opened 2024-07-31
Update on Json Format Prompt

emreds:
Hi! 

I was calculating the summarization score and faithfullness metric with the `azure openai gpt35-turbo-1106 ` but sometimes it was not giving proper JSON output and it was breaking the code for me.

The following change in the JSON formatting prompt solved the problem for me.


shahules786 (commented):
Hi @emreds thanks for the PR. It's understandable that this prompt fixed the issue for you but it might not be generalizable beyond your situation (ie, different model, different dataset). @jjmachan is also working on a better prompt format to fix the issue of output parsing.

emreds:
@shahules786 makes sense, mine was more like a quick fix. Let me know if there is something I can do.

shahules786:
@emreds thanks for asking. I'll close this PR for now - we will be working on a DSPy-like optimization to make this kind of prompt optimization automatic in future releases.

PR #1152, opened 2024-08-01
Fix errors from previous PR(#1142)

Gwenn-LR:
# Description
My previous PR (#1142) was flaky so I propose this new one to avoid any future issues.

# Functionnality
2 issues appeared in the previous PR:
- The previous default behaviour (with rng=np.random.default_rng(seed=42)) was not conserved.
- The `rng` attribute could not be set since the 2 ending underscores of the `__post_init__` method of the `RunConfig` dataclass were missing

# Solution proposed
A small PR to fix the 2 previous missing functionnality

### DIFF
--- a/src/ragas/metrics/_answer_correctness.py
+++ b/src/ragas/metrics/_answer_correctness.py
@@ -281,6 +281,8 @@ class AnswerCorrectness(MetricWithLLM, MetricWithEmbeddings):
             language, self.llm, cache_dir
         )
 
+        self.sentence_segmenter = get_segmenter(language=language, clean=False)
+
     def save(self, cache_dir: t.Optional[str] = None) -> None:
         self.correctness_prompt.save(cache_dir)
 
--- a/src/ragas/metrics/_faithfulness.py
+++ b/src/ragas/metrics/_faithfulness.py
@@ -306,8 +306,11 @@ class Faithfulness(MetricWithLLM):
             language, self.llm, cache_dir
         )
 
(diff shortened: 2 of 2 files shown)


Gwenn-LR:
Wait wait, I've mispelled `__post_init__` again....

Gwenn-LR:
Yeah, that's (finally) good, sry for the extra work I don't think you need this when there are so many people working with your repo.

src/ragas/run_config.py
```diff
@@ -31,9 +31,9 @@ class RunConfig:
         t.Tuple[t.Type[BaseException], ...],
     ] = (Exception,)
     log_tenacity: bool = False
-    seed: t.Optional[int] = None
+    seed: t.Optional[int] = 42
```
jjmachan, line 34: no need for t.Optional here, just int is best

jjmachan (commented):
woun't be merging this in since I made some changes in another PR. but on thing you can work on is adding some tests for this section. 

something simple would work. 

I made the changes you suggested in #1129

jjmachan:
@Gwenn-LR let me know if you want to take it up, or else we'll close this. Thanks a lot for making the PR 🙂, would still love to merge it in if you want to add some small test or something or else feel free to close it.

Gwenn-LR:
@jjmachan I see that you've added a test in #1158, I don't think I've anything extra to add. I'll let the PR opened until this weekend to let me think of it, otherwise I'll close it on Sunday if it's okay with you.

jjmachan:
@Gwenn-LR sure thing 🙂

PR #1157, opened 2024-08-02
Add save function for statement_prompt

miyaketomoya:
Fixed problem with statement_prompt not being saved in cache when applying language adapter to faithfulness


jjmachan:
@miyaketomoya thank you so much for the fix and welcome to the contributors list 🙂 ❤️

PR #1158, opened 2024-08-02
fix: run_config post_init issue

jjmachan:


jjmachan:
@Gwenn-LR something we missed 😅

PR #1161, opened 2024-08-02
Allow Metric.score to work within an existing asyncio loop

abramowskia:
I got errors when running metrics within an existing asyncio loop, the `asyncio.run` part makes the code fail in those cases. Submitting a bugfix.

### DIFF
--- a/src/ragas/metrics/base.py
+++ b/src/ragas/metrics/base.py
@@ -96,7 +96,11 @@ class Metric(ABC):
         callbacks = callbacks or []
         rm, group_cm = new_group(self.name, inputs=row, callbacks=callbacks)
         try:
-            score = asyncio.run(self._ascore(row=row, callbacks=group_cm))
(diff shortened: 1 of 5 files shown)


src/ragas/metrics/base.py
```diff
@@ -96,7 +96,8 @@ def score(self: t.Self, row: t.Dict, callbacks: Callbacks = None) -> float:
         callbacks = callbacks or []
         rm, group_cm = new_group(self.name, inputs=row, callbacks=callbacks)
         try:
-            score = asyncio.run(self._ascore(row=row, callbacks=group_cm))
+            loop = asyncio.get_event_loop()
+            score = loop.run_until_complete(self._ascore(row=row, callbacks=group_cm))
```
jjmachan, line 100: this will still fail in jupyter notebook envs right?

jjmachan: sorry my bad!

PR #1164, opened 2024-08-04
Bugfix Summarization Score

emreds:
Hi! 

I was experimenting with Summarization Score and found an issue that when there are no questions generated by the LLM, it throws division by zero error due to 0 generated answers. 

## Notable Changes
- Added an assertion to keyphrases, questions and answers generation functions.
- Changed the llm response variable name `answer` to `response` to prevent possible confusion with score related `answer`.

Please let me know if there is something I can add. 

Cheers!


jjmachan (commented):
@emreds thanks a lot for pushing this PR and helping us Improve Ragas. I was however not sure if an assertion here is the best flow.

it will break the user code right if the assertions fail? Ideally the behaviour we want is to use `logger.error()` to print out an explainable error and then return a Nan value (or throw and exception since executor should catch it). 

what do you think?

emreds:
@jjmachan sure, makes more sense. I will update it.

src/ragas/metrics/_summarization.py
```diff
@@ -232,7 +235,9 @@ async def _get_questions(
             result_text, p_value, self.llm, self.max_retries
         )
         
-        assert response and response.questions, "No questions generated, unable to calculate the score."
+        if not response or not response.questions:
+            logging.error("No questions generated, unable to calculate the score.")
+            return math.nan
```
jjmachan, line 240:
```suggestion
            return np.nan
```

jjmachan (changes requested):
math.nan to np.nan. 

I've changed one, do change the others too?

just to keep things consistent throughout the codebase 🙂

src/ragas/metrics/_summarization.py
emreds: Sure, I have changed the all and made a new commit.

jjmachan (approved):
sweet! thanks a lot @emreds ❤️

jjmachan:
@emreds I made some small changes for fixing the type error but now it still returns an empty list. This is so that the types make sense. but I'm not sure if it logically correct

PR #1166, opened 2024-08-05
change default model to gpt4o-mini

shahules786:
Reasons
1. gpt4o-mini is 2x cheaper than the current default model of gpt3.5-turbo
2. Extended context length of 128k tokens
3. All the public benchmarks indicate better performance on all tasks by gpt4o-mini compared to gpt3.5-turbo. [Source](https://openai.com/index/gpt-4o-mini-advancing-cost-efficient-intelligence/)


jjmachan (changes requested):
change testset generator also

jjmachan (approved):

PR #1169, opened 2024-08-05
Add tests to `ragas.run_config.RunConfig` `rng` attribute.

Gwenn-LR:
# Description
Previous exchanges about the `numpy.random.Generator` (#1140, #1142 and #1152) led to the need to implement some tests. Previous ones had been setup but a more precise definition of the context box might strengthen the confidence in the overall generation.

# Fonctionnality
Check the ability to generate a specific `numpy.random.Generator`, its fixed behaviour between generations and its consistency accross module reloads but also the actual random generation when the `seed` is set to `None`.

# Solution proposed
Pytest with 2 parametrized test: 1 with the `seed` fixed, the other one with `seed` set to `None`. Each test will check whether the generation is fixed, reproductible and consistent according to the seed.

### DIFF
--- a/tests/unit/test_run_config.py
+++ b/tests/unit/test_run_config.py
@@ -1,12 +1,17 @@
-from typing import TypeAlias, Callable
 import sys, importlib
-
+from packaging.version import parse as parse_version
+from platform import python_version
 import pytest
 from numpy.random import default_rng, Generator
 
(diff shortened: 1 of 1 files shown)


jjmachan (approved):
wow thank you so much for taking the time to add the tests ❤️ 🙌🏽  @Gwenn-LR 🙂

PR #1174, opened 2024-08-06
Add context utilization

shahules786:
reference free context precision

### DIFF
--- a/docs/concepts/metrics/context_utilization.md
+++ b/docs/concepts/metrics/context_utilization.md
@@ -44,7 +44,7 @@ score.to_pandas()
 
 ## Calculation 
 
-Let's examine how context precision was calculated using the low context precision example:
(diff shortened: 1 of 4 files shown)


docs/concepts/metrics/context_utilization.md
```diff
@@ -0,0 +1,65 @@
...
+score.to_pandas()
+```
+
+## Calculation 
+
+Let's examine how context precision was calculated using the low context precision example:
+
+**Step 1**: For each chunk in retrieved context, check if it is relevant or not relevant to arrive at the answer for the given question.
+
+**Step 2**: Calculate precision@k for each chunk in the context.
```
jjmachan, line 51: should be context_utilization here?

jjmachan (approved):

PR #1178, opened 2024-08-08
Helicone Integration

chitalian:
Tested and working integration with Helicone

image


jjmachan:
@chitalian thanks alot for putting this integration together - would love to share it in the socials with you guys on monday. what do you think?

Issue #890, opened 2024-04-22
[R-229] Automatic language adaptation bugs cleanup


rere950303:
- [X] I have checked the [documentation](https://docs.ragas.io/) and related resources and couldn't resolve my bug.

**Describe the bug**
When i try automatic language adaptation for a faithfulness metric, i get `AssertionError: Adapted output key set(output.keys())={'statements'} does not match original output key: output_keys[i]=[]` error

Ragas version: 0.1.7
Python version: 3.8.11

**Code to Reproduce**

```python
    openai_llm = ChatOpenAI(model_name="gpt-4-turbo-2024-04-09")
    openai_embedding = OpenAIEmbeddings(model="text-embedding-3-large")

    faithfulness = Faithfulness()
    answer_relevancy = AnswerRelevancy()
    context_precision = ContextPrecision()
    context_utilization = ContextUtilization()
    context_relevancy = ContextRelevancy()
    context_recall = ContextRecall()
    answer_similarity = AnswerSimilarity()
    answer_correctness = AnswerCorrectness()
    harmfulness = AspectCritique(
        name="harmfulness",
        definition="Does the submission causes or has the potential to cause harm to individuals, groups, or society at large?",  # noqa: E501
        strictness=3
    )
    maliciousness = AspectCritique(name="maliciousness", definition="Is the submission intends to harm, deceive, or exploit users?", strictness=3)
    coherence = AspectCritique(
        name="coherence",
        definition="Does the submission presents ideas, information, or arguments in a logical and organized manner?",  # noqa: E501,
        strictness=3
    )
    correctness = AspectCritique(name="correctness", definition="Is the submission factually accurate and free from errors?", strictness=3)
    conciseness = AspectCritique(
        name="conciseness",
        definition="Does the submission conveys information or ideas clearly and efficiently, without unnecessary or redundant details",  # noqa: E501,
        strictness=3
    )

    adapt(
        metrics=[
            faithfulness, answer_relevancy, context_precision, context_utilization, context_relevancy,
            context_recall, answer_similarity, answer_correctness, harmfulness, maliciousness,
            coherence, correctness, conciseness
        ],
        language=japanese,
        llm=openai_llm,
        cache_dir="./prompt_adaptation_cache"
    )
```

**Error trace**

```
[...]

PR #1181, opened 2024-08-09
solved issue that adaptation crashes when havig multiple nested promp…

ssoima:
I have faced the issue that when I adapt some prompts, some of the key-value pairs are objects itself and then do not get stored properly. The adaptation does not fail, however, when loading those prompt, they break since they are not in proper json format. This fix solves this particular issue I am using json.dumps on those object to store json-conform strings.


jjmachan (approved):
thank you so much for this fix! hopefully it will address things in #890 . I will make sure there are some more tests here in another PR

thank again ❤️ 🙂 🙌🏽

PR #1187, opened 2024-08-10
Fixed metric adaptation issue with the prompt factory

ssoima:
Issue:
Runtime metric language adaptation (e.g. adapting to italian and then to german in the same run) does not work. 

The reason is that the @dataclass factories in the metrics use singleton objects (which only get initialized when the module is loaded) => when the prompt is instantiated a second time, the same object gets returned. Check out this loom for a debug preview: https://www.loom.com/share/2a500a4d83064489936886674e8641f9?sid=682f87fc-4052-4a33-89e4-5e4d1a6663dd 

Solution: 
Changed the prompts to factory methods that returns a new prompt at every call

### DIFF
--- a/src/ragas/metrics/_answer_correctness.py
+++ b/src/ragas/metrics/_answer_correctness.py
@@ -158,9 +158,9 @@ class AnswerCorrectness(MetricWithLLM, MetricWithEmbeddings):
 
     name: str = "answer_correctness"  # type: ignore[reportIncompatibleMethodOverride]
     evaluation_mode: EvaluationMode = EvaluationMode.qga  # type: ignore[reportIncompatibleMethodOverride]
-    correctness_prompt: Prompt = field(default_factory=lambda: CORRECTNESS_PROMPT)
+    correctness_prompt: Prompt = field(default_factory=CORRECTNESS_PROMPT.factory)
     long_form_answer_prompt: Prompt = field(
-        default_factory=lambda: LONG_FORM_ANSWER_PROMPT
+        default_factory=LONG_FORM_ANSWER_PROMPT.factory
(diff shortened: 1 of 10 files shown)


src/ragas/metrics/_answer_correctness.py
```diff
@@ -158,9 +159,9 @@ class AnswerCorrectness(MetricWithLLM, MetricWithEmbeddings):
 
     name: str = "answer_correctness"  # type: ignore[reportIncompatibleMethodOverride]
     evaluation_mode: EvaluationMode = EvaluationMode.qga  # type: ignore[reportIncompatibleMethodOverride]
-    correctness_prompt: Prompt = field(default_factory=lambda: CORRECTNESS_PROMPT)
+    correctness_prompt: Prompt = field(default_factory=correctness_prompt_factory)
     long_form_answer_prompt: Prompt = field(
-        default_factory=lambda: LONG_FORM_ANSWER_PROMPT
+        default_factory=long_form_answer_prompt_factory
```
jjmachan, line 164: what if we make a factory inside `Prompt` object and call it `init()` or `factory()`.

ssoima: I think `init` goes a bit against the scope of `__init__`  and is misleading. The cleanest way would be to create a factory class, but that would add more complexity, so I'd say having a factory method is a good middle-way.

ssoima:
@jjmachan , the failing lint check is not related to my code changes

ssoima:
@jjmachan are you still following this?

jjmachan:
hey @ssoima with the new Prompt object this has been fixed

PR #1216, opened 2024-08-23
fix : solves #1215 - Inconsistency in the Value of EVALMODE_TO_COLUMNS[EvaluationMode.ca]

### LINKED ISSUE #1215: Inconsistency in the Value of EVALMODE_TO_COLUMNS[EvaluationMode.ca]
Hi,

It seems there might be an inconsistency in the value of the dictionary `EVALMODE_TO_COLUMNS[EvaluationMode.ca]` within `src/ragas/validation.py`. Currently, the value is set to `["context", "summary"]`

https://github.com/explodinggradients/ragas/blob/d5b60bba475fb099b3af2544d80412ffc6e8abc8/src/ragas/validation.py#L62-L71

but I believe it should be `["context", "answer"]` as in `src/ragas/metrics/base.py:get_required_columns()`:

https://github.com/explodinggradients/ragas/blob/d5b60bba475fb099b3af2544d80412ffc6e8abc8/src/ragas/metrics/base.py#L38-L59

Is this the expected behavior? If not, could you please take look into it?

Thanks!

Best regards,
Juan

jucgonzalezes:
Sets `EVALMODE_TO_COLUMNS[EvaluationMode.ca] =  ["contexts", "answer"]` in `src/ragas/validation.py` to fix #1215.

### DIFF
--- a/src/ragas/validation.py
+++ b/src/ragas/validation.py
@@ -67,7 +67,7 @@ EVALMODE_TO_COLUMNS = {
     EvaluationMode.ga: ["ground_truth", "answer"],
     EvaluationMode.qga: ["question", "ground_truth", "answer"],
     EvaluationMode.qcg: ["question", "contexts", "ground_truth"],
-    EvaluationMode.ca: ["contexts", "summary"],
+    EvaluationMode.ca: ["contexts", "answer"],
 }
 
 


jjmachan (approved):
thanks a lot for the fix @jucgonzalezes 🙂

PR #1219, opened 2024-08-24
Tiny typo fix: it's -> its

peterkaminski:
A non-urgent, non-critical fix from contraction to possessive.

Elsewhere there is also an "it's" in `Question: Where is France and what is it's capital?`, but since that's a prompt a human might reasonably type, and in the answer "its" is used correctly, I thought it didn't need to be fixed.


jjmachan:
thanks for the PR!

PR #1223, opened 2024-08-26
Bugfix of evaluate function in llama_index.py

vcim:
Dict in dataset of the evaluate function specified. Does not work -> Dataset and then check of "ground_truth" in column_names.


jjmachan (approved):
thank you for the improvement 🙂

PR #1227, opened 2024-08-27
bugfix: fixed prompt adapt from English to other languages

yifan:
Issue:

the adaptation failed to parse the result of translation, resulting the adapted prompts not usable.  Specifically:

-  A indentation error causes json loaded output to be overwritten by raw string, causing output for json in adaptation to be not readable

```
   ```json
   {
     "blah": 1
   }
   
```

- translation output for str_translation needs to be extracted before putting back to prompt. we fix it by cleaning llm translation output.
- the code assumes use str_translation for every value except output, it does not take care of contexts, which is an array of string. deal with translation for non-output items such as ```contexts: [ str1, str2, str3]```, fix with json_loader.

Solution:

- fix indentation error
- parse translation results correctly by only taking the translation result
- check field value, and use json_loader with json_translation when the value in original prompt is a list instead of str

### DIFF
--- a/src/ragas/llms/prompt.py
+++ b/src/ragas/llms/prompt.py
@@ -191,6 +191,7 @@ class Prompt(BaseModel):
 
         logger.info("Adapting %s to %s", self.name, language)
         prompts = []
+        prompt_types = []
         output_keys = []
         for example in self.examples:
             prompts.extend(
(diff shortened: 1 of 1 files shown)


jjmachan:
thanks for this great PR - we did make the improvements in v0.2, do check it out

PR #1249, opened 2024-09-05
Add support to MultiTurn Metrics

shahules786:
```python
from ragas import evaluate
from ragas.metrics import rubrics_score_without_reference
from ragas.dataset_schema import EvaluationDataset, MultiTurnSample
from ragas.messages import HumanMessage

sample1 = MultiTurnSample(user_input=[HumanMessage(content="What is X")])
sample2 = MultiTurnSample(user_input=[HumanMessage(content="What is X")])
ds = EvaluationDataset(samples=[sample1, sample2])
evaluate(ds,metrics=[rubrics_score_without_reference])
```


src/ragas/evaluation.py
```diff
@@ -240,17 +248,32 @@ def evaluate(
...
+            _ = [
+                executor.submit(
+                    metric.single_turn_ascore,
+                    sample,
+                    row_group_cm,
+                    name=f"{metric.name}-{i}",
+                    timeout=run_config.timeout,
+                )
+                for metric in metrics
+                if isinstance(metric, SingleTurnMetric)
```
jjmachan, line 261: what happens if it is not `SingleTurnMetric` it will fail silently confusing the user. It is better to do this in the validation step ealier and then we can safely assume it is SingleTurn

src/ragas/metrics/domain_specific_rubrics/without_reference.py
```diff
@@ -0,0 +1,164 @@
...
+            callbacks=callbacks,
+        )
+        return output.score
+
+    def _create_multi_turn_prompt(self, row: t.Dict) -> MultiTurnWithoutReferenceInput:
+        interaction = row["interaction"]
+        return MultiTurnWithoutReferenceInput(
+            user_input=interaction,
+            rubrics=self.rubrics,
+        )
```
jjmachan, line 143: why do we need this 2 line function?

jjmachan (commented):
just a couple of changes needed

src/ragas/evaluation.py
shahules786: That validation function is this one
https://github.com/explodinggradients/ragas/blob/fe684673a9b79d11ea1e3fe04c18a27696788146/src/ragas/evaluation.py#L168

src/ragas/metrics/_answer_similarity.py
```diff
@@ -38,7 +43,9 @@ class AnswerSimilarity(MetricWithLLM, MetricWithEmbeddings, SingleTurnMetric):
     """
 
     name: str = "answer_similarity"  # type: ignore
-    _required_columns: t.Tuple[str, ...] = ("reference", "response")
+    _required_columns: t.Dict[MetricType, t.Set[str]] = field(
+        default_factory=lambda: {MetricType.SINGLE_TURN: {"reference", "response"}}
+    )
```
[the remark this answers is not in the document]
shahules786: makes sense. What would the default value then be?

PR #1260, opened 2024-09-09
Non LLM based metrics

shahules786:


pyproject.toml
```diff
@@ -19,6 +19,10 @@ dynamic = ["version", "readme"]
 all = [
     "sentence-transformers",
     "transformers",
+    "nltk",
+    "rouge_score",
+    "fuzzywuzzy",
+    "rapidfuzz",
```
jjmachan, line 25: we need to document this- that you can install with `pip install ragas[all]` for these metrics

PR #1266, opened 2024-09-10
feat: add non llm based context recall

shahules786:


src/ragas/metrics/_context_recall.py
```diff
@@ -213,4 +216,78 @@ def save(self, cache_dir: str | None = None) -> None:
         self.context_recall_prompt.save(cache_dir)
 
 
+@dataclass
+class ContextRecall(LLMContextRecall):
+    name: str = "context_recall"
+
+    @deprecated(since="0.2", removal="0.3", alternative="LLMContextRecall")
```
jjmachan, line 223: why is this depricated?

jjmachan (approved):
just the random deprecation warning

src/ragas/metrics/_context_recall.py
shahules786: `LLMContextRecall` is the new ContextRecall.

PR #1272, opened 2024-09-11
feat: general purpose metrics

shahules786:
Do not merge
Splitting this into new 3 PRs


docs/concepts/metrics/general_purpose.md
```diff
@@ -0,0 +1,146 @@
...
+    reference="The Eiffel Tower is located in Paris.",
+)
+
+critic =  AspectCritique(
+    name="maliciousness",
+    definition="Is the submission intended to harm, deceive, or exploit users?",
+)
+await critic.single_turn_ascore(sample)
+```
+**How it works**
```
jjmachan, line 27: use headings?
```suggestion
### How it works
```

docs/concepts/metrics/general_purpose.md
```diff
@@ -0,0 +1,146 @@
...
+```{code-block} python
+from ragas.dataset_schema import SingleTurnSample
+from ragas.metrics._domain_specific_rubrics import RubricsScoreWithReference
+sample = SingleTurnSample(
+    user_input="Where is the Eiffel Tower located?",
+    response="The Eiffel Tower is located in Paris.",
+    reference="The Eiffel Tower is located in Paris.",
+)
+
+scorer =  RubricsScoreWithReference()
```
jjmachan, line 65: it would be great to show the example of the default rubric - maybe as a note?

src/ragas/dataset_schema.py
```diff
@@ -24,12 +24,13 @@ class SingleTurnSample(BaseEvalSample):
...
     multi_responses: t.Optional[t.List[str]] = None
     reference: t.Optional[str] = None
-    rubric: t.Optional[t.Dict[str, str]] = None
+    rubrics: t.Optional[t.Dict[str, str]] = None
 
 
 class MultiTurnSample(BaseEvalSample):
     user_input: t.List[t.Union[HumanMessage, AIMessage, ToolMessage]]
     reference: t.Optional[str] = None
+    rubrics: t.Optional[t.Dict[str, str]] = None
```
jjmachan, line 33: if something is `Optional` the default can be something like an empty dict too - doesn't have to by None always. this is a pattern we have in the codebase, something we should think about

jjmachan (approved):

src/ragas/dataset_schema.py
shahules786: An argument for None is, that it is easy to check and verify if the value is None in case the metric requires a rubric to be specified.

PR #1279, opened 2024-09-11
Add option to disable progress bar in `evaluate`

freddiev4:
Hello there :wave: I noticed that there was no way to disable the progress bar and thought it'd nice to have the option to silence the `tqdm` call in the executor, as it doesn’t play nicely with logs and adds noise.

I've added the following changes:

Add `show_progress` flag to `Executor` class and update `evaluate` function to use it.

* **Executor Class Changes:**
  - Add `show_progress` flag to the `Executor` class in `src/ragas/executor.py`.
  - Use the `show_progress` flag to control the display of the progress bar.
  - If `show_progress` is set to False, disable `tqdm` usage for displaying progress bars.

* **Evaluate Function Changes:**
  - Add `show_progress` parameter to the `evaluate` function in `src/ragas/evaluation.py`.
  - Pass the `show_progress` flag to the `Executor` class during initialization in the `evaluate` function.

### DIFF
--- a/src/ragas/evaluation.py
+++ b/src/ragas/evaluation.py
@@ -66,6 +66,7 @@ def evaluate(
     token_usage_parser: t.Optional[TokenUsageParser] = None,
     raise_exceptions: bool = False,
     column_map: t.Optional[t.Dict[str, str]] = None,
+    show_progress: bool = True,
 ) -> Result:
     """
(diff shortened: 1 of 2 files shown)


jjmachan (approved):
thanks a lot for this improvement @freddiev4 ❤️ 🙌🏽

PR #1300, opened 2024-09-13
Fix wrong input key name

ruanhao:
ValueError: The metric [context_utilization] that that is used requires the following additional columns ['retreived_contexts'] to be present in the dataset.


shahules786:
Nice catch @ruanhao I'm not at all good with spellings - I'm sure one could find more of these. Appreciate the help.

PR #1204, opened 2024-08-16
Fix ensure_ascii for json.dumps

ifsheldon:
Hi! This PR simply fixes all `json.dumps`. With `ensure_ascii=False`, unicodes can be properly serialized into JSON files in the modern days when UTF-8 is basically universal. This may also prevent silent behavioral bugs from LLMs, since they can hardly understand literal strings like "\u00x" except common ones.

Some cases do not necessarily need `ensure_ascii=False`, but they are added for consistency.

PR #1305, opened 2024-09-13
Fixed the issue of changing to Unicode characters in the json.dumps

Youngrok123:
When using json.dumps, if the character is not an ASCII character, it is converted to a Unicode character and output or passed to LLM. Due to this issue, LLM is unable to respond properly.

Added ensure_ascii=False option to output Unicode characters literally.

### DIFF
--- a/src/ragas/dataset_schema.py
+++ b/src/ragas/dataset_schema.py
@@ -87,7 +87,7 @@ class EvaluationDataset(BaseModel):
             for sample in rows:
                 for item in sample["user_input"]:
                     if not isinstance(item["content"], str):
-                        item["content"] = json.dumps(item["content"])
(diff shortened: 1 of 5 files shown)


Youngrok123:
[Fix ensure_ascii for json.dumps](https://github.com/explodinggradients/ragas/pull/1204)
Since the code has not been merged, I am sending a PR by modifying the code in the current ragas version.

wildgeece96:
This is a great PR. Thanks @Youngrok123 .

If we want to evaluate in language that doesn't support ASCII like Japanese and Korean, ragas makes prompt with unicode-like strings.
This library is really great, so I hope that it can be applicable to all languages.

image

jjmachan:
Thanks again @Youngrok123 for fixing this 🙂

PR #1345, opened 2024-09-23
feat: transforms for Knowledge Graphs

jjmachan:
### Transforms

There are basically 3 type of transforms you can do

1. extract new properties for nodes and relationships - `Extrators`
2. split existing nodes into new nodes and relationships - `Splitters`
3. find and build new relationships in the knowledge graph - `RelationshipBuilder`


src/ragas/experimental/testset/transforms/extractors/regex_based.py
```diff
@@ -0,0 +1,43 @@
...
+            raise ValueError(
+                f"node.property('page_content') must be a string, found '{type(text)}'"
+            )
+
+        matches = re.findall(self.pattern, text, re.MULTILINE)
+        return self.property_name, matches
+
+
+# This regex pattern matches URLs, including those starting with "http://", "https://", or "www."
+links_extractor_pattern = r"(?i)\b(?:https?://|www\.)\S+\b"
```
jjmachan, line 27: here we are initing these extractors - should we change that?

shahules786: Yes, please. Such inits can't be scaled cleanly.

src/ragas/experimental/testset/transforms/splitters/headline.py
```diff
@@ -0,0 +1,52 @@
+import typing as t
+from dataclasses import dataclass
+
+from ragas.experimental.testset.graph import Node, Relationship
+from ragas.experimental.testset.transforms.base import Splitter
+
+
+@dataclass
+class HeadlineSplitter(Splitter):
```
shahules786, line 9: This I see handles single-level headline splitting, ideally, we might need an iterative one but let's leave that for now. Also, we would need normal hierarchical splitters like ones in llama-index or langchain.

src/ragas/experimental/testset/transforms/extractors/llm_base.py
```diff
@@ -0,0 +1,160 @@
...
+            StringIO(text="Deep Learning for Natural Language Processing"),
+        )
+    ]
+
+
+class Headlines(BaseModel):
+    headlines: t.Dict[str, t.List[str]]
+
+
+class HeadlinesExtractorPrompt(PydanticPrompt[StringIO, Headlines]):
```
shahules786, line 72: Can you replace the example here with a more generalizable and shorter one? this one targets Arxiv documents specifically.

shahules786 (approved):

src/ragas/experimental/testset/transforms/splitters/headline.py
jjmachan: I've noted these down these will have to address them is separate PRs

PR #1350, opened 2024-09-24
wrap as_complete with async

liupgd:
Originally as_complete is not defined with async, the asyncio.Semaphore is created in main thread not in asynci.run. This will cause 'There is no event loop' error when: 
1. len(coros) is more than max_workers; 
2. run 'evalute' function at the second time and nest_asyncio is not applied.

### DIFF
--- a/src/ragas/executor.py
+++ b/src/ragas/executor.py
@@ -23,7 +23,7 @@ def is_event_loop_running() -> bool:
         return loop.is_running()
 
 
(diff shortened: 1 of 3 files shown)


jjmachan:
@liupgd thanks a lot for putting this PR together 🙂 

let me just benchmark it on my end too before merging it in

jjmachan:
thanks a lot @liupgd ❤️

PR #1352, opened 2024-09-24
feat: transform engine to execute transforms

jjmachan:
Code to orchestrate your transforms

```py
from ragas.experimental.testset.transforms import Parallel, TransformerEngine
from ragas.experimental.testset.transforms.splitters import HeadlineSplitter
from ragas.experimental.testset.transforms.extractors import SummaryExtractor, KeyphrasesExtractor, TitleExtractor, HeadlinesExtractor, EmbeddingExtractor

# define the transforms
summary_extractor = SummaryExtractor()
keyphrase_extractor = KeyphrasesExtractor()
title_extractor = TitleExtractor()
headline_extractor = HeadlinesExtractor()
embedding_extractor = EmbeddingExtractor()
headline_splitter = HeadlineSplitter()

# specify the transforms and their order to be applied
transforms = [
    headline_extractor,
    headline_splitter,
    Parallel(
        embedding_extractor,
        summary_extractor
    )
]

TransformerEngine().apply(transforms, kg)
```

PR #1359, opened 2024-09-25
feat: adding an implementation of abstractQA

jjmachan:
Once you have a kg with #1352  you can run this like
```py
from ragas.experimental.testset.generators.abstract import AbstractGenerator

abstract_qa = AbstractGenerator()
dist = await abstract_qa.generate_distributions(n=10, knowledge_graph=kg)
q = await abstract_qa.generate_user_input(dist[0])
```

merge after merging #1352


shahules786 (approved):
I assume quality checks and improvements on prompts and workflows can be done later.

PR #1366, opened 2024-09-26
fix: fixed verdict comparison in evolutions.py

MarkHmnv:
The bug is located in the ragas.testset.Evolution class. The problem is that you are checking the ‘verdict’ field against a string value of ‘-1’, and you need to compare it against an int value of -1:
{629CB6B1-CCF2-4FD0-988E-60BE9635D6FA}
{0CD30A93-DBB5-49AA-833D-1169960D694E}

So I changed the verdict comparison from a string to an integer in the conditional check.

### DIFF
--- a/src/ragas/testset/evolutions.py
+++ b/src/ragas/testset/evolutions.py
@@ -237,7 +237,7 @@ class Evolution:
         answer = answer if isinstance(answer, dict) else {}
         logger.debug("answer generated: %s", answer)
         answer = (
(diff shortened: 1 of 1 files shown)


jjmachan:
hey @MarkHmnv - firstly thanks a lot for adding the PR but that version of testset generation is going to be removed in the next version but still merging it in

in the meantime would love to get your opinion on the new one if you have time 🙂 . You're valuable feedback would be awesome!

jjmachan (approved):

PR #1376, opened 2024-09-28
feat: added more simulators

jjmachan:
added  2 more simulators

```py
def default_scenarios(llm: BaseRagasLLM) -> QuestionTypes:
    return [
        (AbstractQuestionSimulator(llm), 0.25),
        (ComparativeAbstractQuestionSimulator(llm), 0.25),
        (SpecificQASimulator(llm), 0.5),
    ]
```


shahules786 (approved):
Just one comment @jjmachan , The output of the test generator should also contain what kind of scenario each sample is.  This will be useful when filtering and analysing eval results.

PR #1377, opened 2024-09-28
Draft PR, Need to address the issue with plugin for llm switcher

joelk9895:
Plugin files are located in the root of the directory, named as llm switch


jjmachan:
figured out another way

PR #1380, opened 2024-09-29
Update models in Ragas documentation: change gpt-3.5-turbo-16k to gpt-4o-mini and gpt-4 to gpt-4o

suekou:
According to the OpenAI API reference, As of July 2024, gpt-4o-mini should be used in place of gpt-3.5-turbo, as it is cheaper, more capable, multimodal, and just as fast. gpt-3.5-turbo is still available for use in the API, but the advantages of gpt-4o-mini make it a better choice.

Additionally, the cost of using gpt-4 in the API is too high. Switching to gpt-4o is a more economical option without compromising performance. Personally, I had a significant chunk of my credits consumed due to the high cost of gpt-4 😢.

### DIFF
--- a/docs/concepts/testset_generation.md
+++ b/docs/concepts/testset_generation.md
@@ -61,8 +61,8 @@ from langchain_openai import ChatOpenAI, OpenAIEmbeddings
 # documents = load your documents
 
 # generator with openai models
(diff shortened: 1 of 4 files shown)


jjmachan:
Thank you so much for fixing this Documentation @suekou 🙏🏽

PR #1382, opened 2024-09-30
fix: aspect_critique documentation

### LINKED ISSUE #1374: No module named 'ragas.metrics.critique'

**Describe the bug**
Trying to import `ragas.metrics.critique` results in `ModuleNotFoundError: No module named 'ragas.metrics.critique'`.
Latest working version seems to be 0.1.18. Starting in 0.1.19 the error occurs. I assume this PR broke the behavior: #1286.

It seems that most of the functionality has been moved to `ragas.metric._aspect_critic`. However, this is a private module, which implies that a user should not import from it directly. But the pre-defined AspectCritics (harmfulness, maliciousness, etc.) which could previously be imported from `ragas.metric.critique` are only importable from this module now.

Also, many example notebooks in the `docs` folder still include references to `ragas.metric.critique`.

Ragas version: 0.1.20
Python version: 3.11.7

**Code to Reproduce**
```python
>>> import ragas.metric.critique
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'ragas.metric'
```

**Error trace**
See above

**Expected behavior**
The pre-defined AspectCritics should be importable from a public module. Optimally, they would be importable the exact same way as before to ensure backwards compatibility with older 0.1.x versions, since no breaking have been indicated by the release notes or version number.
```python
from ragas.metrics.critique import harmfulness

# Suggested Alternatives
from ragas.metrics.aspect_critic import harmfulness
from ragas.metrics import harmfulness
```

**Additional context**
As a workaround, the metrics can be imported from the private module:
```python
from ragas.metrics._aspect_critic import harmfulness
```

jjmachan:
fixes #1374 

but a couple of things to note here
- metrics should exposed from `ragas.metrics` because all our metrics modules are private with `_` prefix. 
- need to mention this in the migration guide, about aspect critic

### DIFF
--- a/docs/concepts/metrics/general_purpose.md
+++ b/docs/concepts/metrics/general_purpose.md
@@ -10,7 +10,7 @@ Aspect critic is an evaluation metric that can be used to evaluate responses bas
 
 ```python
 from ragas.dataset_schema import SingleTurnSample
-from ragas.metrics.critique import AspectCritique
+from ragas.metrics import AspectCritic
 
 sample = SingleTurnSample(
     user_input="Where is the Eiffel Tower located?",
@@ -18,7 +18,7 @@ sample = SingleTurnSample(
     reference="The Eiffel Tower is located in Paris.",
 )
 
-critic =  AspectCritique(
+critic =  AspectCritic(
     name="maliciousness",
     definition="Is the submission intended to harm, deceive, or exploit users?",
 )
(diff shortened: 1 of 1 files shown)


shahules786 (approved):
There are some other parts same mistake was done. I'll correct it before release.

PR #1386, opened 2024-09-30
feat: output new Testset object for generation

jjmachan:
image

### DIFF
--- a/src/ragas/experimental/testset/simulators/generate.py
+++ b/src/ragas/experimental/testset/simulators/generate.py
@@ -4,10 +4,10 @@ import logging
 import typing as t
 from dataclasses import dataclass, field
 
-from ragas.dataset_schema import EvaluationDataset
 from ragas.executor import Executor
 from ragas.experimental.testset.graph import KnowledgeGraph, Node, NodeType
 from ragas.experimental.testset.simulators import default_scenarios
+from ragas.experimental.testset.simulators.testset_schema import Testset, TestsetSample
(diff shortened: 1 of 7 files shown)


src/ragas/experimental/testset/simulators/generate.py
```diff
@@ -71,7 +71,7 @@ def generate_with_langchain_docs(
 
         return self.generate(
             test_size=test_size,
-            scenarios=scenarios,
+            simulators=scenarios,
```
shahules786, line 74: Using simulators=scenarios created confusion on what is what. This is also present as an argument.

src/ragas/experimental/testset/simulators/generate.py
```diff
@@ -80,21 +80,21 @@ def generate_with_langchain_docs(
     def generate(
         self,
         test_size: int,
-        scenarios: t.Optional[QuestionTypes] = None,
+        simulators: t.Optional[QuestionTypes] = None,
```
shahules786, line 83: Also here, if it's a simulator what is then QuestionTypes?

shahules786 (changes requested):

src/ragas/experimental/testset/simulators/generate.py
jjmachan: changed this to `simulator_distributions` - how does that sound? is it clear?

shahules786 (approved):

PR #1389, opened 2024-10-01
added better styling to reder jupyter notebooks nicely.

zodwick:
especially fixed the spacing issues.
also deleted an unused block in one of the doc pages.


jjmachan:
great job @zodwick!

just one more thing regarding the code blocks

jupyter
image

md
image


can we make it the same font and size? I think the code block itself is a bit too large (the copy botton is larger too)

shahules786:
I observed that mkdocs [content tabs](https://squidfunk.github.io/mkdocs-material/reference/content-tabs/#grouping-code-blocks-c_1) used to set different LLMs  does not get rendered in notebook https://github.com/explodinggradients/ragas/pull/1398

jjmachan:
those are not supported in mkdocs-jupyter actually, we will have to render it as html [danielfrg/mkdocs-jupyter: Use Jupyter Notebook in mkdocs](https://github.com/danielfrg/mkdocs-jupyter?tab=readme-ov-file#mkdocs-material-notes)

zodwick:
i made the clipboard icon smaller, but mkdocs-jupyter uses a different svg compared to normal md so there is going to be slight differences
jupyter vs md as of now. close but not exact

image
image

@jjmachan @shahules786

jjmachan (approved):
this looks great!

PR #1396, opened 2024-10-01
non-asii characters support in faithfulness

gargewp:
There is a bug with json.dumps: non-ascii charecters are converting into non-readable symbols.

It affects Faithfulness and NoiseSensitivity only.


shahules786 (approved):
Nice one @gargewp , thanks for the PR

PR #1404, opened 2024-10-02
Fix typo in metric error message

suekou:
This pull request addresses a minor typo in the error message related to metric validation. The message previously contained a duplicated "that," which has been corrected for clarity.


shahules786:
Nice catch @suekou , thanks a lot

PR #1406, opened 2024-10-02
feat: NER extractor and Jaccard

shahules786:


src/ragas/experimental/testset/transforms/extractors/llm_based.py
```diff
@@ -160,3 +192,16 @@ async def extract(self, node: Node) -> t.Tuple[str, t.Any]:
...
+class NERExtractor(LLMBasedExtractor):
+    property_name: str = "entities"
+    prompt: NERPrompt = NERPrompt()
+
+    async def extract(self, node: Node) -> t.Tuple[str, t.Any]:
+        node_text = node.get_property("page_content")
+        if node_text is None:
+            return self.property_name, None
+        result = await self.prompt.generate(self.llm, data=StringIO(text=node_text))
+        return self.property_name, result.entities.model_dump()
```
jjmachan, line 207: is this the correct return type?
(property_name: str, property_value: t.Any)

shahules786:
Why is it throwing pydantic version related error when this PR does not modify anything related to it?

jjmachan:
This has something to do with the order of running tests. I changed that in the main which is why this seems to be fixed in other PRs

PR #1408, opened 2024-10-02
docs:Auto Gen of Reference Page

zodwick:
added auto gen for reference pages plus the  makefile update for the same.
There is some issues as a lot of  pages ( both __init__ and otherwise ) are empty.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,6 +34,8 @@ test-e2e: ## Run end2end tests
 	
 # Docs
 docsite: ## Build and serve documentation
+	@echo "Generating reference pages..."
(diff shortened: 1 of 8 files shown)


jjmachan:
quick tip: try to keep PRs with: [docs:Auto Gen of Reference Page by zodwick · Pull Request #1408 · explodinggradients/ragas](https://github.com/explodinggradients/ragas/pull/1408)

jjmachan (approved):
edda everything looks good but can you make it flatter?

image
right now everything is under ragas folder, could we drop that?

PR #1410, opened 2024-10-02
fix key error for FaithfulnesswithHHEM

linzzzzzz:
The current implementation for evaluating FaithfullnessWithHHEM requires legacy 'contexts' key and would throw out key error. This simple fix would resolve the issue.


jjmachan:
thanks a lot @linzzzzzz for this fix!
which version are you using?

Issue #1434, opened 2024-10-06
[0.2.0-beta] issue with faithfulness metric - Exception raised in Job[1]: AttributeError('dict' object has no attribute 'verdict')


linzzzzzz:
[ ] I have checked the [documentation](https://docs.ragas.io/) and related resources and couldn't resolve my bug.

**Describe the bug**
Came across error when running very basic evaluation on faithfulness metric.

Ragas version: 0.2.0-beta
Python version: 3.10.13

**Code to Reproduce**
Share code to reproduce the issue
```
import os

# Set the environment variable
os.environ['RAGAS_DO_NOT_TRACK'] = 'true'

# You can now access the environment variable like this:
do_not_track = os.environ.get('RAGAS_DO_NOT_TRACK', 'false')
print(f"RAGAS_DO_NOT_TRACK is set to: {do_not_track}")

from datasets import Dataset, load_dataset
import os
from ragas import evaluate
from ragas.metrics import faithfulness, answer_correctness
import json

from langchain_community.chat_models import BedrockChat
from langchain_community.embeddings import BedrockEmbeddings
import boto3
from ragas import evaluate
import nest_asyncio  # CHECK NOTES

from ragas.metrics import (
    faithfulness,
)
from ragas.run_config import RunConfig
from botocore.config import Config

import pandas as pd
import ragas
import numpy as np

# Create a RunConfig with a specified timeout
bedrock_config = Config(read_timeout=1000)

config = {
    "credentials_profile_name": "your-profile-name",  # E.g "default"
    "region_name": "us-west-2",  # E.g. "us-east-1"
    "model_id":"anthropic.claude-3-sonnet-20240229-v1:0",
    "model_kwargs": {"temperature": 0.4},
}

def get_bedrock_client(region):
    session = boto3.Session(profile_name='personal')
    bedrock_client = session.client("bedrock-runtime", region_name=region, config=bedrock_config)
    return bedrock_client

bedrock_client = get_bedrock_client('us-west-2')

bedrock_model = BedrockChat(
    region_name=config["region_name"],
    # endpoint_url=f"https://bedrock-runtime.{config['region_name']}.amazonaws.com",
    client=bedrock_client,
    model_id=config["model_id"],
    model_kwargs=config["model_kwargs"],
)

# init the embeddings
bedrock_embeddings = BedrockEmbeddings(
    # credentials_profile_name=config["credentials_profile_name"],
    client=bedrock_client,
    region_name=config["region_name"],
)

[...]

PR #1436, opened 2024-10-06
fix attribute error for faithfulness metric

linzzzzzz:
fix attribute error in https://github.com/explodinggradients/ragas/issue…](https://github.com/explodinggradients/ragas/issues/1434.


jjmachan:
thanks a lot @linzzzzzz - merging it in!

Issue #1441, opened 2024-10-07
Answer Correctness not working after pydantic update


hubkrieb:
[ ] I have checked the [documentation](https://docs.ragas.io/) and related resources and couldn't resolve my bug.

**Describe the bug**

Ragas version: 0.2.0
Python version: 3.12

**Error trace**
AttributeError('StatementsAnswers' object has no attribute 'dicts')

**Additional context**
The dicts method was deleted from the StatementAnswers class after the pydantic update 95f8d7bf590f19f1d0fc302d3e64556ae5e49f18

PR #1442, opened 2024-10-07
fix: add updated dicts method to StatementAnswers

hubkrieb:
#1441 Added back the dicts method to StatementAnswers similarly to the one in StatementFaithfulnessAnswers to be able to evaluate with the answer_correctness metric

### DIFF
--- a/src/ragas/metrics/_faithfulness.py
+++ b/src/ragas/metrics/_faithfulness.py
@@ -42,9 +42,6 @@ class Statements(BaseModel):
 class StatementsAnswers(RootModel):
     root: t.List[Statements]
 
-    def dicts(self):
-        return self.model_dump()
-
 
 _statements_output_instructions = get_json_format_instructions(StatementsAnswers)
 _statements_output_parser = RagasOutputParserOld(pydantic_object=StatementsAnswers)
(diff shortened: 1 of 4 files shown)


src/ragas/metrics/_faithfulness.py
```diff
@@ -42,6 +42,9 @@ class Statements(BaseModel):
 class StatementsAnswers(RootModel):
     root: t.List[Statements]
 
+    def dicts(self):
+        return self.model_dump()
+
```
jjmachan, line 47: Nice catch! the proper fix would be to replace the lines that use this function with `self.model_dump()` to stay consistant with the rest of the codebase.

could you rework the PR so that it changes those lines?

hubkrieb: I just reworked it! I noticed that the dicts method was still used in other parts of the code base (for the faithfulness metric and in noise sensitivity) so I also replaced it there

jjmachan:
@hubkrieb fixed the whole with #1462 

thanks a lot for taking time but I'm sorry I couldn't merge this in 🙁 

let me know if there is anything else I can help with

PR #1482, opened 2024-10-13
docs: moved from ipnb files to md files

jjmachan:
From now on documentation will be expected to written as .md files 

we have a script to do the conversion `make ipynb-to-md` that will do the conversion. 
- all generated `md` files should be have the `_` prefix so that we know it's a converted files and **generated md files are not expected to be edited manually.**
- We can now use all the md formats available (but should have worked otherwise too) 
image

you can visualize the diff here
- https://docs.ragas.io/en/latest/howtos/customizations/metrics/write_your_own_metric/ (Generated from ipynb)
- https://ragas--1482.org.readthedocs.build/en/1482/howtos/customizations/metrics/_write_your_own_metric/ (Generated from .md)

Do compare them side by side before merging


shahules786 (changes requested):
There are places in which the these ipynb files were linked to which have to be changed. For example https://ragas--1482.org.readthedocs.build/en/1482/howtos/customizations/
Links under metrics does not work.

PR #1487, opened 2024-10-14
Update Bedrock modules for LangChain v0.3.x

minorun365:
Updating LangChain modules for Amazon Bedrock (AWS) in "customizations" section.

- Use `langchain_aws` instead of `langchain_community.`
- Use `ChatBedrock` instead of `BedrockChat`

### DIFF
--- a/docs/extra/components/choose_evaluvator_llm.md
+++ b/docs/extra/components/choose_evaluvator_llm.md
@@ -21,19 +21,20 @@
     config = {
         "credentials_profile_name": "your-profile-name",  # E.g "default"
         "region_name": "your-region-name",  # E.g. "us-east-1"
-        "model_id": "your-model-id",  # E.g "anthropic.claude-v2"
-        "model_kwargs": {"temperature": 0.4},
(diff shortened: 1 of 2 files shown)


shahules786 (changes requested):
Hey @minorun365 , thanks for this. Would you nice if you can make this change in this [file](https://github.com/explodinggradients/ragas/blob/main/docs/extra/components/choose_evaluvator_llm.md) too which is used in several docs to switch to AWS

minorun365:
> Hey @minorun365 , thanks for this. Would you nice if you can make this change in this [file](https://github.com/explodinggradients/ragas/blob/main/docs/extra/components/choose_evaluvator_llm.md) too which is used in several docs to switch to AWS

@shahules786 Of course! Along with that, I've also updated `ChatBedrock` to the even more recent `ChatBedrockConverse`.

jjmachan (approved):
LG from my end!
thanks alot @minorun365 for helping us improve the docs 🙂 ❤️

shahules786 (approved):
LGTM. Thanks bro :)

PR #1490, opened 2024-10-14
Fixed an issue where non-ASCII characters were changed to Unicode characters within the prompt

Youngrok123:
- There is an issue where non-ASCII characters are converted to Unicode characters within the prompt.
- Due to this issue, Users that use non-English characters will receive an abnormal response when communicating with LLM.
- This issue can be resolved by adding the ensure_ascii=False option when using the json.dumps function.


jjmachan:
@Youngrok123 also created something for v0.2 prompts https://github.com/explodinggradients/ragas/pull/1494

do let me know if there is any mistakes :)

PR #1492, opened 2024-10-14
feat: added save and load to RagasDataset

jjmachan:
- save and load
- Migration docs


docs/getstarted/rag_evaluation.md
```diff
@@ -16,16 +16,7 @@ Converting data to ragas [evaluation dataset](../concepts/components/eval_datase
 ```python
 from ragas import EvaluationDataset, SingleTurnSample
```
shahules786, line 17: remove SingleTurnSample as it's not used.

shahules786 (approved):

PR #1502, opened 2024-10-15
docs: fix broken links in testset generation custimisation

jjmachan:


MariosKef:
Will there be any examples in the Testset Generation? It would be very useful if you could show a minimal case. Thanks!

jjmachan:
@MariosKef will add them, thanks for the suggestion! 🙂

PR #1503, opened 2024-10-15
docs: add faq

shahules786:


jjmachan:
image
the toggle is not correct orientation

is doesn't mkdocs support toggle by default - which resources are you checking?

shahules786:
I couldn't find toggle list [here](https://squidfunk.github.io/mkdocs-material/reference/lists/?h=list)

Screenshot 2024-10-15 at 12 12 05 PM
Just fixed it. Also aligned it with rest of color scheme.

PR #1513, opened 2024-10-15
fix: propagate callbacks in RagasOutputParser

dlyz:
Callbacks are now correctly passed to `fix_output_format_prompt.generate()` during output format fixing in `RagasOutputParser`.


jjmachan:
thanks alot @dlyz - will release this is a few hours

PR #1520, opened 2024-10-16
fix: use get_temperature in base LLM generate

sapphire-arches:
Matches the behavior in agenerate, avoiding some API call failures


jjmachan (approved):

jjmachan:
thanks alot @sapphire-arches for the fix 🙂 ❤️

PR #1529, opened 2024-10-18
feat: translate instruction when adapting prompt

Yunnglin:
When adapting prompts, both the examples and the corresponding instructions are important and need to be translated.

### DIFF
--- a/src/ragas/prompt/mixin.py
+++ b/src/ragas/prompt/mixin.py
@@ -53,7 +53,7 @@ class PromptMixin:
             setattr(self, key, value)
 
     async def adapt_prompts(
-        self, language: str, llm: BaseRagasLLM
+        self, language: str, llm: BaseRagasLLM, adapt_instruction: bool = False
     ) -> t.Dict[str, PydanticPrompt]:
         """
         Adapts the prompts in the class to the given language and using the given LLM.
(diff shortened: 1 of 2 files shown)


shahules786:
Hey @Yunnglin Did you observe better results when translating both? We observed pretty good results with only translating a few shot examples. Another reason why we did not translate instruction was that time it introduces ambiguity. FOr example, let's say if the instruction was "Ouptu False if a number greater than zero", this would also translate the word "False", which then causes issues while post-processing.

Yunnglin:
Whether to translate the "instruction" could perhaps be made into an option. When I was generating a Chinese dataset, I found that some "extractor prompts" would lack examples, which led to the generated data not being very effective.

shahules786:
@Yunnglin that suggestion makes sense. Could you make a PR with the same? 
@jjmachan Please share if you have any opinion on this.

jjmachan:
This makes  a ton of sense 

what we can do is take another argument for https://github.com/explodinggradients/ragas/blob/807c313569ba22f69788249c20152a129c21b60b/src/ragas/prompt/pydantic_prompt.py#L217-L223

which converts it 

and it can be controlled through https://github.com/explodinggradients/ragas/blob/807c313569ba22f69788249c20152a129c21b60b/src/ragas/prompt/mixin.py#L55

would be really helpful 🙂

Yunnglin:
*  Add `adapt_instruction: bool=False` parameter.

    Now you can adapt prompt as follows:
    ```python
    import asyncio
    from ragas.metrics import Faithfulness
    from ragas.llms import LangchainLLMWrapper
    
    
    instance = Faithfulness()
    adapted_prompts = asyncio.run(instance.adapt_prompts(language="chinese", llm=LangchainLLMWrapper(chat_model), adapt_instruction=True))
    print(adapted_prompts)

    ```

*  Add `Ensure that the number of output data rows is equal to the number of input data rows.` to the translation prompt since sometimes LLM breaks the single line into multiple lines.

shahules786 (approved):
LGTM, thanks a lot.

PR #1533, opened 2024-10-18
fix error OpenAIEmbeddings' object has no attibute 'embed_text'

Yuri-Albuquerque:
Compatibility Between Different Embedding Interfaces:

We use the OpenAIEmbeddings API for embedding generation. However, there are variations in the API's interface across different versions and implementations.
In some versions, the embedding function is exposed as embed_text, while in others, it might be called embed_query.
Without this try-except block, calling the incorrect method would lead to an AttributeError, causing a failure in generating the necessary embeddings for the RAGAS computation.

Error Handling:

The try block first attempts to call the embed_text method, which is the expected API in some versions or implementations of the embeddings class.
If the embed_text method does not exist, an AttributeError is caught by the except block. In this case, we gracefully handle the error by switching to the alternative method, embed_query, which is expected to work in other versions.

Future-Proofing and Flexibility:

By adding this error handling, we make the code more resilient to future changes or variations in the API without requiring major refactoring.
This also improves the flexibility of the code by allowing it to work across different versions of the embedding libraries without modification.

Finalization with finally:

Regardless of which embedding method is called (embed_text or embed_query), the finally block ensures that the output (embeddings) is processed into the expected format (NumPy arrays).
This guarantees that the rest of the RAGAS pipeline remains unaffected by the variation in the embedding call.

### DIFF
--- a/src/ragas/metrics/_answer_similarity.py
+++ b/src/ragas/metrics/_answer_similarity.py
@@ -48,7 +48,7 @@ class SemanticSimilarity(MetricWithLLM, MetricWithEmbeddings, SingleTurnMetric):
     )
     is_cross_encoder: bool = False
     threshold: t.Optional[float] = None
-
+        
     def __post_init__(self: t.Self):
         # only for cross encoder
         if isinstance(self.embeddings, HuggingfaceEmbeddings):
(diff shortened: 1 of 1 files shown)


jjmachan:
hey @Yuri-Albuquerque how are you defining the embedding and passing it for Ragas?

Yuri-Albuquerque:
> hey @Yuri-Albuquerque how are you defining the embedding and passing it for Ragas?

```python
from langchain_openai.embeddings import OpenAIEmbeddings
from ragas.metrics import AnswerCorrectness

embeddings = OpenAIEmbeddings(
                               model="text-embedding-ada-002",
                               max_ritries=2,
                           )
and then I'm injecting in the class answer_correctness = AnswerCorrectness( ..., embeddings = embeddings) and also in 
evaluate(..., embeddings = embeddings, metrics=[answer_correctness])```

to get the scores

Yuri-Albuquerque:
> hey @Yuri-Albuquerque how are you defining the embedding and passing it for Ragas?
> 
> are you wrapping with
> 
> https://github.com/explodinggradients/ragas/blob/807c313569ba22f69788249c20152a129c21b60b/src/ragas/embeddings/base.py#L68
> 
> ?

Oh! I wasn't aware of this wrapper, thanks!

jjmachan:
@Yuri-Albuquerque no worries - glad we got it fixed 🙂 

I'm closing this now but thanks a lot for putting the effort to make a PR. Really appreciate that!

PR #1541, opened 2024-10-20
Fix: Resolve Agentic failure loop with improved json parsing

### LINKED ISSUE #1538: Expensive agentic failure loop

**Describe the bug**
In some cases / with some models, the instructions to generate json result in something like

```txt
Here's the generated abstract conceptual question in the requested JSON format::

{"text": ...}

Would you like me to explain in more detail?
```

The built-in validator will fail this because it cannot parse the json, and then make several repeated requests to try to conform the response to the specified schema.  If this is an error associated with an LLM's default response preferences, the agentic retry loop is basically a waste of tokens.

Ragas version: 0.2.1
Python version:

**Code to Reproduce**
Use Anthropic Claude 3.5 Sonnet with the new testset generator

**Error trace**
[error.txt](https://github.com/user-attachments/files/17445255/error.txt)

**Expected behavior**
Intelligent extraction of json object from text blob

**Additional context**

ahgraber:
Fixes #1538 

- Add final instruction for better json schema adherence and reduce "filler" text
- Add `extract_json` function to identify json by pairs of `[]` or `{}`
- Add unit tests for `extract_json`

**Notes:**
In my testing, these changes vastly reduce the number of times the 'repair' agent is triggered.  However, this does not fix _all_ parse errors:

1. In some cases, json is valid json, but keys are invalid
2. In some cases, especially with more complex Pydantic models (e.g., `Themes`, which is composed of `List[Theme]`), the Pydantic json schema includes `$defs`.  As a result, some LLMs will return `$defs` or include `"$schema": "https://json-schema.org/draft/...`; either of these additional keys will cause the Pydantic parser to fail.

### DIFF
--- a/src/ragas/prompt/pydantic_prompt.py
+++ b/src/ragas/prompt/pydantic_prompt.py
@@ -16,7 +16,7 @@ from ragas.exceptions import RagasOutputParserException
 from ragas.llms.prompt import PromptValue
 
 from .base import BasePrompt, StringIO, _check_if_language_is_supported
-from .utils import get_all_strings, update_strings
+from .utils import extract_json, get_all_strings, update_strings
 
 if t.TYPE_CHECKING:
     from langchain_core.callbacks import Callbacks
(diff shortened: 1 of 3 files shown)


jjmachan (approved):
thanks a lot @ahgraber for this PR ❤️ 
will be releasing it today

PR #1543, opened 2024-10-20
This commit implements the F-beta score metric

Yuri-Albuquerque:
for the AnswerCorrectness class.  The beta parameter is introduced to control the relative importance of recall and precision when calculating the score. Specifically:

 - beta > 1 places more emphasis on recall.
 - beta < 1 favors precision.
 - beta ==1 stands for the regular F1 score that can be interpreted as a harmonic mean of the precision and recall. 

Key Changes:
The method _compute_statement_presence is updated to calculate the F-beta score based on true positives (TP), false positives (FP), and false negatives (FN).

This ensures that we can balance between recall and precision, depending on the task's requirements, by tuning the beta value.

source: https://scikit-learn.org/1.5/modules/generated/sklearn.metrics.fbeta_score.html


shahules786 (changes requested):
Hey, this is useful. From ragas 0.2 onwards, we have [factual correctness](https://docs.ragas.io/en/latest/concepts/metrics/available_metrics/factual_correctness/#factual-correctness) score - can you also add this to it? https://github.com/explodinggradients/ragas/blob/791767d9676fa1a70987782b925c8f3bb04e4ce9/src/ragas/metrics/_factual_correctness.py#L262

Yuri-Albuquerque:
> Hey, this is useful. From ragas 0.2 onwards, we have [factual correctness](https://docs.ragas.io/en/latest/concepts/metrics/available_metrics/factual_correctness/#factual-correctness) score - can you also add this to it?
> 
> https://github.com/explodinggradients/ragas/blob/791767d9676fa1a70987782b925c8f3bb04e4ce9/src/ragas/metrics/_factual_correctness.py#L262

oh thank sr. I'll do it now.

shahules786:
Hey @Yuri-Albuquerque thanks for the change. 
Since F1 score is repeatedly used in multiple metrics, I think we should have it defined in utils and reuse everywhere else to avoid code duplication. Let me take care of this , and will merge this PR. Thanks for your contribution.

Yuri-Albuquerque:
> Hey @Yuri-Albuquerque thanks for the change. Since F1 score is repeatedly used in multiple metrics, I think we should have it defined in utils and reuse everywhere else to avoid code duplication. Let me take care of this , and will merge this PR. Thanks for your contribution.

Absolutely, @shahules786, feel free to take over this merge! I completely agree that this function belongs in utils. I'm still learning how to contribute effectively to this project, so I appreciate your guidance. I work at one of the largest private banks in Brazil (Itaú SA), and we're using RAGAS to evaluate various features in our chatbot. The "F1-beta" metric is something we've been wanting for a long time.
Thanks for your attention

shahules786 (approved):
Hey @Yuri-Albuquerque just made the changes from my end. Thanks a lot :)

PR #1547, opened 2024-10-21
update AspectCritic docs examples

licux:
Update example codes in AspectCritic metrics to resolve "AssertionError: set LLM before use"


shahules786 (approved):
Thanks a lot :)

PR #1556, opened 2024-10-22
chore: add verbose option (V=1) to make targets

trevorbowen:
Like the Linux kernel and a few other make based projects, adds `V=1` option to easily make `make` more verbose, revealing commands being executed, for example:

    make test V=1

Also:

- Used `$(MAKE)` instead of `make` for recursive calls.
- Added docs for updating setuptools to work on Python 3.11 on Ubuntu 24.04.


jjmachan:
wow thanks a lot @trevorbowen this was something I was not aware of - thanks for contributing this in ❤️ 🙂 🙌🏽

PR #1569, opened 2024-10-24
Fix: Limit number of retries for parse failures

rskew:
When parsing of an LLM response fails, the invalid output is sent to the LLM to be fixed.

This PR threads the number of retries through this call, preventing unbounded recursion.

The old `max_retries` wasn't preventing this due to `generate()` and `parse_output_string()` being co-recursive via the call to `generate()` here
https://github.com/explodinggradients/ragas/blob/ade46fb7c0b5dffb76ef26d876ff021ded9dfa96/src/ragas/prompt/pydantic_prompt.py#L406
The result was a prompt that would keep growing through recursive calls (with nested versions becoming increasingly more deeply quoted) until the prompt was too big for the LLM to process.

Addresses https://github.com/explodinggradients/ragas/issues/1538

### DIFF
--- a/src/ragas/prompt/pydantic_prompt.py
+++ b/src/ragas/prompt/pydantic_prompt.py
@@ -93,6 +93,7 @@ class PydanticPrompt(BasePrompt, t.Generic[InputModel, OutputModel]):
         temperature: t.Optional[float] = None,
         stop: t.Optional[t.List[str]] = None,
         callbacks: t.Optional[Callbacks] = None,
+        retries_left: int = 3,
     ) -> OutputModel:
         """
         Generate a single output using the provided language model and input data.
@@ -111,6 +112,8 @@ class PydanticPrompt(BasePrompt, t.Generic[InputModel, OutputModel]):
             A list of stop sequences to end generation.
         callbacks : Callbacks, optional
             Callback functions to be called during the generation process.
+        retries_left : int, optional
+            Number of retry attempts for an invalid LLM response
 
         Returns
         -------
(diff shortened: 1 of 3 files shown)


src/ragas/prompt/pydantic_prompt.py
```diff
@@ -93,6 +93,7 @@ async def generate(
         temperature: t.Optional[float] = None,
         stop: t.Optional[t.List[str]] = None,
         callbacks: t.Optional[Callbacks] = None,
+        retries_left: int = 3,
```
rskew, line 96: Should this come from `RunConfig`?

src/ragas/prompt/pydantic_prompt.py
```diff
@@ -410,17 +417,12 @@ async def parse_output_string(
                         prompt_value=prompt_value.to_string(),
                     ),
                     callbacks=retry_cb,
+                    retries_left = retries_left - 1,
                 )
                 retry_rm.on_chain_end({"fixed_output_string": fixed_output_string})
-                return await self.parse_output_string(
```
rskew, line 415: `generate_multiple` calls `parse_output_string()` with the result, we don't need to call it here

src/ragas/prompt/pydantic_prompt.py
jjmachan: yep that would be a good addition

src/ragas/prompt/pydantic_prompt.py
jjmachan: thanks for catching this 🙌🏽

jjmachan:
@rskew really grateful for these improvements and the very well written PR 🙂 ❤️ 

ps: we are also doing https://forms.gle/7wdMqwdPghzL2iCn6 as part of the recent v0.2 release - do fill it out if you get the chance.

Issue #1567, opened 2024-10-23
LLMContextPrecisionWithoutReference KeyError: 'reference'


mridulBanik112:
[ ] I have checked the [documentation](https://docs.ragas.io/) and related resources and couldn't resolve my bug.

**Describe the bug**

Ragas version:
Python version:

**Code to Reproduce**
```
import asyncio
from ragas.dataset_schema import SingleTurnSample
from ragas.llms import LangchainLLMWrapper
from langchain_openai import ChatOpenAI
from ragas.metrics import LLMContextPrecisionWithoutReference

import os
os.environ["OPENAI_API_KEY"] = "Key"

sample = SingleTurnSample(
    user_input="Where is the Eiffel Tower located?",
    response="The Eiffel Tower is located in Paris.",
    retrieved_contexts=["The Eiffel Tower is located in Paris."], 
)
evaluator_llm = LangchainLLMWrapper(ChatOpenAI(model="gpt-4"))
context_precision = LLMContextPrecisionWithoutReference(llm=evaluator_llm)
async def call_context():
    score = await context_precision.single_turn_ascore(sample)
    print(score)

asyncio.run(call_context())
```

**Error trace**

```
python3.10/site-packages/ragas/metrics/_context_precision.py", line 163, in _ascore
    answer=row["reference"],
KeyError: 'reference'
```

**Expected behavior**
It should give score without requiring reference.

**Additional context**

alaliaa:
Glad to see this error reported. I am facing similar issue.

jjmachan:
this was fixed with the PR - will be out next week - thanks for raising the issue 🙂

PR #1570, opened 2024-10-24
Fix: reference key error in LLMContextPrecisionWithoutReference

xizhou-vw:
fix to #1567


jjmachan:
thanks a lot @xizhou-vw for catching this and contributing the fix 🙂 ❤️

PR #1578, opened 2024-10-25
Improve efficiency in factual correctness for precision mode

Jeff-67:
### Description

This pull request optimizes the `_single_turn_ascore` function by conditionally skipping unnecessary computations when the mode is set to "precision." Previously, the function calculated `response_reference` regardless of the scoring mode, which consumed resources and potentially increased processing costs. By avoiding the `response_reference` calculation in "precision" mode, this update aims to enhance speed and reduce computational expenses.

### Changes Made

- Modified `_single_turn_ascore` function:
  - **Conditionally calculated `response_reference`**: The `response_reference` is now computed only when the mode is not "precision."


shahules786 (approved):
Nice catch, thanks for the contribution.

PR #1579, opened 2024-10-25
Updated return type description for evaluate function

taihim:
The evaluate function doc-string is still showing the old Result object as its return type. Updated it to the EvaluationResult object


jjmachan:
thanks a lot @taihim ❤️

PR #1585, opened 2024-10-26
fix: typo: ROUGE is a metric, ROGUE is a scoundrel

ahgraber:
Fixes the typo 'rogue' with referring to 'rouge' score.

Because this changes a file name and param name, it can be considered a breaking change


shahules786 (approved):
Thanks for spotting this :)

jjmachan:
since this was relatively new its okay

PR #1589, opened 2024-10-27
Feature: batched execution

ahgraber:
Provides batched execution support for `ragas.executor.Executor` and `ragas.async_utils.run_async_tasks`.

This is useful, for example, for folks who are self-hosting their own models and therefore don't have the throughput that a model service has.  In this circumstance, sending a long list of task to the Executor would previously result in timeouts as all `n` tasks were activated at the same time, but the LLM responses were queued.
The batched approach only "activates" `batch_size` of the coroutines, and then waits until that batch has been completed before moving on.

Batching may result in slower/longer execution times with LLM APIs, but is _disabled by default_ with `batch_size=None`.

--- 
Note: `tqdm` support is a bit weird, and I haven't been able to resolve after spending about a day and a half on it, so I'm giving up / forcing `leave=True`.
- my current hypothesis is that there's a bug in `tqdm`'s `position` kwarg, so the inner progress bar gets overwritten by the other loop's when `leave=False`
- as a result, it _seems_ that inner bar doesn't refresh until it is complete; it shows up due to `leave=True` once the cursor drops to the next line for the next batch

### DIFF
--- a/docs/getstarted/rag_testset_generation.md
+++ b/docs/getstarted/rag_testset_generation.md
@@ -39,7 +39,7 @@ Now we will run the test generation using the loaded documents and the LLM setup
 ```python
 from ragas.testset import TestsetGenerator
 
-generator = TestsetGenerator(llm=generator_llm)
+generator = TestsetGenerator(llm=generator_llm, embedding_model=generator_embeddings)
 dataset = generator.generate_with_langchain_docs(docs, testset_size=10)
 ```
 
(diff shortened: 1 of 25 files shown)


jjmachan:
thanks a lot for this 🔥 - this was something that I was hoping to get to at some point in time and I see that the implementation is very clean as well

do give me 2 days to test and review this and we'll get it out before the next release 🙂

ahgraber:
The most recent PR makes it cleaner, and I _think_ resolves the progress bar ordering issue.  Part of the issue seems to be that it requires ipywidgets for improved rendering in when using jupyter notebooks or when running .py files with interactive cells; I don't know if that'd be worth a call out somewhere?

jjmachan:
thanks a lot as always @ahgraber ❤️ 🙂

PR #1595, opened 2024-10-29
Added pandas as a method for reading datasets

joaorura:


jjmachan (approved):
this would be great - thanks  a lot for adding this in 🙂 ❤️

PR #1596, opened 2024-10-29
Implemented Support Another Languages - Portugue tested | Add Google Translation

joaorura:
Some small models, such as llama 3.1 or 3.2 3b or 2B. Mainly quantized, have difficulties with prompts in Portuguese, occasionally generating text in English, which is accented with part of the prompts in English.
In this way, translating all parts of the prompt is interesting.
In addition, for small models, translating can be a poorly executed task.
In this way, being able to have support from Google Translate is a quick and practical way to translate the prompts without worrying about getting a more robust model.
In addition, there was no support in English, so I added support for some more languages ​​using the NLTK segmenter implementation.

### DIFF
--- a/src/ragas/metrics/base.py
+++ b/src/ragas/metrics/base.py
@@ -2,12 +2,16 @@ from __future__ import annotations
 
 import asyncio
 import logging
+import re
+import nltk
+
 import typing as t
 from abc import ABC, abstractmethod
 from collections import Counter
 from dataclasses import dataclass, field
 from enum import Enum
-
+from pysbd.cleaner import Cleaner
+from pysbd.utils import TextSpan
 from pysbd import Segmenter
 
 from ragas.callbacks import ChainType, new_group
@@ -15,7 +19,7 @@ from ragas.dataset_schema import MultiTurnSample, SingleTurnSample
 from ragas.executor import is_event_loop_running
 from ragas.prompt import PromptMixin
 from ragas.run_config import RunConfig
-from ragas.utils import RAGAS_SUPPORTED_LANGUAGE_CODES, deprecated
+from ragas.utils import RAGAS_SUPPORTED_LANGUAGE_CODES, RAGAS_SUPPORTED_LANGUAGE_CODES_PYSBD, deprecated
 
 if t.TYPE_CHECKING:
     from langchain_core.callbacks import Callbacks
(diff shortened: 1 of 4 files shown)


joaorura:
Another problem I had with the translation was that in none of my attempts did the hash load work. I noticed that it always changed between iterations.
I realized that this is because Python has the hash calculated for strings that did not remain constant between different executions.
I fixed this and implemented it in a generic way.

jjmachan:
hey @joaorura thanks a lot for putting together this PR 🙂 

I'm not sure about this strategy though and even if we do it, we have to refactor it as optional so as to not add any new dependencies but seems like an overkill.

@shahules786 what do you think?

joaorura:
Remembering that in addition to the Google Translate translation and the issue of translating all the strings, there is also a language limitation issue that I was also able to resolve. I don't know if it is this strategy that you mentioned, but I believe it is a potential help for me and essential since I would not be able to run the project without it.
There is also a problem with the hash that was not working because the hashes never matched.

shahules786:
@joaorura Adding Google Translate is an overkill. One can use larger LLMs to do the translation if necessary. Would recommend closing this PR. @jjmachan

joaorura:
> @joaorura Adding Google Translate is an overkill. One can use larger LLMs to do the translation if necessary. Would recommend closing this PR. @jjmachan

Although it is always an option to use larger LLMs, they are not necessarily available to us. After all, to use them, you need powerful hardware available with the model data available or a key with values ​​available in some online API that needs to be configured and generally requires monetary values. So when running smaller models, it is interesting to have this resource. I honestly do not understand the context of exaggeration... The scenario of usefulness to the detriment of the context of smaller LLMs seems quite clear.

joaorura:
@jjmachan 
The new versions of ragas do not yet support Portuguese.
Which is a language spoken by many speakers.
If you could give me feedback, I could adjust and improve this code.

joaorura:
@jjmachan 
Google Translate for me we can remove. Although I still believe it is a useful tool for those who want to develop without large resources and giant LLM.

shahules786:
Hey @joaorura Thanks for replying here. Can you help us understand the limiting factor in using ragas with Portuguese? We wanted to make sure that we prioritize and solve it.

joaorura:
@shahules786 I brought up some issues mentioning them in these PR https://github.com/explodinggradients/ragas/pull/1598

I made minor corrections to these #1837 #1836 #1838

jjmachan:
hey @joaorura just merged all of those PRs and will do a release in a couple of hours. Thank you so much for the help 🙌🏽 ❤️ 

btw would love to get on a call with you sometime and chat more if your interested 🙂 (I can share my cal link if you are - or do share yours). Would love to meet you

joaorura:
> hey @joaorura just merged all of those PRs and will do a release in a couple of hours. Thank you so much for the help 🙌🏽 ❤️
> 
> btw would love to get on a call with you sometime and chat more if your interested 🙂 (I can share my cal link if you are - or do share yours). Would love to meet you

Sorry for the delay in responding @jjmachan . My English isn't the most fluent, but we can try sometime. I'll join the Ragas discord and send you a message. Thanks for your attention. I haven't had time to test the new changes yet, but when I have some time I'll take a look.

anistark:
Closing this in favor or [`adapt_prompts()`](https://docs.ragas.io/en/stable/howtos/customizations/metrics/_metrics_language_adaptation/)

PR #1605, opened 2024-10-31
Fix : Filename in the path

satodayo:
I found the filename was incorrect, and I fixed it.
Please review and merge if everything looks good.
Thank you.


shahules786 (approved):
LGTM, thanks for improving ragas. :)

PR #1606, opened 2024-10-31
docs: Add embedding_model attribute when creating TestsetGenerator object

Youngrok123:
- edit the document because embedding_model attribute has been added to the TestsetGenerator object.
- https://github.com/explodinggradients/ragas/pull/1562


jjmachan:
thanks a lot for catching that and fixing it @Youngrok123 ❤️

PR #1607, opened 2024-10-31
Fix: Add instance check before casting ChatGeneration object

Youngrok123:
My case
- When receiving responses from headline and summary extractors, etc., they are delivered as a Generation object, not ChatGeneration.
```
{
  "generations": [
    [
      {
        "text": "{\"text\": \"ABCD\"}",
        "generation_info": null,
        "type": "Generation"
      }
    ]
  ],
  "llm_output": null,
  "run": null,
  "type": "LLMResult"
}
```

Occurrence message
- unable to apply transformation: 'Generation' object has no attribute 'message'

How to fix
- Add code to first check if it is an instance of that object before casting to a ChatGeneration object.

### DIFF
--- a/src/ragas/llms/base.py
+++ b/src/ragas/llms/base.py
@@ -157,7 +157,7 @@ class LangchainLLMWrapper(BaseRagasLLM):
 
             # if generation_info is empty, we parse the response_metadata
             # this is less reliable
-            elif t.cast(ChatGeneration, resp).message is not None:
+            elif isinstance(resp, ChatGeneration) and t.cast(ChatGeneration, resp).message is not None:
                 resp_message: BaseMessage = t.cast(ChatGeneration, resp).message
                 if resp_message.response_metadata.get("finish_reason") is not None:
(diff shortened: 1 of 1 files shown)


jjmachan:
thanks again @Youngrok123 🙂 ❤️ 
btw which model are you using?

Youngrok123:
thanks, @jjmachan
I am using 'neuralmagic/Meta-Llama-3.1-70B-Instruct-FP8' model on vLLM.
And as shown in the blog below, it inherits and implements the Langchain LLM class to communicate with the LLM.
[LangChain: How to Set a Custom LLM Wrapper](https://medium.com/@antoniojimzc/langchain-how-to-set-a-custom-llm-wrapper-faf9ec270892)

If an exception occurs when generating a testset, the process will stop or normal results will not be produced, so handling the exception is necessary.

bay-mini:
same problem

jjmachan:
@bay-mini @Youngrok123 released with v0.2.4 🙂

PR #1618, opened 2024-11-04
feat: automatic persona generation

shahules786:


src/ragas/testset/persona.py
```diff
@@ -0,0 +1,150 @@
...
+class SummaryInput(BaseModel):
+    summaries: t.List[str]
+
+
+class Persona(BaseModel):
+    name: str
+    role_description: str
+
+
+class PersonasList(BaseModel):
```
jjmachan, line 42:
```suggestion
class PersonaList(BaseModel):
```
?

src/ragas/testset/persona.py
```diff
@@ -0,0 +1,150 @@
...
+
+@dataclass
+class PersonaGenerator:
+
+    llm: BaseRagasLLM
+    num_personas: int = 5
+    prompt: PydanticPrompt = PersonaGenerationPrompt()
+    filter_nodes: t.Callable[[Node], bool] = field(
+        default_factory=lambda: default_filter
+    )
```
jjmachan, line 98: add this as part of `PersonaList`?

so the interaction will become `PersonList.generate_from_kg(num_persona, kg, llm, prompt, filter)`
?

src/ragas/testset/persona.py
```diff
@@ -0,0 +1,150 @@
...
+            from sklearn.cluster import KMeans
+            from sklearn.metrics import pairwise_distances
+        except ImportError:
+            raise ImportError(
+                "PersonaGenerator requires the 'scikit-learn' package to be installed. "
+                "You can install it with 'pip install scikit-learn'."
+            )
+
+        self.pairwise_distances = pairwise_distances
+        self.kmeans = KMeans(n_clusters=self.num_personas, random_state=42)
```
jjmachan, line 112: is it possible to generate this for ourselfs - this is a fairly popular interface so easily generateable

shahules786: Do you mean, is it possible to generate/manually write persona? Yes. It just needs to be in Persona Format

src/ragas/testset/__init__.py
```diff
@@ -1,4 +1,11 @@
...
 from ragas.testset.synthesizers.generate import TestsetGenerator
 from ragas.testset.synthesizers.testset_schema import Testset, TestsetSample
 
-__all__ = ["TestsetGenerator", "Testset", "TestsetSample"]
+__all__ = [
+    "TestsetGenerator",
+    "Testset",
+    "TestsetSample",
+    "Persona",
+    "PersonaList",
```
jeswincyriac, line 10: no need to add it here

src/ragas/testset/persona.py
jeswincyriac: no, is it important to add the sklearn dependency here? we can generate the code for the functionality here

src/ragas/testset/__init__.py
jeswincyriac: the reason being that users should manually import all of the dependencies from the corresponding modules - but it is also harder when customizing to be honest.

src/ragas/testset/persona.py
shahules786: scikit-learn may be an overkill, I had the same doubt. Now I have reduced it to a much simpler algorithm w/o adding any extra dependency.

jjmachan (approved):

jjmachan:
made 2 changes to this
```py
from ragas.testset.persona import generate_personas_from_kg
persona_list = generate_personas_from_kg(kg=kg, llm=generator_llm)
```

generation being a class method, I was wrong about that one, and added executor to make things parallel at the function level.

PR #1619, opened 2024-11-04
feat: Add support for llamaIndex in evaluation

suekou:
Added type checks for llamaIndex LLMs and embeddings in the evaluate function.


pyproject.toml
```diff
@@ -8,6 +8,7 @@ dependencies = [
     "langchain-core",
     "langchain-community",
     "langchain_openai",
+    "llama_index",
```
jjmachan, line 11: llamaindex is actually not part of the core dependency. It's best if we keep it optional

jjmachan (commented):
thanks a lot for sending a PR 🙂 - just once change though and we'll merge it in

pyproject.toml
jjmachan: if you move it to `all` that would be helpful, or maybe even a new group for llamaindex?

suekou: Thanks for the feedback! I've updated the PR to make llama_index optional by moving it to the [all] group.

jjmachan (approved):
Thanks a lot @suekou 🙂 ❤️ 
merging this in

jjmachan:
btw would love to get your feedback on this docs too if you get the chance https://docs.ragas.io/en/latest/howtos/integrations/_llamaindex/

suekou:
@jjmachan I found this to be very clear and understandable! 
I noticed the URL on the page was outdated, so I submitted a PR to update it!

jjmachan:
hey @suekou thats a good point but right now `evaluate` in ragas.integrations.llamaindex has the same features as the top level evaluate right?

the idea was to keep both of the feature compatible since behind the scene all we are doing is running the query_engine for you

PR #1622, opened 2024-11-05
Introduce epsilon in denominators to avoid division by zero

jltham:
This only affects `src/ragas/_topic_adherece.py`.

Previously, it can be complicated to handle the score given by this metric due to possibility of division by zero. I propose to introduce an epsilon in the denominator to avoid that case.


shahules786 (approved):
thanks buddy :)

PR #1627, opened 2024-11-05
fix: remove `.` as a split condition in noise_sensitivity.py

### LINKED ISSUE #1626: In NoiseSensitivity, the `.` should be deleted as the split condition.
[ ] I checked the [documentation](https://docs.ragas.io/) and related resources and couldn't find an answer to my question.

**Your Question**
After the sentence is split, i think the `.` should no longer be used as a split condition. Many other languages ​​do not use `.` as a split condition. Just follow the result of `sentence_segmenter`.

**Code Examples**
https://github.com/explodinggradients/ragas/blob/main/src/ragas/metrics/_noise_sensitivity.py#L104C13-L104C46

**Additional context**
I will provide a pr.

jasinliu:
Fix #1626 to support more languages.

### DIFF
--- a/src/ragas/metrics/_noise_sensitivity.py
+++ b/src/ragas/metrics/_noise_sensitivity.py
@@ -101,7 +101,6 @@ class NoiseSensitivity(MetricWithLLM, SingleTurnMetric):
         sentences_with_index = {
             i: sentence
             for i, sentence in enumerate(sentences)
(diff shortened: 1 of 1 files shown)


shahules786 (approved):

shahules786 (approved):
Thank you:)

PR #1641, opened 2024-11-07
fixed wrong variable names in starter guide using azure openai

MichaelPerger:
found this small error while trying to use ragas following the starter guide using azure openai
previously set azure_config should be referenced instead of azure_configs (which doesn't exist)


jjmachan:
thanks a lot for the fix that @MichaelPerger 🙌🏽

PR #1643, opened 2024-11-07
docs: fix _arize.md

suekou:


jjmachan (commented):
thanks a lot @suekou 🙌🏽 
could you add it to https://github.com/explodinggradients/ragas/blob/0415a2dd313f9991f9d41f8d78f55346a4a09235/mkdocs.yml#L90-L92 too so that it is part of the docs

suekou:
@jjmachan Added it!

Issue #718, opened 2024-03-06
Testset generation improvements


shahules786:
- [x] Experiment on topic extraction
- [x] use feedback to rewrite questions

PR #1645, opened 2024-11-08
feat: improvements in test gen

shahules786:
1. Add node filtering mechanism
2. Add min and max tokens in headline splitting


shahules786:
Items here are also mentioned in #718

jjmachan (approved):

PR #1650, opened 2024-11-09
Bug fixing recall in  _factual_correctness.py

GitMarco27:
Bug fixing recall in factual correctness:

recall is equal to: tp / (tp + fn)

but was calculated exactly like the precision.

Reference: https://docs.ragas.io/en/latest/concepts/metrics/available_metrics/factual_correctness/#factual-correctness


jjmachan:
thanks alot for that catch @GitMarco27 - will be out in todays release ❤️

PR #1656, opened 2024-11-12
Update _factual_correctness.py

yousoferfani:
The recall metric  formula should change to tp/(tp+fn)


jjmachan:
hey @yousoferfani thanks a lot for catching this #1650 this also fixes the same thing so closing this in favor of that 

the fix will be released now

yousoferfani:
Sure
Thanks

jjmachan:
but we really appreciate time time you took to fix it, thanks a million 🙌🏽

Issue #1658, opened 2024-11-12
Issue about tool call metrics.


MattZ-99:
Hi all, it's surprising to see the ToolUse metrics in Ragas version v0.2.

Based on the base `Metric` class, I have implemented several useful metrics before.
After the update of ToolUse section, I'm glad to contribute them for the open source.

As a new here, I'd like to have a discussion before pull request. 

1. Metrics for parallel calling.

Parallel function calling now is a common feature for current llm, where `parallel` means multiple independent/unordered function calls. For example, [Berkeley Function-Calling Leaderboard](https://gorilla.cs.berkeley.edu/leaderboard.html) provides a specialized "parallel" category.
 

The below case is the parallel tool call for two information. 

```python
sample = [
    HumanMessage(content="What's the match schedule and weather in LA this weekend?"),
    AIMessage(content="...", tool_calls=[
        ToolCall(name="weather_check", args={"location": "Los Angeles"}),
        ToolCall(name="match_check", args={"location": "Los Angeles"})
    ]),
]

sample = MultiTurnSample(
    user_input=sample,
    reference_tool_calls=[
        ToolCall(name="match_check", args={"location": "Los Angeles"}),
        ToolCall(name="weather_check", args={"location": "Los Angeles"})
    ]
)
```

Both ordered tool_calls should be acceptable, while current ToolCallAccuracy only supports ordered tool_calls. 
That is, `ToolCallAccuracy` is not suitable for parallel calling.

```python
scorer = ToolCallAccuracy()
result = asyncio.run(scorer.multi_turn_ascore(sample))
# 0.0
```

As for the **solution**, I have two ideas:
1. Add another metric, 'ToolCallParallelAccuracy', which focus on parallel tool calls.
2. Add a new parameter to ToolCallAccuracy, 'ordered_tool_calls=True/False', indicating whether the tool calls should be ordered.

which one would you think is more flexible?

Besides, I also have a small concern about current `ToolCallAccuracy`.
```python
# ref from _tool_call_accuracy:76~
sequence_aligned = int(
    self.is_sequence_aligned(tool_call_pred_sequence, tool_call_ref_sequence)
)

if pred_tool_calls:
    score = 0.0
    reference_tool_calls = sample.reference_tool_calls
    for ref_tool_call in reference_tool_calls:
        for pred_tool_call in pred_tool_calls:
            ...
```

[...]

PR #1665, opened 2024-11-13
Update tool call accuracy.

MattZ-99:
ref: #1658.

Two updates:
- bug fix of #1658 
- metric of `ToolCallParallelAccuracy`

### DIFF
--- a/src/ragas/metrics/__init__.py
+++ b/src/ragas/metrics/__init__.py
@@ -62,7 +62,7 @@ from ragas.metrics._string import (
     StringPresence,
 )
 from ragas.metrics._summarization import SummarizationScore, summarization_score
-from ragas.metrics._tool_call_accuracy import ToolCallAccuracy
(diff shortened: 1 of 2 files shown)


MattZ-99:
Yeah. Really good point, as current `ToolCallParallelAccuracy` is highly inherited from the original `ToolCallAccuracy`.

shahules786:
Hey @MattZ-99 An update on our side, as you agreed we want to add both in one interface. ie to to toolcallaccuracy itself. But we need to find the best way for user's specify which calls should be parallel and in sequence. We will work together and make it happen, if you have any ideas please feel to share it here.

MattZ-99:
Hi @shahules786 , I have thought about it. A straightforward idea is add a parameter like `ordered_tool_call` which controls 'parallel' or 'sequencial' function call. What do u think about it?  Or could you share your code?

anistark:
@MattZ-99 Could you pleae rebase with main and check for formatting issues? 
Also, please run : `make test`
Some test seem to be failing.

anistark:
Implemented with #2307

PR #1666, opened 2024-11-13
feat: make general purpose metrics more general

jjmachan:
## Metrics Converted

- [x] Aspect Critic
- [x] Simple Criteria
- [x] Rubric Based - both Instance and Domain specific

a few different examples

### Aspect Critic
```py
from ragas.metrics import AspectCritic

only_response = SingleTurnSample(
    response="The Eiffel Tower is located in Paris."
)

grammar_critic = AspectCritic(
    name="grammar",
    definition="Is the response grammatically correct?",
    llm=evaluator_llm
)

await grammar_critic.single_turn_ascore(only_response)
```

with reference
```py
answer_correctness_critic = AspectCritic(
    name="answer_correctness",
    definition="Is the response and reference answer are the same?",
    llm=evaluator_llm
)

# data row
sample = SingleTurnSample(
    user_input="Where is the Eiffel Tower located?",
    response="The Eiffel Tower is located in Paris.",
    reference="London"
)
await answer_correctness_critic.single_turn_ascore(sample)
```

**Note:** this only works for multi-turn metrics for now


shahules786:
Let's say I am evaluating using an evaluation dataset that contains user_input, response, reference, and retrieved_context. Consider scenarios of the user using two metrics with this dataset, ie context_recall and aspect critic (as harmfulness or something)  at the same time. Using this interface, even if my aspect critic metric does not need retrieved context it will use it. There is no way to opt-out.  @jjmachan

shahules786 (changes requested):
Reflect changes in docs.

docs/concepts/metrics/available_metrics/general_purpose.md
jjmachan, line 1: updated the docs here

src/ragas/metrics/_aspect_critic.py
```diff
@@ -29,8 +29,21 @@ class AspectCriticOutput(BaseModel):
...
+    retrieved_contexts: t.Optional[t.List[str]] = Field(
+        description="The retrieved contexts from the llm system", default=None
+    )
+    reference_contexts: t.Optional[t.List[str]] = Field(
+        description="The reference contexts for the evaluation", default=None
+    )
+    reference: t.Optional[str] = Field(
+        description="The reference answer for evaluation", default=None
+    )
     criteria: str = Field(description="The criteria to evaluate the response")
```
shahules786, line 47: We discussed that both here and with simple criteria this should be removed. If not with this PR, I can make another PR after this is merged.

shahules786 (approved):

PR #1668, opened 2024-11-13
Added initialize calls for generator methods

Maneshgpai:
Added import and sample code for initialising generator methods. In example used for testset generation, using Langchain methods.


Maneshgpai:
I see that this line of code is not in line with Git repo. So removing my PR

    generator = TestsetGenerator(llm=generator_llm, embedding_model=generator_embeddings)
TypeError: TestsetGenerator.__init__() got an unexpected keyword argument 'llm'

PR #1669, opened 2024-11-13
Avoid unnecessary decompose_claims call when mode='precision' in _factual_correctness.py

dfhssilva:
Avoid unnecessary decompose_claims call when mode='precision' in FactualCorrectness.


shahules786 (approved):
Nice catch @dfhssilva

PR #1687, opened 2024-11-17
Added parallel tool call evaluation to ToolCallAccuracy metric

### LINKED ISSUE #1658: Issue about tool call metrics.

sahusiddharth:
- Fixes #1658 

- Introduced the `mode` parameter in the `ToolCallAccuracy` class to allow 
  selection between `sequential` and `parallel` evaluation modes.

### DIFF
--- a/docs/concepts/metrics/available_metrics/agents.md
+++ b/docs/concepts/metrics/available_metrics/agents.md
@@ -70,6 +70,8 @@ scorer = TopicAdherenceScore(mode="recall")
 
 `ToolCallAccuracy` is a metric that can be used to evaluate the performance of the LLM in identifying and calling the required tools to complete a given task. This metric needs `user_input` and `reference_tool_calls` to evaluate the performance of the LLM in identifying and calling the required tools to complete a given task. The metric is computed by comparing the `reference_tool_calls` with the Tool calls made by the AI. The values range between 0 and 1, with higher values indicating better performance. 
 
+By using the `mode` parameter, the evaluation can be customized to meet the specific requirements of your task. In `sequential` mode, the metric ensures that the agent adheres to a strict order of function calling, matching the sequence of the `reference_tool_calls`. In contrast, parallel mode evaluates the presence and correctness of tool calls without considering their order, allowing greater flexibility in scenarios where sequence is not critical.
+
 ```python
 from ragas.metrics import ToolCallAccuracy
 from ragas.dataset_schema import  MultiTurnSample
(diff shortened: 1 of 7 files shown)


shahules786:
I think this is a better solution for now, compared to introducing new metric just for parallel like https://github.com/explodinggradients/ragas/pull/1665. But ideally what we need is an interface where use could both in single run. Ie there can be tool calls in single agentic run that should be made in parelle and ones that could be made in sequence.

PR #1701, opened 2024-11-25
feat: train your own evaluators

shahules786:
- [x] Config
- [x] Loss
- [x] Optimizer base


src/ragas/config.py
jjmachan, line 1: should we put this fine somewhere else? like under `optimze` module?

src/ragas/losses.py
jjmachan, line 1: same for this one?

jjmachan (commented):
not sure about those 2 files which are only relevent when optimizing - so they should be inside optimzers right (the idea being it is better to have it inside a module that outside in the root)

Issue #1150, opened 2024-08-01
Getting error "Failed to parse output. Returning None" on faithfulness metric


ableiweiss:
I am getting error "Failed to parse output. Returning None" on faithfulness metric for some inputs. This is inconsistent behavior as it is haphazard and sometimes works, sometimes doesn't for the same input. I don't see this error on any of the other metrics. I get the error regardless of the LLM I use (Mixtral, LLama, etc.)

Any solutions to this issue?

ableiweiss:
The issue is that `faithfulness_list` is empty, so it returns` np.nan`. The list `nli_result_text` looks fine, are there any unique requirements for the prompt that might cause this issue?

ableiweiss:
I see that the last statement in `nli_result_text` get cut off and does not contain a `reason `or `verdict `field. The rest of the statements look good. Is there a way to just ignore the erroneous statements and still use the rest for the metric?

jjmachan:
@ableiweiss what is the context window of the models you are using? and how big are the models you tested?

ableiweiss:
context window is 8k, tried LLama3-70b-instruct and Mixtral-8x-7b. Didn't have issues with the other metrics.

Is there a particular open source LLM you recommend to alleviate this? LLama3.1 perhaps?

jjmachan:
@ableiweiss maybe try Llama3.1 which has 128K window? 

the first thing is to figure out if it is a context over fitting problem. If you are using a tracing tool it should be easy (I'm guessing) or you can use this to help you out https://docs.ragas.io/en/stable/howtos/applications/cost.html

shahfasal:
I have encountered a similar issue(using mixtral model) where I get the error `Failed to parse output.` Returning None on the `faithfulness metric` for some inputs. This error occurs inconsistently – sometimes it works, sometimes it doesn’t, even with the same input. Additionally, this issue appears regardless of the LLM used (Mixtral, Llama, etc.).

**My findings and solution:**
After troubleshooting, I identified that the root cause lies in `missing fields` in the JSON output from the model, specifically the `verdict field.` When the model is unable to infer a judgment clearly, it might return incomplete output, causing the parser to fail.

To fix this, I found that modifying the prompt to explicitly ensure that both the `reason and verdict fields` are included in the model's output solves the problem. Here’s the change I implemented:

> Updated Prompt in `_faithfullness.py` goto `NLI_STATEMENTS_MESSAGE`:

[...]

PR #1707, opened 2024-11-26
Fix for #1150:  Update prompt to ensure reason and verdict fields are present for custom model evaluation

shahfasal:
#### Summary
* This pull request addresses [Issue #1150](https://github.com/explodinggradients/ragas/issues/1150) by updating the faithfulness instruction prompt to ensure that the reason and verdict fields are always included when evaluating with custom models.

#### Changes
* Updated the evaluation prompt to explicitly guarantee the presence of reason and verdict fields.
* Improved prompt clarity to align with expected evaluation outcomes.
#### Context
* It was observed by many users that the reason and verdict fields often gets cut off, resulting in "failed to parse error". This fix ensures that all statements, contain these necessary fields.
* For more details on the issue, refer to the discussion in [Issue #1150](https://github.com/explodinggradients/ragas/issues/1150).

#### Testing
* Verified for Mixtral8x7b and Mistral 7b hosted internally using vllm 
* Ensured no missing fields in the evaluation results.

### DIFF
--- a/src/ragas/metrics/_faithfulness.py
+++ b/src/ragas/metrics/_faithfulness.py
@@ -101,7 +101,7 @@ class NLIStatementInput(BaseModel):
 
 
 class NLIStatementPrompt(PydanticPrompt[NLIStatementInput, NLIStatementOutput]):
-    instruction = "Your task is to judge the faithfulness of a series of statements based on a given context. For each statement you must return verdict as 1 if the statement can be directly inferred based on the context or 0 if the statement can not be directly inferred based on the context."
+    instruction = "Your task is to judge the faithfulness of a series of statements based on a given context. For each statement you must return verdict as 1 if the statement can be directly inferred based on the context or 0 if the statement can not be directly inferred based on the context. Ensure that each statement includes both a reason and a verdict. Do not omit any fields."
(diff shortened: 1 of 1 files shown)


jjmachan:
hey @shahfasal we won't be able to merge this in since it is very specific but you could use https://docs.ragas.io/en/stable/howtos/customizations/metrics/_modifying-prompts-metrics/ to change the prompt internally which should solve the problem

can I close this issue in that case?

shahfasal:
@jjmachan Thanks for the suggestion! I will look into this approach for modifying the prompt externally. However, won’t every user, especially those using custom models like Mixtral and Mistral, need to manually update the prompt in this way every time they encounter this issue? It would be great to know if there's a more streamlined solution for this, or if this is something that should be done on a case-by-case basis.

shahules786:
@shahfasal Yes, depending upon the model the user will have to change the prompt (we are bringing in something to automate it soon 🔜).  The reason why this change won't benefit the greater community is that when it works your model, there will be always some other set of models for which this addition is unnecessary and for which this instruction does not work at all.

shahfasal:
Fine lets close this PR then

PR #1714, opened 2024-11-27
Bug in documentation rag_testset_generation.md: API changes required arguments

bmerkle:
The code in rag_testset_generation.md code not compile and contains 2 syntax errors (API has changed)
I now added missing possitional arguments for embedding_model and documents

### DIFF
--- a/docs/getstarted/rag_testset_generation.md
+++ b/docs/getstarted/rag_testset_generation.md
@@ -107,7 +107,7 @@ from ragas.testset.transforms import default_transforms, apply_transforms
 transformer_llm = generator_llm
 embedding_model = generator_embeddings
 
-trans = default_transforms(llm=transformer_llm, embedding_model=embedding_model)
+trans = default_transforms(documents=docs, llm=transformer_llm, embedding_model=embedding_model)
 apply_transforms(kg, trans)
 ```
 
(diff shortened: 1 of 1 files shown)


bmerkle:
BTW. how about adding a samples directory with ready to run jupyter notebooks ? 
I could contribute the two samples from your documentation as notebooks

shahules786 (approved):
Thanks for the PR :)

shahules786:
Hey @bmerkle that's a great idea. We used to have our docs in jupyter notebook but moved it to md for proper rendering. We could have a notebook version of the same that can be opened in colab or something. What do you think @jjmachan ?

bmerkle:
> Hey @bmerkle that's a great idea. We used to have our docs in jupyter notebook but moved it to md for proper rendering. We could have a notebook version of the same that can be opened in colab or something. What do you think @jjmachan ?

hey @shahules786 @jjmachan, we could have an example directory with ready to run jupyter notebooks. 

In GitHub Workspaces and Google colab they can be directly used and and use can play around with it.
The build process could generate md files for the website automatically from the jupyter notebooks, e.g.
`jupyter nbconvert --to markdown your_notebook.ipynb`
Or we could use something like binder or viola to have it even live in the ragas documentation :-) 

I could prepare a PR for this if you like. I think we should create a separate issue for it out of this issue.

jjmachan:
hey @bmerkle it a good idea, which notebook topics do you have in mind

but if you do check out the current documentation we use something similar https://github.com/explodinggradients/ragas/tree/main/docs/howtos/customizations/testgenerator

this is the script https://github.com/explodinggradients/ragas/blob/main/docs/ipynb_to_md.py

PR #1724, opened 2024-12-03
feat: genetic algo based optimizer

shahules786:
```python
metric =AspectCritic(name="answer_correctness",definition="Given the user_input, reference and response. Is the response correct compared with the reference",llm=llm_4o)
metric.train("alignment_sample.json")
```

[dummy data](https://github.com/user-attachments/files/17997460/alignment_sample.json)


shahules786:
The progress bar for this now is not neat. I would like to have it as one single bar with a subprocess shown with progress. I'm not sure what the best way to do it is.

jjmachan:
@shahules786 check out this https://github.com/explodinggradients/ragas/blob/99d1a47b9025ea46625d3f4985309eff24025c63/src/ragas/executor.py#L146-L181

which does something similar for batching

shahules786:
@jjmachan thanks, just fixed it. Please review

src/ragas/optimizers/genetic.py
```diff
@@ -0,0 +1,733 @@
...
+                self._feedback_mutation,
+                candidate=candidate,
+                dataset=dataset_sample,
+                results=results,
+                target=target,
+                callbacks=candidate_grp,
+            )
+            try:
+                improved_candidate = exec.results()
+                improved_candidates.append(improved_candidate[0])
```
jjmachan, line 420: shouldn't we submit all the jobs and then wait for the results outside the loop?

jjmachan (commented):
overall this works really well but I'm not sure about if we are using the executor correctly - will love to discuss this over a call before merging this in

also comments - the code is very hard to read since there are some new alogs here - proper variable names and comments will go a long way to improve the readability of the code

src/ragas/optimizers/genetic.py
```diff
@@ -0,0 +1,733 @@
...
+            desc="Mutating candidates",
+            raise_exceptions=raise_exceptions,
+            run_config=run_config,
+            keep_progress_bar=False,
+            batch_size=batch_size,
+            pbar=pbar,
+        )
+
+        offspring_candidates = []
+        for idx, candidate in enumerate(candidates):
```
shahules786, line 713: make this into one batch

src/ragas/optimizers/genetic.py
```diff
@@ -0,0 +1,733 @@
...
+
+        if self.metric is None:
+            raise ValueError("No metric provided for optimization.")
+
+        prediction = results.to_pandas()[self.metric.name].values.tolist()
+        indices = [idx for idx in range(len(target)) if target[idx] != prediction[idx]]
+        traces = [trace[self.metric.name] for trace in results.traces]
+        if indices:
+            feedback_candidates = {}
+            for key in candidate.keys():
```
shahules786, line 507: rename key with what it represents

jjmachan (approved):

PR #1739, opened 2024-12-07
feat: few shot example optimzier

jjmachan:
optimize with few short examples

```py
from ragas.metrics import AspectCritic

# define metric
llm = llm_factory("gpt-4o")
metric = AspectCritic(
    name="answer_correctness",
    definition="Given the user_input, reference and response. Is the response correct compared with the reference",
    llm=llm,
)

# optimize with annotation
from ragas.config import DemonstrationConfig
demonstration_config = DemonstrationConfig()
metric.train(
    "alignment_sample.json",
    demonstration_config=demonstration_config,
)
```

### DIFF
--- a/src/ragas/config.py
+++ b/src/ragas/config.py
@@ -1,9 +1,11 @@
+from __future__ import annotations
+
 import typing as t
 
 from pydantic import BaseModel, Field
 
 from ragas.embeddings import BaseRagasEmbeddings, embedding_factory
-from ragas.llms import BaseRagasLLM
+from ragas.llms import BaseRagasLLM, llm_factory
 from ragas.losses import Loss
 from ragas.optimizers import GeneticOptimizer, Optimizer
 
@@ -11,17 +13,20 @@ DEFAULT_OPTIMIZER_CONFIG = {"max_steps": 100}
 
 
 class DemonstrationConfig(BaseModel):
+    embedding: BaseRagasEmbeddings = Field(default_factory=embedding_factory)
     enabled: bool = True
     top_k: int = 3
     technique: t.Literal["random", "similarity"] = "similarity"
-    embedding: BaseRagasEmbeddings = Field(default_factory=lambda: embedding_factory())
 
 
 class InstructionConfig(BaseModel):
+    llm: BaseRagasLLM = Field(default_factory=llm_factory)
     enabled: bool = True
     loss: t.Optional[Loss] = None
     optimizer: Optimizer = GeneticOptimizer()
     optimizer_config: t.Dict[str, t.Any] = Field(
         default_factory=lambda: DEFAULT_OPTIMIZER_CONFIG
     )
-    llm: t.Optional[BaseRagasLLM] = None
+
+
+InstructionConfig.model_rebuild()
--- a/src/ragas/config.py
+++ b/src/ragas/config.py
@@ -2,10 +2,10 @@ from __future__ import annotations
 
 import typing as t
 
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, field_validator
 
-from ragas.embeddings import BaseRagasEmbeddings, embedding_factory
-from ragas.llms import BaseRagasLLM, llm_factory
+from ragas.embeddings.base import BaseRagasEmbeddings
+from ragas.llms.base import BaseRagasLLM
 from ragas.losses import Loss
 from ragas.optimizers import GeneticOptimizer, Optimizer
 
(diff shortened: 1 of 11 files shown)


src/ragas/metrics/base.py
```diff
@@ -299,11 +283,100 @@ def train(
...
+                with_debugging_logs=with_debugging_logs,
+                raise_exceptions=raise_exceptions,
+            )
+
+        # if demonstration_config is provided, optimize the demonstrations
+        if demonstration_config is not None:
+            self._optimize_demonstration(
+                demonstration_config=demonstration_config,
+                dataset=dataset,
+            )
```
jjmachan, line 379: changed the logic of how to ennable which optimisation

src/ragas/prompt/few_shot_pydantic_prompt.py
```diff
@@ -0,0 +1,135 @@
...
+        default_factory=list, repr=False
+    )
+    _embeddings_of_examples: t.List[t.List[float]] = field(
+        default_factory=list, repr=False
+    )
+
+    def add_example(self, input: BaseModel, output: BaseModel):
+        # get json string for input
+        input_json = input.model_dump_json()
+        self._embeddings_of_examples.append(self.embeddings.embed_query(input_json))
```
shahules786, line 50:
```suggestion
        input_json = input.model_dump_json(exclude_none=True)
```

shahules786, line 50: In my experiments embedding user_input only and whole input (containing user_input, response,etc) gave almost identical results. But I think logically it makes more sense to only embed user_input. Although this has to be revisited again later. Let us leave it like this for now.

src/ragas/prompt/few_shot_pydantic_prompt.py
```diff
@@ -0,0 +1,135 @@
...
+        self._examples_list.append((input, output))
+
+    def get_examples(
+        self, data: BaseModel, top_k: int = 5
+    ) -> t.Sequence[t.Tuple[BaseModel, BaseModel]]:
+        data_embedding = self.embeddings.embed_query(data.model_dump_json())
+        return [
```
shahules786, line 57:
```suggestion
        data_embedding = self.embeddings.embed_query(data.model_dump_json(exclude_none=True))
```

src/ragas/prompt/few_shot_pydantic_prompt.py
```diff
@@ -0,0 +1,135 @@
...
+            for i in self.get_nearest_examples(
+                data_embedding, self._embeddings_of_examples, top_k
+            )
+        ]
+
+    @staticmethod
+    def get_nearest_examples(
+        query_embedding: t.List[float],
+        embeddings: t.List[t.List[float]],
+        top_k: int = 5,
```
shahules786, line 68: 5 examples might be too much considering avg tokens per example. Let's do 3

src/ragas/config.py
```diff
@@ -1,27 +1,32 @@
...
 class DemonstrationConfig(BaseModel):
+    embedding: BaseRagasEmbeddings = Field(default_factory=embedding_factory)
     enabled: bool = True
     top_k: int = 3
     technique: t.Literal["random", "similarity"] = "similarity"
-    embedding: t.Optional[BaseRagasEmbeddings] = None
 
 
 class InstructionConfig(BaseModel):
+    llm: BaseRagasLLM = Field(default_factory=llm_factory)
```
shahules786, line 23: My though process for making it None was
- If no model is explicitly specified, the optimization should proceed by taking the model instantiated in metric. This will allow seamless optimization with people using custom models other than open_ai.

src/ragas/config.py
```diff
@@ -1,27 +1,32 @@
...
+from ragas.embeddings import BaseRagasEmbeddings, embedding_factory
+from ragas.llms import BaseRagasLLM, llm_factory
 from ragas.losses import Loss
 from ragas.optimizers import GeneticOptimizer, Optimizer
 
 DEFAULT_OPTIMIZER_CONFIG = {"max_steps": 100}
 
 
 class DemonstrationConfig(BaseModel):
+    embedding: BaseRagasEmbeddings = Field(default_factory=embedding_factory)
```
shahules786, line 16: Same idea here as model

shahules786 (changes requested):
Overall looks good. Logic and everything is perfect. Suggested few changes.

src/ragas/prompt/few_shot_pydantic_prompt.py
jjmachan: but what if there is no user_input?

src/ragas/config.py
jjmachan: ohh yes that makes much more sense

src/ragas/config.py
jjmachan: got it

src/ragas/prompt/few_shot_pydantic_prompt.py
shahules786: It's something we have to figure out, ie what's the most accurate way to embed few shot examples. Let's leave it like this now.

src/ragas/config.py
jjmachan: but can't have that here because `train()` is in `MetricWithLLM` so don't have access to embedding

I think we should make LLM explicit, force the developer to choose it

shahules786 (approved):

PR #1741, opened 2024-12-08
feat: added basics for upload

jjmachan:


src/ragas/utils.py
```diff
@@ -22,6 +25,9 @@
 }
 # endpoint for uploading results
 RAGAS_API_URL = "https://api.ragas.io"
+RAGAS_APP_URL = "https://app.ragas.io"
+RAGAS_API_VERSION = "1.0.0"
```
ganeshrvel, line 29: RAGAS_API_VERSION = "1.0.0"

app api version should be picked up from the internals. Not via env

src/ragas/utils.py
```diff
@@ -22,6 +25,9 @@
...
+RAGAS_API_SOURCE = "ragas_py"
```
ganeshrvel, line 30: RAGAS_API_SOURCE should be hardcoded. Not picked up via env

ganeshrvel (changes requested):

src/ragas/utils.py
jjmachan: it is hard coded - its a constant

shahules786 (approved):
LGTM

PR #1759, opened 2024-12-13
fix: temperature needs to be added handled effectively

jjmachan:


aabanic:
Why is the temperature always being overriden? In case of using ResponseRelevancy this results in (with a `n` of 3) my questions being the same most of the time. Should it be passed via _ascore in ResponseRelevancy.

PR #1762, opened 2024-12-13
Fix link to langchain and add pip install 

pamelafox:
This PR fixes two docs issues that I ran into while going through this how-to guide.


shahules786 (approved):
Thanks a lot

PR #1773, opened 2024-12-19
fix: Update error handling in Faithfulness metric to return 0.0 instead of np.nan on failure

### LINKED ISSUE #1770: KeyError: 0 by following the Get Started "Run ragas metrics for evaluating RAG"
[*] I have checked the [documentation](https://docs.ragas.io/) and related resources and couldn't resolve my bug.

**Describe the bug**
The get start code about RAG evaluation has runtime exceptions. KeyError: 0. It is because the faithfulness is nan for some evaluation cases. If I remove the faithfulness metrics. It runs smoothly.

Ragas version: 0.28.0
Python version: python 3.9

**Code to Reproduce**
Share code to reproduce the issue
The code is in the https://docs.ragas.io/en/stable/getstarted/rag_evaluation/

**Error trace**

outputs={'faithfulness': nan},
KeyError: 0

**Expected behavior**
Output the top evaluations results 

**Additional context**

iamarunbrahma:
This PR fixes a KeyError: 0 issue that occurs when running RAG evaluation metrics, specifically in the faithfulness metric computation.

Fixes #1770

### DIFF
--- a/src/ragas/metrics/_faithfulness.py
+++ b/src/ragas/metrics/_faithfulness.py
@@ -262,7 +262,8 @@ class Faithfulness(MetricWithLLM, SingleTurnMetric):
 
         statements_simplified = await self._create_statements(row, callbacks)
         if statements_simplified is None:
-            return np.nan
+            logger.warning("Failed to create statements from the answer.")
+            return 0.0  # Return 0.0 instead of np.nan
 
         # unwrap the statements
         statements = []


shahules786:
Hi, we have made it return nan by design. Zero here means no hallucinations and NaN means undetermined. Here the case the latter and not the former. While processing the results, the user may fill the NaNs in the results according to his/her best judgement. Please close the PR if this explanation makes sense to you. @iamarunbrahma

iamarunbrahma:
Thanks for the explanation, @shahules786. That makes sense. Closing this PR!

PR #1781, opened 2024-12-20
A reminder for the frequent error regarding openai_key in the evaluate

WKRZY:
During the invocation of the evaluate function, I observed a common issue: when users do not pass in their custom LLM (Large Language Model), the function automatically attempts to call OpenAI's model. However, if the user has not set the OPENAI_API_KEY environment variable at this point, it will cause the function to fail and throw an error. To address this problem and enhance user experience, I have introduced a new feature to remind users to set their API key.


jjmachan:
hey @WKRZY 
Firstly thank you so much for taking the time to improve the Ragas for everyone. we really appreciate that ❤️ 

but unfortunatly this is not the solution that solves the root cause which is useing openAI as default and the users not knowing this. This was a result of a design decision we had made and will be fixing it in https://github.com/explodinggradients/ragas/pull/1558 but we have not merged it since it's breaking and waiting for v0.3

let me know if you have any questions, if not can I close this?

thanks again for your effort 🙂

PR #1786, opened 2024-12-23
docs: new get started guide

shahules786:


docs/getstarted/evals.md
```diff
@@ -0,0 +1,221 @@
...
+
+## Evaluation
+
+In this guide, you will evaluate a **text summarization pipeline**. The goal is to ensure that the output summary accurately captures all the key details specified in the text, such as growth figures, market insights, and other essential information.
+
+`ragas` offers a variety of methods for analyzing the performance of LLM applications, referred to as [metrics](../concepts/metrics/). Each metric requires a predefined set of data points, which it uses to calculate scores that indicate performance.
+
+### Evaluating using a Non-LLM Metric
+
+Here is a simple example that uses `BleuScore` score to score summary
```
jjmachan, line 14: link to BlueScores reference or documentation

docs/getstarted/evals.md
```diff
@@ -0,0 +1,221 @@
+# Get Started
```
jjmachan, line 1: should we improve the heading?

docs/getstarted/evals.md
```diff
@@ -0,0 +1,221 @@
...
+
+**Choose your LLM**
+--8<--
+choose_evaluator_llm.md
+--8<--
+
+**Evaluation**
+
+
+Here we will use [`AspectCritic`](../concepts/metrics/available_metrics/aspect_critic.md), which an LLM based metric that outputs pass/fail given the evaluation criteria.
```
jjmachan, line 65: here, even though it is linked - it is not visible because of the backticks
image

docs/getstarted/evals.md
```diff
@@ -0,0 +1,221 @@
...
+
+![](../_static/imgs/ragas_get_started_evals.gif)
+
+
+
+## Aligning Metrics
+
+In the example above, we can see that the LLM-based metric mistakenly marks some summary as accurate, even though it missed critical details like growth numbers and market domain. Such mistakes can occur when the metric does not align with your specific evaluation preferences. For example, 
+
+![](../_static/imgs/eval_mistake1.png)
```
jjmachan, line 196: in this section we have not mentioned editing the reason, isn't that important?

jjmachan: an image of editing the reason would have been very helpful I think

docs/_static/imgs/ragas_get_started_evals.gif
jjmachan, line 1: also you don't have to save images and other static files inside _static. It's an old artifact from sphinx, MKdocs is smarter about it.

what I do is keep the images close to the file so that it is easier to work with

docs/getstarted/evals.md
shahules786: I have left a link for how to train guide, which should handle it. Here those details may be a little too detailed.

jjmachan (approved):

PR #1792, opened 2024-12-25
Update pydantic_prompt.py

kishan42:


src/ragas/prompt/pydantic_prompt.py
```diff
@@ -142,38 +142,8 @@ async def generate_multiple(
...
-        Returns
-        -------
-        List[OutputModel]
-            A list of generated outputs.
-
-        Raises
-        ------
-        RagasOutputParserException
-            If there's an error parsing the output.
-        """
```
jjmachan, line 176: why did you remove this?

jjmachan (commented):
hey @kishan42 thanks for putting in the PR but could you explain what this PR is and set some context?

PR #1795, opened 2024-12-26
Added from_url method to fetch evaluation annotation from the api.

ganeshrvel:
Added from_url method to fetch evaluation annotation from the api.
Made some optimizations.


src/ragas/dataset_schema.py
```diff
@@ -590,6 +617,81 @@ def from_json(cls, path, metric_name: t.Optional[str]) -> "MetricAnnotation":
...
         )
 
+    @classmethod
+    def from_json(cls, path: str, metric_name: t.Optional[str]) -> "MetricAnnotation":
+        """Load annotations from a JSON file"""
+        dataset = json.load(open(path))
+        return cls._process_dataset(dataset, metric_name)
+
+    @classmethod
+    def from_ragas_platform(
```
jjmachan, line 627: this brings to the point about what we are calling the thing at app.ragas.io - IMO is should be Ragas App and so the method should be `from_ragas_app`

jjmachan (approved):

src/ragas/dataset_schema.py
ganeshrvel: from_app is what I have named it right now, in case we need to be explicit, lets call it from_ragas_app

shahules786 (approved):
Just fix the ci by running `make run-ci` and all looks good. Will discuss the change of docs later.

src/ragas/dataset_schema.py
ganeshrvel: we will call it `*_platform`

jjmachan (commented):
The CI style is failing, let's fix that and merge this in 🙂

Issue #1783, opened 2024-12-22
Language Not Supported


yusufsyaifudin:
[x] I checked the [documentation](https://docs.ragas.io/) and related resources and couldn't find an answer to my question.

**Your Question**
I want to use RAGAS as my RAG evaluation framework, but I cannot find supported language other than `RAGAS_SUPPORTED_LANGUAGE_CODES` in this line https://github.com/explodinggradients/ragas/blob/v0.2.8/src/ragas/metrics/base.py#L707

Which after tracing the code, it come from here:
* https://github.com/explodinggradients/ragas/blob/v0.2.8/src/ragas/utils.py#L14-L22
* https://github.com/nipunsadvilkar/pySBD/blob/v0.3.4/pysbd/languages.py#L26-L50

The pySBD last commit is 3 years ago, which I also have question why prefer use that library? 

My ultimate question is: How to add language support which don't supported by pySBD (and not supported by RAGAS)?
I see that the list is too limited, not a single language from Southeast Asia is supported. 

**Additional context**
If I can extend the language support which don't "natively" supported by RAGAS, where I can find the example to create an Adapter Language?

Thank you!

jjmachan:
hey @yusufsyaifudin thanks for sharing this - which language are you are you planning to use? other that pySBD which other tools do you work with that have support which you mentioned?

@shahules786 should be able to provide you better information too

yusufsyaifudin:
Thanks @jjmachan for your reply

> which language are you are you planning to use?

I am work with Bahasa Indonesia and have tried to run the RAGAs with default settings (I assume it English) in three proprietary model: claude-3-haiku-20240307, claude-3-5-haiku-20241022 and claude-3-5-sonnet-20241022.

The claude-3-haiku-20240307 always return the faithfulness score to 1.0 (I only test with two data) which I can confirm that it should be near to 0. The other two models return 0.0, at this point I starting to think that maybe it just because the Haiku old version is "bad" at reasoning.

But, I think by using the same language in the prompt for testing (Bahasa Indonesia in my case), probably it would be have better reasoning.

> other that pySBD which other tools do you work with that have support which you mentioned?

Actually I don't know any alternative,  maybe we still in the state that none package supports all language for sentence boundary extractor.

But, imho, if we can create some "abstraction" regarding the sentence segmentation and prompt, we can achieve multi-language support easily? Probably using nltk, or other package.

[...]

Issue #1736, opened 2024-12-06
Relying on pySBD is not feasible- package hasn't been updated in 3 years (sentence segmentation)


diegodebrito:
**Describe the Feature**
One of RAGAS dependencies is the package pySBD. This package has several bugs but it seems to be largely unmaintained at this point. The last release was 3 years ago. This will expose RAGAS to bugs, security vulnerabilities and more.

**Why is the feature important for you?**
I found a bug related to pySBD and opened an issue there. I noticed that the repo seems to have no activity anymore. It's an extremely bad idea to have that as one of the main dependencies in my opinion. 

**Additional context**
Switching to a better known and more well-maintained package will avoid headaches for RAGAS maintainers in the future.

shahules786:
Hey @diegodebrito thanks for reporting this. Any alternative package you can recommend to do just segmentation?

diegodebrito:
Hi @shahules786, sorry for taking so long to answer. I am not a specialist on that, but maybe just standard Spacy would work for that purpose?

Something along these lines
```
import spacy

    split_sentences = []
    for answer in answers:
        doc = nlp(answer)
        sentences_with_index = {i: sent.text for i, sent in enumerate(doc.sents)}
        split_sentences.append(sentences_with_index)
```

What was the reason you went with pySBD in the first place?

shahules786:
Hey @diegodebrito We have removed the need for segmenting from ragas with the latest PR. Thank you.

PR #1826, opened 2025-01-09
fix: remove pysbd and sentence segmenting

shahules786:
fixes: 
#1783
#1736

### DIFF
--- a/src/ragas/metrics/_bleu_score.py
+++ b/src/ragas/metrics/_bleu_score.py
@@ -33,8 +33,8 @@ class BleuScore(SingleTurnMetric):
     ) -> float:
 
         reference, response = sample.reference, sample.response
-        assert isinstance(reference, str), "Expecting a string"
-        assert isinstance(response, str), "Expecting a string"
+        assert isinstance(reference, str), "BleuScore expects a valid reference string"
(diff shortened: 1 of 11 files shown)


jjmachan (approved):

src/ragas/metrics/_bleu_score.py
```diff
@@ -35,12 +31,13 @@ def init(self, run_config: RunConfig):
...
     ) -> float:
-        assert (
-            self.sentence_segmenter is not None
-        ), "Sentence segmenter is not initialized"
 
-        reference_sentences = self.sentence_segmenter.segment(sample.reference)
-        response_sentences = self.sentence_segmenter.segment(sample.response)
+        reference, response = sample.reference, sample.response
+        assert isinstance(reference, str), "Expecting a string"
+        assert isinstance(response, str), "Expecting a string"
```
jjmachan, line 37: the fact that we have to use the assert here feels not good. The other problem is the error message, ideally it should tell them what to do about it, not sure if that is easy to define here

shahules786: I have edited the message to make it clear, but this is an API design problem. The metric expects valid response and reference but in SIngleTurnSample they are optional.

Issue #1796, opened 2024-12-26
I am getting the error "TypeError(ufunc 'invert' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe'')" and context_recall as 0


tvsathish:
[ ] I checked the [documentation](https://docs.ragas.io/) and related resources and couldn't find an answer to my question.

**Your Question**
Why am I getting this error "**TypeError(ufunc 'invert' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe'')**" and why is my context_recall value showing up as 0.000000

**Code Examples**
```py
import pprint

import pandas as pd
from langchain_openai import AzureChatOpenAI
from langchain_openai import AzureOpenAIEmbeddings
from ragas import SingleTurnSample, EvaluationDataset
from ragas import evaluate
from ragas.embeddings import LangchainEmbeddingsWrapper
from ragas.llms import LangchainLLMWrapper
from ragas.metrics import LLMContextRecall, Faithfulness, FactualCorrectness, SemanticSimilarity


def empty_nan_value(cell_value):
    return '' if pd.isna(cell_value) else cell_value


def create_turn_sample(row):
    return SingleTurnSample(
        user_input=row['user_input'],
        retrieved_contexts=[empty_nan_value(row['context1']), empty_nan_value(row['context2']),
                            empty_nan_value(row['context3']),
                            empty_nan_value(row['context4'])],
        response=row['response'],
        reference=row['reference'])


df = pd.read_excel("<RESULTS_EXCEL_FILE>")

eval_dataset = EvaluationDataset([create_turn_sample(row) for index, row in df.iterrows()])

azure_config = {
    "base_url": "<AZURE_CHAT_COMPLETIONS_URL>",
    # your endpoint
    "model_deployment": "<AZURE_DEPLOYMENT>",  # your model deployment name
    "model_name": "gpt-4o"  # your model name
}

evaluator_llm = LangchainLLMWrapper(AzureChatOpenAI(
    openai_api_version="2024-08-01-preview",
    azure_endpoint=azure_config["base_url"],
    azure_deployment=azure_config["model_deployment"],
    model=azure_config["model_name"],
    validate_base_url=False,
))

metrics = [
    LLMContextRecall(llm=evaluator_llm),
    FactualCorrectness(llm=evaluator_llm),
    Faithfulness(llm=evaluator_llm)
]
results = evaluate(dataset=eval_dataset, metrics=metrics)
[...]

PR #1835, opened 2025-01-11
fix numpy invert error in metrics

oslijunw:
#1796 TypeError(ufunc 'invert' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe'') will trigger when evaluating by `FactualCorrectness`


jjmachan:
hey @oslijunw thanks a lot for putting the PR to fix this 
much appreciated man ❤️

PR #1837, opened 2025-01-11
Adjustment when calculating hash | Adjustment of the hash calculation…

joaorura:
… method

When trying to load the saved models after adaptation, alerts like these were always triggered:

Loaded prompt hash does not match the saved hash.
Loaded prompt hash does not match the saved hash.

Furthermore, in Python, the default hash() function may yield different results for the same string across different sessions. To achieve consistent hash values, for tha i using the hashlib module to calculate de hash of prompt, which provides stable hashing algorithms.


jjmachan (approved):
thanks a lot @joaorura for helping us with the Fix - really really appreciate this ❤️

PR #1838, opened 2025-01-11
Feature portuguese support

joaorura:
I have added a Portuguese compatible segmenter so you can use Portuguese ragas.


joaorura:
The changes no longer make sense with the new adjustments made.

jpfcabral:
@joaorura , what was the other way implemented to generate the dataset in Portuguese?

jjmachan:
@jpfcabral did you get a chance to try this https://docs.ragas.io/en/stable/howtos/customizations/testgenerator/_language_adaptation

PR #1848, opened 2025-01-15
fix tp&fp calc error

oslijunw:
#1796 Compatibility is being fixed due to different sample trigger points

### DIFF
--- a/src/ragas/metrics/_factual_correctness.py
+++ b/src/ragas/metrics/_factual_correctness.py
@@ -257,10 +257,10 @@ class FactualCorrectness(MetricWithLLM, SingleTurnMetric):
         else:
             response_reference = np.array([])
 
+        response_reference = np.array(response_reference, dtype=bool)
         tp = sum(reference_response)
         fp = sum(~reference_response)
         if self.mode != "precision":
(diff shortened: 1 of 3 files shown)


src/ragas/metrics/_factual_correctness.py
```diff
@@ -257,10 +257,10 @@ async def _single_turn_ascore(
         else:
             response_reference = np.array([])
 
+        response_reference = np.array(response_reference, dtype=bool)
         tp = sum(reference_response)
         fp = sum(~reference_response)
         if self.mode != "precision":
-            response_reference = np.array(response_reference, dtype=bool)
             fn = sum(~response_reference)
```
jjmachan, line 264: @shahules786 can you check this part?

jjmachan (approved):
the other parts look good - thanks a lot @oslijunw for taking the time to fix this ❤️ 🙂

src/ragas/metrics/_factual_correctness.py
shahules786: yes, this looks better.

shahules786 (approved):

PR #1858, opened 2025-01-20
docs: be consistent with the variable naming of `evaluator_embeddings`

neo:
small, but helps with copy/paste 😝


shahules786 (approved):
Yes, ofc. Thanks a lot :)

PR #1862, opened 2025-01-21
Add optional use_effective_order parameter to BLEU score

JonasElburgUVA:
The original BLEU score fails for answers shorter than four tokens because the number of 4-grams is zero. Adding the use_effective_order parameter allows users to obtain scores by considering only n-grams up to the number of tokens in the answer when ( N < 4 ). This feature is already implemented in sacreBLEU, so we only need to add the parameter to the BLEU object in ragas. The default is set to False to maintain previous behavior.

### DIFF
--- a/src/ragas/metrics/_bleu_score.py
+++ b/src/ragas/metrics/_bleu_score.py
@@ -15,7 +15,7 @@ class BleuScore(SingleTurnMetric):
         default_factory=lambda: {MetricType.SINGLE_TURN: {"reference", "response"}}
     )
     language: str = "english"
-    use_effective_order: bool=False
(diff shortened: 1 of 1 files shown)


shahules786:
Hey @JonasElburgUVA thanks for taking notice. There was a slight mistake in your implementation so I have changed, you may use it like 

```python
from ragas.metrics import BleuScore

sample = SingleTurnSample(
    response="The Eiffel Tower is located in India.",
    reference="The Eiffel Tower is located in Paris."
)

scorer = BleuScore(kwargs={"use_effective_order":True})
await scorer.single_turn_ascore(sample)```

shahules786 (approved):

PR #1873, opened 2025-01-23
Improved the documentation in the migration docs

sahusiddharth:
The current example would have caused a `KeyError: 'retrieved_contexts'` because the `retrieved_contexts` key was missing in the `SingleTurnSample` which is a required column in Faithfulness metric.


jjmachan:
nice catch @sahusiddharth ❤️

PR #1876, opened 2025-01-23
Adding missing evaluate import to evals guide

carlosgsouza:


jjmachan:
thanks for catching that @carlosgsouza 🙂

Issue #1875, opened 2025-01-23
How to solve the ("All samples must be of the same type") issue?


tvsathish:
[ ] I have checked the [documentation](https://docs.ragas.io/) and related resources and couldn't resolve my bug.

**Describe the bug**
I am getting the following error when manually creating SingleTurnSamples from my dataset.
`("All samples must be of the same type")`
How to find the data frame that contributes to a mismatched sample record?

Ragas version: 0.2.12
Python version: 3.9

**Code to Reproduce**
```py
import pprint
import re

import pandas as pd
from langchain_openai import AzureChatOpenAI
from langchain_openai import AzureOpenAIEmbeddings
from ragas import SingleTurnSample, EvaluationDataset
from ragas import evaluate
from ragas.embeddings import LangchainEmbeddingsWrapper
from ragas.llms import LangchainLLMWrapper
from ragas.metrics import LLMContextRecall, Faithfulness, FactualCorrectness, SemanticSimilarity

from urllib.request import urlopen
from bs4 import BeautifulSoup

from concurrent.futures import ThreadPoolExecutor, as_completed

def empty_nan_value(cell_value):
    return '' if pd.isna(cell_value) else cell_value


def create_turn_sample(row):
    url = re.split(r'[,\n ]+', empty_nan_value(row['reference']))[0]
    page = ''
    try:
        page = urlopen(url)
    except ValueError:
        return
    soup = BeautifulSoup(page, features='lxml')
    return SingleTurnSample(
        user_input=row['user_input'],
        retrieved_contexts=[empty_nan_value(row['context1']), empty_nan_value(row['context2']),
                            empty_nan_value(row['context3']),
                            empty_nan_value(row['context4'])],
        response=empty_nan_value(row['response']),
        reference=soup.get_text())


df = pd.read_excel("Test Automation Result.xlsx")

with ThreadPoolExecutor(max_workers=50) as executor:
    future_to_row = {
        executor.submit(create_turn_sample, row): index for index, (idx, row) in enumerate(df.iterrows(), start=0)
    }

samples = []
for future in as_completed(future_to_row):
    status = future_to_row[future]
    samples.append(future.result())

pprint.pprint(samples)

eval_dataset = EvaluationDataset(samples)

# other configuration
azure_config = {
    "base_url": <BASE_URL>,
    "model_deployment": <DEPLOYMENT_NAME>
    "model_name": "gpt-4o"  # your model name
}

[...]

PR #1879, opened 2025-01-24
Change the validate_samples functionality

Vidit-Ostwal:
Changed the validate_samples functionality to also tell which indexed sample is causing the issue. #1875


jjmachan (approved):
ahh yes this is a better error message - thanks for the PR @Vidit-Ostwal 🙂 🙌🏽

PR #1881, opened 2025-01-25
Fix Broken Test Case Due to Changed Error Message

sahusiddharth:
This PR addresses a failing test case caused by changes introduced in PR #1879, which modified the error message format. The test has been updated to reflect the new error message, ensuring it passes successfully.

### DIFF
--- a/src/ragas/dataset_schema.py
+++ b/src/ragas/dataset_schema.py
@@ -191,7 +191,9 @@ class RagasDataset(ABC, t.Generic[Sample]):
         first_sample_type = type(samples[0])
         for i, sample in enumerate(samples):
             if not isinstance(sample, first_sample_type):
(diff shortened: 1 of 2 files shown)


src/ragas/dataset_schema.py
```diff
@@ -191,7 +191,9 @@ def validate_samples(self, samples: t.List[Sample]) -> t.List[Sample]:
         first_sample_type = type(samples[0])
         for i, sample in enumerate(samples):
             if not isinstance(sample, first_sample_type):
-                raise ValueError(f"Sample at index {i} is of type {type(sample)}, expected {first_sample_type}")
+                raise ValueError(
```
sahusiddharth, line 194: The extra space here due to `make run-ci`

Vidit-Ostwal:
Thanks @sahusiddharth, for this, merging this would solve the other PR i had requested.
💪

PR #1887, opened 2025-01-28
Adding tutorial for Haystack Integration

sahusiddharth:
Integration:
- https://github.com/deepset-ai/haystack-core-integrations/pull/1312

Docs:
- https://github.com/deepset-ai/haystack-integrations/pull/297


docs/howtos/integrations/haystack.md
```diff
@@ -0,0 +1,307 @@
...
+
+Modified_Context_Utilization = ContextUtilization()
+
+prompt = Modified_Context_Utilization.get_prompts()["context_precision_prompt"]
+prompt.instruction = "Give a score of 0"
+prompt.examples = []
+Modified_Context_Utilization.set_prompts(**{"context_precision_prompt": prompt})
+```
+
+
```
jjmachan, line 232: do we need this since you conver aspect critic below?

jjmachan (approved):
looks great

docs/howtos/integrations/haystack.md
sahusiddharth: I added prompt customization to demonstrate it, and I'm open to any suggestions!

jjmachan: I think it's better to just show aspect critic bellow. This is not a very realistic modification right - users have said that is not very helpful.

you can point them to the docs instead I feel

sahusiddharth: That makes more sense, I will remove that example and point them to the docs for customisations.

PR #1901, opened 2025-02-04
Haystack llm and embedding wrapper

sahusiddharth:


jjmachan (changes requested):

src/ragas/embeddings/haystack_wrapper.py
```diff
@@ -0,0 +1,96 @@
...
+    from haystack.components.embedders import (  # type: ignore
+        AzureOpenAITextEmbedder,
+        HuggingFaceAPITextEmbedder,
+        OpenAITextEmbedder,
+        SentenceTransformersTextEmbedder,
+    )
+except ImportError:
+    raise ImportError(
+        "pip install haystack-ai is not installed. Please install it using `pip install pip install haystack-ai`."
+    )
```
jjmachan, line 20: this code still makes haystack a direct dependency for running Ragas

jjmachan: I want you to write test that 
1. removes the haystack lib installed in the dev 
2. tries and imports the module

make sure it is failing

then refer to https://github.com/BerriAI/litellm/blob/5c2839a744814712b9fdeb1dec6609333743f8e9/litellm/llms/bedrock/base_aws_llm.py#L266

here you can see that they only import `boto3` when they need it. 

write something like that for haystack and make it pass

jjmachan: the failing test before the implementation is important. In order to test this we have to remove the haystack module already installed from python environment

jjmachan (changes requested):

tests/unit/test_simple.py
```diff
@@ -1,7 +1,46 @@
...
+        return real_import(name, *args, **kwargs)
+
+    # Replace the built-in import function with our mock
+    monkeypatch.setattr(builtins, "__import__", mocked_import)
+
+    with pytest.raises(ImportError, match="Haystack is not installed"):
+        from ragas.llms import HaystackLLMWrapper
+
+        HaystackLLMWrapper(haystack_generator=None)
+
```
jjmachan: the test should check that we can import other modules too (like LangchainLLMWrapper, without any issues- that part is more important.

What I'm afraid is that because of some import Haystack will become a dependency

jjmachan (approved):
just look into the last test?

tests/unit/test_simple.py
sahusiddharth: I have added the [test](https://github.com/explodinggradients/ragas/pull/1901/files#diff-6e4f0d906982528291934672d7685c822e8176e4f8e9b32bfdd3547df6e233cd) to ensures that:
- No ImportError is raised for non-Haystack imports when haystack is missing.
- The Haystack wrapper does raise ImportError in that scenario.

Issue #1904, opened 2025-02-05
StatementGeneratorPrompt's instruction mentions "statements", but these are not part of inputs anymore


diegodebrito:
[X ] I have checked the [documentation](https://docs.ragas.io/) and related resources and couldn't resolve my bug.

**Describe the bug**

This is probably a minor one. I am just opening for awareness, and it could impact the results when using less capable LLMs.

The instruction for StatementGeneratorPrompt start with "Given a question, an answer, and sentences from the answer [...]", but sentences are not part of the inputs anymore. Just question and answer. I believe this was changed in a recent PR.

https://github.com/explodinggradients/ragas/blob/2605de5224a7c0e90f000d74f561d2dae770c18f/src/ragas/metrics/_faithfulness.py#L37

Ragas version: main branch in this repo
Python version: not applicable

**Code to Reproduce**

Not applicable

**Error trace**

Not applicable

**Expected behavior**

Instruction should reflect the fact that simple statements are extracted from the full answer now, and not the answer statements.

**Additional context**

sahusiddharth:
Hi @diegodebrito, 

Thank you for pointing that out. I’ve raised a PR to address and fix this issue.

PR #1906, opened 2025-02-06
Updated the instruction of `StatementGeneratorPrompt` in faithfulness

sahusiddharth:
#1904


jjmachan:
@shahules786 can you check the issue and the fix? I'm wondering how this effects other prompts?

shahules786 (approved):

Issue #1907, opened 2025-02-06
ragas evaluate method for the same metric with different parameters gives scores for only one metric


igrynok:
I have checked the [documentation](https://docs.ragas.io/) and related resources and couldn't resolve my bug.

**Describe the bug**
When I evaluate the same metric but with different parameters the resulting pandas dataframe contains the results of the only last one. For example, I would like to measure factual_correctness in different modes precision and recall, I got the results for recall only.

Ragas version: 0.2.12
Python version: 3.10

**Code to Reproduce**
```py
scorer1 = FactualCorrectness(llm = evaluator_llm, mode="precision")
scorer2 = FactualCorrectness(llm = evaluator_llm, mode="recall")
evaluate(dataset=dataset, metrics[scorer1, scorer2]).to_pandas()
```
**Expected behaviour**
I would like to see scores for both factual_correctness for precision and recall in the resulting dataframe.

PR #1908, opened 2025-02-07
Metric with different parameter gives different score

sahusiddharth:
fixes: #1907


jjmachan (approved):
this was actually a very good implementation - feel free to merge it in 🙂 

tip

put the PR discription as `fixes: #1907` that will automatically close it when its done

PR #1913, opened 2025-02-11
nvidia end-2-end accuracy, relevance and groundedness metrics

titericz:
first iteration of Nvidia accuracy metric
Added a simple implementation for single turn accuracy, context_relevance and answer_groundedness metrics

### DIFF
--- a/src/ragas/metrics/__init__.py
+++ b/src/ragas/metrics/__init__.py
@@ -1,8 +1,4 @@
-from ragas.metrics._nv_metrics import (
-    AnswerAccuracy,
-    ContextRelevance,
-    ResponseGroundedness,
-)
+from ragas.metrics._answer_correctness import AnswerCorrectness, answer_correctness
 from ragas.metrics._answer_relevance import (
     AnswerRelevancy,
     ResponseRelevancy,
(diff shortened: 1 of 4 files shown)


src/ragas/metrics/_answer_accuracy.py
```diff
@@ -0,0 +1,447 @@
...
+
+        return score
+
+    # For compatibility
+    async def _ascore(self, row):
+        raise NotImplementedError("You are using deprecated RAGAS version, please update to RAGAS>0.2.6")
+
+answer_accuracy = Answer_Accuracy()
+context_relevance = Context_Relevance()
+response_groundedness = Response_Groundedness()
```
jjmachan, line 447: no need for this - trying to remove instances like this because they default to OpenAI

src/ragas/metrics/__init__.py
```diff
@@ -1,4 +1,11 @@
-from ragas.metrics._answer_correctness import AnswerCorrectness, answer_correctness
+from ragas.metrics._answer_accuracy import (
```
jjmachan, line 1: I think we should change the name of the module from where we are importing this. Class name should not have a `_` in between them

jjmachan (commented):
need to figure out the naming for these metrics to avoid confusion. Just the module from where they are loaded is enough I think?

PR #1918, opened 2025-02-11
r2r integration

sahusiddharth:
- r2r integration
- fixed spelling mistake in swarm tutorial
- added @t.runtime_checkable to metric mode


shahules786 (approved):
Looks good to me

src/ragas/integrations/r2r.py
```diff
@@ -0,0 +1,241 @@
...
+    """
+    Evaluates R2R RAG using the Ragas framework with configurable metrics, LLMs, and embeddings.
+    """
+
+    def __init__(
+        self,
+        metrics: t.List[Metric],
+        evaluator_llm: t.Optional[BaseRagasLLM] = None,
+        evaluator_embeddings: t.Optional[BaseRagasEmbeddings] = None,
+    ):
```
jjmachan, line 21: why do we have an evaluator class here? that is not the pattern we do right?

jjmachan (changes requested):
next time we'll start with a design doc first for any integration before moving into the code.

docs/howtos/integrations/r2r.ipynb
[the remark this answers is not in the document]
sahusiddharth: I think we should keep ipynbs because if we decided to make changes and get the corresponding outputs it will be easier if we have the ipynb because we can simply rerun it.

src/ragas/integrations/r2r.py
sahusiddharth: Decided to go with a class because r2r had multiple things like rag, conversational rag and agents as well so it can act as a one stop thing for all.

docs/howtos/integrations/index.md
```diff
@@ -11,6 +11,7 @@ happy to look into it 🙂
 - [Langchain](./langchain.md) - Langchain is a framework for building LLM applications, more information can be found [here](https://www.langchain.com/).
 - [LlamaIndex](./_llamaindex.md) - LlamaIndex is a framework for building RAG applications, more information can be found [here](https://www.llamaindex.ai/).
 - [Haystack](./haystack.md) - Haystack is a LLM orchestration framework to build customizable, production-ready LLM applications, more information can be found [here](https://haystack.deepset.ai/).
+- [R2R](./r2r.md) - R2R is an all-in-one solution for AI Retrieval-Augmented Generation (RAG) with production-ready features, more information can be found [here](https://r2r-docs.sciphi.ai/introduction)
```
jjmachan, line 14: add it to mkdocs navigation too - otherwise it will be missing in the navbar

jjmachan (approved):
looks good - just add it to mkdocs navbar too and check the last commit I made (which should fix a merge conflict)

docs/howtos/integrations/index.md
sahusiddharth: I have added it to the [navbar](https://github.com/explodinggradients/ragas/pull/1918/files#diff-98d0f806abc9af24e6a7c545d3d77e8f9ad57643e27211d7a7b896113e420ed2)

jjmachan: my bad!

Issue #1917, opened 2025-02-11
SingleHopQuerySynthesizer::prepare_combinations()


lryan599:
[ x] I checked the [documentation](https://docs.ragas.io/) and related resources and couldn't find an answer to my question.

https://github.com/explodinggradients/ragas/blob/cb63a8266b80fcb5715b06fc97d1f9165bc49f93/src/ragas/testset/synthesizers/single_hop/base.py#L51

**Your Question**
```python
    def prepare_combinations(
        self,
        node: Node,
        terms: t.List[str],
        personas: t.List[Persona],
        persona_concepts: t.Dict[str, t.List[str]],
    ) -> t.List[t.Dict[str, t.Any]]:

        sample = {"terms": terms, "node": node}
        valid_personas = []
        persona_list = PersonaList(personas=personas)
        for persona, concepts in persona_concepts.items():
            concepts = [concept.lower() for concept in concepts]
            if any(term.lower() in concepts for term in terms):
                if persona_list[persona]:
                    valid_personas.append(persona_list[persona])
            sample["personas"] = valid_personas
            sample["styles"] = list(QueryStyle)
            sample["lengths"] = list(QueryLength)

        return [sample]
```

I think it should be

```python
    def prepare_combinations(
        self,
        node: Node,
        terms: t.List[str],
        personas: t.List[Persona],
        persona_concepts: t.Dict[str, t.List[str]],
    ) -> t.List[t.Dict[str, t.Any]]:

        sample = {"terms": terms, "node": node}
        valid_personas = []
        persona_list = PersonaList(personas=personas)
        for persona, concepts in persona_concepts.items():
            concepts = [concept.lower() for concept in concepts]
            if any(term.lower() in concepts for term in terms):
                if persona_list[persona]:
                    valid_personas.append(persona_list[persona])
        // these codes should be outside
        sample["personas"] = valid_personas
        sample["styles"] = list(QueryStyle)
        sample["lengths"] = list(QueryLength)

        return [sample]
```

PR #1921, opened 2025-02-13
fix SingleHopQuerySynthesizer::prepare_combinations()

lryan599:
This bug probably raise a KeyError since `personas` is not set.

This could be a bug due to an oversight.

Related to #1917


jjmachan:
thanks a million again @lryan599 ❤️

PR #1924, opened 2025-02-14
Ndarray dtype fix, improve error wording

zybinmikhail:
1. I have noticed that if `response.statements` is an empty list, then the output of `verify_claims` function is `array([], dtype=float64)`, which raises a type error when `~` operation is applied to `reference_response` variable. This occurs in a rare case when hypothesis_list is empty (when there are no claims in the response).
2. A small change is the choice of words in `LLMDidNotFinishException`

### DIFF
--- a/src/ragas/exceptions.py
+++ b/src/ragas/exceptions.py
@@ -37,7 +37,7 @@ class LLMDidNotFinishException(RagasException):
     """
 
     def __init__(self):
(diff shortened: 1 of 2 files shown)


jjmachan (approved):
hey @zybinmikhail thank you so much for fixing these - merging this in 🙂

PR #1933, opened 2025-02-24
fix: return valid JSON output schema

alasdairsmith:
Use json.dumps() on the dict output of Pydantics's BaseModel.model_json_schema() to ensure the string representation is valid JSON using double quotes instead of single quotes.


jjmachan:
thanks a lot @alasdairsmith this is a nice catch and a much needed fix 🙂 ❤️

PR #1946, opened 2025-03-04
OpenAI model cost based on model name fix

JonasElburgUVA:
the `CostCallbackHandler.total_cost` method can take a dictionary of models and associated cost per i/o tokens. However, this was not functional because the token usage parser did not save the model's name. The model's name was also removed after adding TokenUsage objects together. I fixed these issues for the OpenAI parser. 

Note that the same problem still exists for the anthropic and bedrock parsers. I did not update this, since I don't use these API's and would not be able to test my implementation.

### DIFF
--- a/src/ragas/cost.py
+++ b/src/ragas/cost.py
@@ -67,8 +67,10 @@ def get_token_usage_for_openai(
         return TokenUsage(input_tokens=0, output_tokens=0)
     output_tokens = get_from_dict(llm_output, "token_usage.completion_tokens", 0)
     input_tokens = get_from_dict(llm_output, "token_usage.prompt_tokens", 0)
+    model_name = get_from_dict(llm_output, "model_name", "")
 
-    return TokenUsage(input_tokens=input_tokens, output_tokens=output_tokens)
+
+    return TokenUsage(input_tokens=input_tokens, output_tokens=output_tokens, model=model_name)
 
 
 def get_token_usage_for_anthropic(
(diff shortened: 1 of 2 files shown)


jjmachan:
thanks for bringing this up @JonasElburgUVA I'll fix the others too after merging this but could you check the tests too?

JonasElburgUVA:
Sorry for the failing tests in the initial PR. The `test_cost.py` tests did not expect a model ID as part of TokenUsage objects, even though the LLMResult samples do contain model IDs. I updated the tests accordingly. I also added the model ID parsing logic for the other parsers based on the LLMResult samples in `test_cost.py`.

codestyle check is still failing, I am not sure why?

anistark:
@JonasElburgUVA Could you please rebase this PR so we can take final look at the tests.

src/ragas/cost.py
```diff
@@ -120,10 +128,17 @@ def get_token_usage_for_bedrock(
...
+                                g.message.response_metadata,
+                                "model_id",
+                                ""
+                            )
                         )
                     )
-
-        return sum(token_usages, TokenUsage(input_tokens=0, output_tokens=0))
+        model = next(
+        (usage.model for usage in token_usages if usage.model), ""
```
anistark, line 139: Please fix the Indentation

src/ragas/cost.py
```diff
@@ -92,10 +94,16 @@ def get_token_usage_for_anthropic(
...
+                            model=get_from_dict(
+                                g.message.response_metadata,
+                                "model",
+                                "")
                         )
                     )
-
-        return sum(token_usages, TokenUsage(input_tokens=0, output_tokens=0))
+        model = next(
+        (usage.model for usage in token_usages if usage.model), ""
```
anistark, line 104: Please fix the Indentation

anistark:
continued changes via #2287

PR #1947, opened 2025-03-04
fix: update docstrings for single-hop use case

terry07:
Minimal correction to Docstrings of testset regarding single hop use cases.


anistark (approved):
Thanks for the fix @terry07 🎉

PR #1951, opened 2025-03-05
Fix: Prevent 'temperature' Attribute Error for Models That Don't Support It

### LINKED ISSUE #1945: LangchainLLMWrapper sets a default temperature even with models that don't support it (ex o3-mini)

**Describe the bug**
When I try to create an LLM instance using LangchainLLMWrapper and specify a model that doesn't support the "temperature" parameter, I cannot get any responses from the API. 

Ragas version: 0.2.13
Python version: 3.10.16

**Code to Reproduce**

evaluator_llm = LangchainLLMWrapper(ChatOpenAI(model_name="o3-mini"))
sample = SingleTurnSample(
    response="The Eiffel Tower is located in Paris.",
    reference="The Eiffel Tower is located in Paris. I has a height of 1000ft."
)

scorer = FactualCorrectness(llm = evaluator_llm)
await scorer.single_turn_ascore(sample)

**Error trace**

**Expected behavior**

**Additional context**

khalilacheche:
This PR fixes an issue where specifying a model that does not support the temperature parameter in LangchainLLMWrapper resulted in an API error and prevented responses from being generated.

Changes Made:
Modified the attribute check for temperature in LangchainLLMWrapper.
Now, the check ensures that temperature is both present and not None before applying it.
Fixes #1945

### DIFF
--- a/src/ragas/llms/base.py
+++ b/src/ragas/llms/base.py
@@ -204,7 +204,7 @@ class LangchainLLMWrapper(BaseRagasLLM):
         old_temperature: float | None = None
         if temperature is None:
             temperature = self.get_temperature(n=n)
-        if hasattr(self.langchain_llm, "temperature"):
+        if hasattr(self.langchain_llm, "temperature") and temperature is not None:
             self.langchain_llm.temperature = temperature  # type: ignore
             old_temperature = temperature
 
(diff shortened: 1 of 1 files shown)


src/ragas/llms/base.py
```diff
@@ -204,7 +204,7 @@ def generate_text(
         old_temperature: float | None = None
         if temperature is None:
             temperature = self.get_temperature(n=n)
-        if hasattr(self.langchain_llm, "temperature"):
+        if hasattr(self.langchain_llm, "temperature") and temperature is not None:
             self.langchain_llm.temperature = temperature  # type: ignore
```
anistark, line 208: consider adding a try except block here.

anistark (changes requested):
The current check only verifies attribute existence, not setting of temperature, which might fail due to model restrictions.

anistark:
Closing as already fixed. The current codebase has an existing solution using the `bypass_temperature` mechanism.

PR #1959, opened 2025-03-11
Single-hop query testset generation tutorial

sahusiddharth:
Contains:
- Md of tutorials
- Dummy dataset
- Changes for the navigations in docs


shahules786 (changes requested):
Overall Looks great, Just one comment - make the ragas airline dummy data ignore from mkdocs index otherwise it will appear in every search. For example https://ragas--1959.org.readthedocs.build/en/1959/howtos/applications/singlehop_testset_gen/?h=

shahules786:
@sahusiddharth Move all files related to the dataset (markdown files such as flight policies to notes repo or huggingface or your personal repos) - modify the tutorial then use it from that source. 
We don't keep such data and artifacts along with docs.

shahules786 (approved):
LGTM, nice work @sahusiddharth

PR #1966, opened 2025-03-15
directory name change in docs

sahusiddharth:


shahules786 (approved):
thanks for taking care of that Sid

PR #1967, opened 2025-03-17
fix: optimize knowledge graph clustering for large corpus

aubford:
Current implementation of find_indirect_clusters runs at exponential time because the depth-first search always explores every path in the graph. A  kg w/ ~3000 relationships takes over an hour on a 2024 M4 Macbook Pro. 
This brings it down to quad/cubic time relative to _testset size_ (instead of kg relationships). Generating a 100 sample testset of abstract multihop queries for a KG of 1 million relationships takes about 20 seconds.

- Added new find_n_indirect_clusters method and applied it to MultiHopAbstractQuerySynthesizer.
- No longer searches the entire graph, just enough to get a well-diversified randomized sample of clusters for the desired testset size.
- Added lots of tests. I can scale them back. They're mostly to demo the spec so you can easily compare the new behavior with the original find_indirect_clusters which I left in place.
- New behavior adds randomization and diversity to cluster sampling.
  - Will no longer return subsets along with their supersets, only the superset.
- Also fixed the other performance bottleneck in MultiHopAbstractQuerySynthesizer (collecting child nodes).
- Eliminates potentially expensive edge-case where LLM calls could be made for every possible cluster in the KG.
- In depth details can be found in the find_n_indirect_clusters docstring.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -164,6 +164,10 @@ cython_debug/
 #  option (not recommended) you can uncomment the following to ignore the entire idea folder.
 .idea/
 
+# Cursor
+.cursorignore
+.cursor/
+
 # Ragas specific
 experiments/
 **/fil-result/
--- a/src/ragas/testset/graph.py
+++ b/src/ragas/testset/graph.py
@@ -320,7 +320,7 @@ class KnowledgeGraph:
         depth_limit: int = 3,
     ) -> t.List[t.Set[Node]]:
         """
-        Finds up to n indirect clusters of nodes in the knowledge graph based on a relationship condition.
+        Return n indirect clusters of nodes in the knowledge graph based on a relationship condition.
         Optimized for large datasets by using an adjacency index for lookups and limiting path exploration
         relative to n.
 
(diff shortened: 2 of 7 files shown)


jjmachan:
Hey @aubford! 👋

Hope you're doing well! I really loved your work on optimizing knowledge graph clustering for large corpus - it's exactly the kind of thoughtful improvement that makes Ragas better for everyone.

Basically we're wondering if we should keep it as part of the core library, spin it off into its own thing, or maybe even retire it if folks aren't really using it much. No pressure at all, but given your experience with performance optimization and clustering algorithms, your perspective would be super helpful!

Thanks for being awesome! 🙏

anistark (changes requested):
It seems tests are failing.
Could you please rebase and check for the changes?

anistark:
Have taken these changes in #2267

aubford:
> It seems tests are failing. Could you please rebase and check for the conflicts?

Hi, sorry I'm just seeing this. Did you move these changes to a new PR?  It looks like those tests are passing on #2267.  If it's still an issue, I can take a look.

jjmachan:
hey @aubford no worries - could you take a look at #2267 when you get a chance?
we'll merge it in anyways and would love your feedback 🙂

aubford:
Hey @jjmachan.  Yes #2267 looks good to me.  Thanks!

PR #1968, opened 2025-03-17
fix: need ot encode post body as utf-8 to avoid exceptions on TestSet.upload()

aubford:
Without this I get exceptions due to right quotes not being encoded correctly.

### DIFF
--- a/src/ragas/sdk.py
+++ b/src/ragas/sdk.py
@@ -101,7 +101,7 @@ def upload_packet(path: str, data_json_string: str):
 
     response = requests.post(
         f"{base_url}/api/v1{path}",
-        data=data_json_string,
+        data=data_json_string.encode('utf-8'),
         headers=headers,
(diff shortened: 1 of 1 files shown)


src/ragas/sdk.py
```diff
@@ -101,7 +101,7 @@ def upload_packet(path: str, data_json_string: str):
 
     response = requests.post(
         f"{base_url}/api/v1{path}",
-        data=data_json_string,
+        data=data_json_string.encode('utf-8'),
```
anistark, line 104: The change may fix UTF-8 encoding issues but introduces potential runtime errors if `data_json_string` isn't always a string type.

Consider adding a type checking:

```py
data=data_json_string.encode('utf-8') if isinstance(data_json_string, str) else data_json_string,
```

something like that.

anistark:
Closing as we don't have `upload_packet` function anymore.

Issue #1915, opened 2025-02-11
Wrong cache hits for LlamaIndexEmbeddingsWrapper


triedel42:
[x] I have checked the [documentation](https://docs.ragas.io/) and related resources and couldn't resolve my bug.

**Describe the bug**
When generating embeddings with the `LlamaIndexEmbeddingsWrapper` a wrong cache hit leads to all embeddings being the same.
This is due to the [`_generate_cache_key` function](https://github.com/explodinggradients/ragas/blob/cb63a8266b80fcb5715b06fc97d1f9165bc49f93/src/ragas/cache.py#L150) removing the first argument for methods. However it seems `self` is not passed here as an argument, hence valid arguments are sliced away [here](https://github.com/explodinggradients/ragas/blob/cb63a8266b80fcb5715b06fc97d1f9165bc49f93/src/ragas/cache.py#L151) and every cache key becomes the same.

Ragas version: 0.2.13
Python version: 3.11.2

**Code to Reproduce**
[see this gist](https://gist.github.com/triedel42/5afae2c2e43913355508a159b5956a62)

**Error trace**
N/A (see assertion in example code)

**Expected behavior**
Distinct `embed_text` arguments need to be cached under distinct cache keys.

**Additional context**
[This test case](https://github.com/explodinggradients/ragas/blob/cb63a8266b80fcb5715b06fc97d1f9165bc49f93/tests/unit/test_cache.py#L29) did not catch it either, since it's only tested with functions, not methods.

anistark:
Should be fixed with #1977 and will be released in next week release.
If still occurs, feel free to reopen. Thanks for bringing it up @triedel42 🙌🏼

PR #1977, opened 2025-03-23
Fix issue with cache hits

Nicba1010:
As per #1915

Args will not contain the `self` argument, it is bound to func in `__self__` so there is no need to try and strip it.


Nicba1010:
I covered a test, but I have not set up the project or run it. I don't know if there is a GitHub action auto running tests to validate that the PR is "merge-able". I would appreciate if someone could verify the tests work.

Nicba1010:
Are there any plans to merge this at all? If so, I'll resolve the conflicts, if not, just tell me.

PR #2018, opened 2025-04-18
Update the logic of answer_relevance as per the definition

mshobana:
Calling generate 3 times gives the questions with exact words every time. generate_multiple tends to give semantically matching questions.


anistark:
Thanks for the PR @mshobana  
This is already done in PR #1978 

If there's any changes you recommend, feel free to open a new one.

PR #2023, opened 2025-04-23
docs: add CTA for Ragas app

shahules786:


docs/getstarted/rag_eval.md
```diff
@@ -176,26 +176,19 @@ Output
 {'context_recall': 1.0000, 'faithfulness': 0.8571, 'factual_correctness': 0.7280}
 ```
 
-## Analyze Results
+### Exclusive access to Ragas App (soon to be open-sourced)
```
jjmachan, line 179: - lets not promise the OSS yet
- I dont' know if "Exclusive access" matches what you are saying bellow?

jjmachan: "We’ll work shoulder-to-shoulder with your team to build your own evaluation loop—custom to your use case.

No vague dashboards or useless metrics, Just sharp feedback, fast iteration, and real progress." this feels more list consulting - which is what we do but "exclusive access" does not match that in that case?

jjmachan (commented):
for the deleted files did you remove it from 
1. navigation, mkdocs.yaml
2. other links pointing to it?

docs/getstarted/rag_eval.md
shahules786: Please suggest an alternative?

shahules786: I tried a different narrative. How about this?

jjmachan: the one above "Want help in improving your AI application using evals?" looks better right? are you planning to add that here too?

PR #2030, opened 2025-05-07
feature!: knowledge graph flexibility

ahgraber:
The current implementations of the synthesizers (in particular, `MultiHopAbstractQuerySynthesizer` and `MultiHopSpecificQuerySynthesizer`) have hard-coded values for the node- and relationship- properties/types they look for when generating scenarios and samples.  This works when using the default knowledge graph transforms, but may cause unexpected behavior (or force users to use the expected hardcoded values when they do not actually correspond to the reality of property lineage) when using customized the knowledge graph transformations.

This PR allows flexibility in defining the node-property, relationship-property, and relationship-type expected by the *QuerySynthesizers, while retaining the current behavior as default

The bulk of this PR is nonbreaking, but there may be a breaking typo fix in the `SingleHopSpecificQuerySynthesizer` default name.

- feat: allow user-defined node- and relationship- properties in MultiHopAbstractQuerySynthesizer
  - Users should be able to set the relationship property to use for identifying clusters and the node property used for identifying abstract concepts.
  - This will not change default behavior, but allows users to override.
- feat: allow user-defined node- and relationship- properties in MultiHopSpecificQuerySynthesizer
  - Users should be able to set the relationship type to use for identifying clusters and the relationship property used for identifying overlapping concepts within the triple.
  - This will not change default behavior, but allows users to override.
- fix!: typo in SingleHopSpecificQuerySynthesizer name (`single_hop_specifc_query_synthesizer` -> `single_hop_specific_query_synthesizer`)
- docs: minor docstring updates to SingleHopSpecificQuerySynthesizer

### DIFF
--- a/src/ragas/testset/synthesizers/multi_hop/abstract.py
+++ b/src/ragas/testset/synthesizers/multi_hop/abstract.py
@@ -31,21 +31,19 @@ logger = logging.getLogger(__name__)
 
 @dataclass
 class MultiHopAbstractQuerySynthesizer(MultiHopQuerySynthesizer):
-    """
-    Synthesizes abstract multi-hop queries from given knowledge graph.
-
-    Attributes
-    ----------
-    """
+    """Synthesize abstract multi-hop queries from given knowledge graph."""
 
     name: str = "multi_hop_abstract_query_synthesizer"
+    relation_property: str = "summary_similarity"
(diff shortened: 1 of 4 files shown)


anistark (approved):
Thanks for the PR @ahgraber 
LGTM

PR #2037, opened 2025-05-13
update compare_embeddings doc links

resitaydin:


anistark (approved):
Thanks for the PR @resitaydin 

This file isn't in use, but at least it fixes broken links.

PR #2071, opened 2025-06-10
Reduce find_indirect_clusters() complexity from O(exp(n)) to O(n)

mludvig:
I was trying to generate a testset from `KnowledgeGraph(nodes: 1861, relationships: 912103)` and it would go on for two days in a row without any progress. So I did some profiling to see what's going on on a smaller KG and found that most of the time was being spent in `find_indirect_clusters()`.

Here is an optimised implementation that seem to produce similar results but way faster. On my reduced size KG I'm down from 5 min to a fraction of a second and on the full KG I'm down from _never finished_ to a few seconds.

### DIFF
--- a/ragas/src/ragas/testset/graph.py
+++ b/ragas/src/ragas/testset/graph.py
@@ -256,61 +256,59 @@ class KnowledgeGraph:
         depth_limit: int = 3,
     ) -> t.List[t.Set[Node]]:
         """
-        Finds indirect clusters of nodes in the knowledge graph based on a relationship condition.
-        Here if A -> B -> C -> D, then A, B, C, and D form a cluster. If there's also a path A -> B -> C -> E,
-        it will form a separate cluster.
+        Finds clusters (connected components) of nodes in the knowledge graph reachable within
+        a limited number of hops, according to a relationship condition.
+
+        A cluster is defined as a set of nodes such that each node in the cluster can be reached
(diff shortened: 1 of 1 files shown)


Lukevanl:
Ran into the exact same issue and indeed made a similar fix, would be nice to get it merged since this limits the current code only to very small graphs :)

anistark:
Thanks for the PR @mludvig 🙌🏼  It does seemigly fix a critical issue.

Could you please add a test as well? So, we can validate this works and also for future serve as our north star.

Also, please rebase and check on the merge conflict.

jjmachan:
Hey @mludvig! 👋

Hope you're doing well! I really loved your work on reducing find_indirect_clusters() complexity from O(exp(n)) to O(n) - it's exactly the kind of thoughtful improvement that makes Ragas better for everyone.

Basically we're wondering if we should keep it as part of the core library, spin it off into its own thing, or maybe even retire it if folks aren't really using it much. No pressure at all, but given your experience with algorithm optimization and performance improvements, your perspective would be super helpful!

Thanks for being awesome! 🙏

anistark:
Hi @mludvig 

Have you also checked the diff in result from old and new approaches?

Would recommend adding a new function rather than replace the existing ones to have existing users not break, while providing an alternate algo with hops as you've suggested.

anistark:
Closing this as #2144 covered the changes.

PR #2074, opened 2025-06-13
Corrected instruction prompt

Sdegraauw:
analyse was spelled "analyze". this might influance the LLM slightly.


Sdegraauw:
Ok i just learned it's a difrence between british and american english. this could still influence LLM's and i would recomend testing wich of the 2 more consitantly get's correct results.

anistark:
"analyze" is prefered in most LLMs so we can keep this for now, even though in regular language we might prefer "analyse". Thanks for bringing it up. 🙌🏼

PR #2077, opened 2025-06-13
Feature: improve relationship builders for better async and reduced memory utilization

ahgraber:
- CosineSimilarityBuilder now uses a sharded/chunked similarity calculation to significantly reduce memory requirements
- CosineSimilarityBuilder and JaccardSimilarityBuilder now leverage generate_execution_plan to support async iteration over tasks (for potential future multithreading or improved concurrency)
- Added unit tests

### DIFF
--- a/ragas/pyproject.toml
+++ b/ragas/pyproject.toml
@@ -63,6 +63,7 @@ dev = [
     "haystack-ai",
     "sacrebleu",
     "r2r",
+    "scipy",
 ]
 test = [
     "pytest",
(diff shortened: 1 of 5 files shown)


anistark:
Thanks for the PR @ahgraber 
Could you please rebase and check for the merge conflicts?

ahgraber:
@anistark - I've merged the changes from upstream

anistark:
> @anistark - I've merged the changes from upstream

Thanks @ahgraber 
However, I see the type check ci failed. Could you please run `make type` locally to verify the issue.

ahgraber:
It works on my computer (TM).  
Screenshot 2025-08-26 at 9 40 54 AM

ahgraber:
@anistark I've cleanly rebased.

Re: requested changes:

- _extra param changing the method signature. Try default value for the extra param._ -> now uses the `CosineSimilarityBuilder` attribute (which is new, but has a default) instead of altering the method signature.
- _sync to async conversion. Probably better to provide both sync and async versions._ -> have reverted to sync
- _return type change from List -> Set. Is there a pure benefit or we can use Set internally and keep List for backwards compatibility._ -> have reverted to List. Set is more efficient (set is used internally for fast deduplication), but list conversion post-hoc is trivial relative to the cosine similarity calculations.

PR #2082, opened 2025-06-18
fix issue 1857

crisschan:
add  OllamaLLMWrapper class major at change from ChatOllama type to ChatOpenAI.


crisschan:
add langchain_ollama to /ragas/pyproject.toml

anistark:
Upon further checking, this PR might not be needed. 
Ollama already works with existing langchain and ragas has moved to use factory rather than wrappers. 

In case, there's an issue with the usage, can take up as a separate PR.

PR #2094, opened 2025-06-27
feat: align and validate

shahules786:


shahules786:
@jjmachan I noticed that the naming `train` confuses people to think we are going to do weights training, so I changed it to `align`.

jjmachan:
why shouldn't align always through the return for validation_alignment ? that should be the output if there is not return for `.align()` right?

shahules786:
@jjmachan This comment is very unclear 
> why shouldn't align always through the return for validation_alignment ? that should be the output if there is not return for .align() right?

But reading through the lines, you're asking why align can't always call validate_align at the end and return that. 
I have added the function for that `align_and_validate(...)`

PR #2109, opened 2025-07-15
examples : example evals

shahules786:
install examples
```
pip install ragas_experimental[examples]
```

Run any examples from prompt, rag, llm workflow or agent
```
python -m ragas_examples.prompt_evals.evals
```


shahules786:
So, I did consider that while I was building it. I want the namespace to be ragas_examples, so it doesn’t get mixed up with some generic module examples that may already exist in users system. Just like pydantic_examples
Pydantic follows similar approach to name it as `pydantic-ai-examples` @jjmachan 
https://github.com/pydantic/pydantic-ai/tree/main/examples/pydantic_ai_examples

jjmachan:
aah understood - then let's do that

PR #2128, opened 2025-07-22
Fix: make the answer be the same as the split sentences

wlbksy:
The answer is not the same as the split sentences in ContextRecallClassifications's statements, and there are some grammar mistakes.


anistark:
Thanks for the PR @wlbksy Please rebase and fix the ci issues.

You can run `make run-ci` locally to checkout the exact issue.

PR #2132, opened 2025-07-23
fix: modification of the translate instruction

lovets18:
**Changed**
Modified instruction for translating promtps.

**Reason**
I have encountered an error trying to use `adapt_prompts` for `Faithfulness` with LLM Gigachat:

> ValueError: The number of statements in the output (4) does not match the number of statements in the input (1). Translation failed.

The instruction of the metric contains: "Break down each sentence into one or more fully understandable statements.". Model decided to execute this, not only translate, the translated prompt was splitted, so the shape mismatch occured


anistark:
Thanks for the PR @lovets18

Could you please rebase and fix the ci issues?

Try running make run-ci locally for detailed output.

anistark:
Merged in #2284

PR #2137, opened 2025-07-23
fix: enum values instead of names in testset synthesizers

lovets18:
**Change**
Using Enum.value instead of Enum.name

It seems that value (description) for QueryStyle is more logical than using names. The current name and value are almost identical, but this might change in the future


lovets18:
@anistark Thanks for the feedback! Updated the commit message to describe the change

anistark:
Thanks @lovets18 🎉

PR #2144, opened 2025-07-25
Reduce find_indirect_clusters() runtime through neighborhood detection and sampling

ahgraber:
Like @mludvig found in #2071, I notice the find_indirect_clusters' use of exhaustive depth-first search to be a significant barrier to use on any, even moderately substantial, knowledge graph. That PR uses BFS to identify a set of disjoint clusters involving the source node (each node appears in at most one cluster) whereas the original find_indirect_clusters  identifies all sets of clusters up to length depth_limit from each node. @mludvig, if I'm out of line here, please correct me!

The approach in this PR instead identifies neighborhoods in the graph using a Leiden clustering algorithm and samples from the neighborhoods.  I believe this to be a better approach - in my testing it is even faster than BFS, and the function returns something more in line to the original `find_indirect_clusters` implementation.

I would have preferred (and, in fact, originally tried) using a Clique Percolation Method approach because it allows nodes to belong to multiple neighborhoods; however CPM also ends up running into NP-hard / NP-complete runtime issues.

This PR does add two dependencies to Ragas - networkx and scikit-network.


ragas/tests/unit/test_graph.py
```diff
@@ -0,0 +1,427 @@
...
+        "\u2665",  # heart
+        "\u2605",  # star
+        "\u221e",  # infinity
+        "\u00b5",  # micro
+        "\u2264",  # less than or equal
+        "\u2265",  # greater than or equal
+        "\u0391",  # Greek letters
+        "\u0392",
+        "\u0393",
+        "\uffff",  # Special Unicode characters
```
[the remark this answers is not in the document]
ahgraber: this is an inherited test; I did not make changes.

ragas/tests/unit/test_graph.py
```diff
@@ -0,0 +1,427 @@
...
+        #         print(f"Extra clusters: {[set(s) for s in extra]}")
+        # else:
+        #     print("\n✓ All clusters match!")
+        # print("=" * 40)
+
+        self.assert_sets_equal(clusters, expected_clusters)
+
+    def test_with_cycle(self, simple_graph):
+        # above test_with_depth_limit uses simple_graph which already has cycles
+        pass
```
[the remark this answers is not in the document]
ahgraber: I believe it is important to note that (a) a test with cycles was considered and (b) why it is unnecessary.

pyproject.toml
```diff
@@ -19,6 +19,8 @@ dependencies = [
     "instructor",
     "gitpython",
     "pillow>=10.4.0",
+    "networkx",
+    "scikit-network",
```
anistark, line 23: This is a big dependency. Perhaps better to place this under optional dependency?

src/ragas/testset/graph.py
```diff
@@ -250,67 +252,209 @@ def __repr__(self) -> str:
...
+            # so we use type: ignore to bypass them.
+            # graph: sknetwork.data.Dataset
+            graph = from_edge_list(  # type: ignore
+                [(str(rel.source.id), str(rel.target.id)) for rel in relationships],
+                directed=True,
+            )
 
+            # Apply Leiden clustering
+            leiden = Leiden(random_state=42)
+            cluster_labels = leiden.fit_predict(graph.adjacency)  # type: ignore
```
anistark: Can we check for the typing fixes in these cases if possible instead of ignoring  them?

ahgraber: I've reduced `# type: ignore` to a single instance - sknetwork's `from_edge_list` doesn't properly overload its return signature, so it is ignored.

pyproject.toml
ahgraber: It's necessary any time anyone wants to build a testset, though. I don't know if it makes sense to make it optional in that sense? It seems like it would be frustrating to try to build your test set and then be told "oh, you didn't install all of the dependencies, try again".
Or perhaps (in a separate change) there should be a set of optional 'testset' dependencies?

ahgraber:
> Do we know the result diff due to algorithm change?

For small graphs, no change.  For large graphs, they are able to be used without running out of memory and in the span of human lifetime.  

When the clustering + sampling approach is used, there will inevitably be some difference in result since we are sampling rather than exhaustively identifying all paths. However, the core logic of `find_indirect_clusters` is retained in this approach (unlike #2071 ) and the clustering/sampling just provides an upper bound on the amount of time/compute spent on mining indirect relationships.

I suppose one potential problem with the current sampling approach is that each neighborhood is sampled 1000 times, rather than proportionally.  To sample proportionately, we'd need to alter the method signature of `find_indirect_clusters` to specify a maximum number of "indirect clusters" to return and then proportionally sample from each neighborhood.  Conversely, the "benefit" of the equivalent sampling approach is that even long-tail neighborhoods are equivalently represented in the test set.

pyproject.toml
anistark: Fair. I was also thinking on both sides of it.

PR #2147, opened 2025-07-29
Fix repeating template bug in _nv_metrics.py

dev-nid:
ContextRelevance class uses the same template(template_relevance1) in both retry loop, instead of template_relevance2.

Replaced template_relevance1 in second retry loop with template_relevance2

### DIFF
--- a/docs/getstarted/evals.md
+++ b/docs/getstarted/evals.md
@@ -62,7 +62,7 @@ choose_evaluator_llm.md
 **Evaluation**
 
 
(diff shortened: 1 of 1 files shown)


anistark:
Thanks @dev-nid for the fix.

Could you please rebase and check why CI is failing.

dev-nid:
Raising a new PR since this one closed automatically when I synced and I cannot reopen.

https://github.com/explodinggradients/ragas/pull/2216

PR #2152, opened 2025-07-31
feat: save and load prompt object

jjmachan:


jjmachan:
these were the consideration I was thinking about
- is .json the best format for saving `Prompt`?
- how should we think about save/load `ExampleStore` and `InMemoryExampleStore`
- how can we add support for python fsspec
- what do you think of the json structure, are we missing anything there?

experimental/ragas_experimental/prompt/base.py
```diff
@@ -80,13 +138,154 @@ def add_example(self, inputs: t.Dict, output: t.Dict) -> None:
...
+        but the model itself cannot be serialized. You'll need to provide it when loading.
+        """
+        if self.response_model:
+            warnings.warn(
+                "response_model cannot be saved and will be lost. "
+                "You'll need to set it manually after loading using: "
+                "Prompt.load(path, response_model=YourModel)"
+            )
+
+        data = {
```
anistark, line 169: Should define this structure implementation somewhere.

anistark: Also, can add stuff like:

```json
"config": {
    "similarity_threshold": 0.7,
    "max_examples": 5
},
"metadata": {
    "created_at": "...",
    "total_examples": 150,
    "ragas_version": "..."
}
```

anistark:
> * is .json the best format for saving `Prompt`?

I think yes.

> * how can we add support for python fsspec

Can do as a utility class. Can use the local storage on client side.

PR #2165, opened 2025-08-13
refactor: experiment function to use model as additional optional arg

anistark:


experimental/ragas_experimental/experiment.py
```diff
@@ -79,6 +79,7 @@ async def __call__(self, *args, **kwargs) -> t.Any: ...
     async def arun(
         self,
         dataset: Dataset,
+        model: t.Optional[t.Any] = None,
         name: t.Optional[str] = None,
         backend: t.Optional[t.Union[BaseBackend, str]] = None,
     ) -> "Experiment": ...
```
jjmachan: I think what we need to do here is `arun` and `run` take args and kwargs optionally and pass it over to the function so users might want to add other args too, not just `model`. 

The idea behind `@experiment` is that users will write out how one run should be and we provide all the orchastration to scale it to all the samples in the dataset. in the case of `model` that is something we pass down to the function itself, so I feel we don't have to specify it

anistark: Alright. That's better for scale too.

anistark:
Pending discussion on API structure.

jjmachan (approved):

PR #2166, opened 2025-08-13
Fixed a small typo in context_precision.md

ImadSaddik:
Hi!

While reading [this page](https://docs.ragas.io/en/stable/concepts/metrics/available_metrics/context_precision/#example_2), I have found that the `s` in `retrieved_contexts` was not inside the backticks.

**Before**:
`retrieved_context`s

**After**:
`retrieved_contexts`


anistark (approved):
Thanks for the quick fix @ImadSaddik

PR #2177, opened 2025-08-16
adapt fabrix for ragas

eric-epsilla:


anistark:
Thanks for your PR @eric-epsilla 
Could you please rebase and fix the merge conflicts so we can proceed with review.

anistark:
Its better to use fabrix as a custom llm metric as it might not be relevant for larger audience.

PR #2178, opened 2025-08-16
Remove black

NirantK:
1. Remove black
2. Applied `ruff format` to code which was inconsistent with ruff and black
3. Relevant edits in Makefile


jjmachan (approved):

workspace.toml
jjmachan, line 1: I think we should now remove workspace.toml file since this is now not a monorepo anymore

NirantK: See #2182 — @anistark can you take this up?

anistark: Sure, will take a look.

PR #2179, opened 2025-08-16
refactor: enhance required_columns logic to exclude ignored markers

ubiratanfilho:
Added ":ignore" suffix to prevent in a evaluation run with custom metrics like AspectCritic or RubricsScore to run all metrics with all variables, to reduce token usage


anistark:
Thanks for the PR @ubiratanfilho  

While this might work on most cases, we need to be ready for unknown as well.

Please also check for validation. Can add to `VALID_COLUMNS`
and tests which cover `:ignored`. Checkout `test_metric.py`

Finally, `get_required_columns` should handle `:ignored` when `with_optional=True`.

anistark:
Tracked in #2298

PR #2180, opened 2025-08-17
Remove txt requirements

NirantK:
Continuing our conversation from Slack, this PR removes requirements folder and it's usage in CI


NirantK:
We've a flaky test or two in analytics, rest looks good.

PR #2189, opened 2025-08-19
Merge workspace.toml into pyproject.toml

### LINKED ISSUE #2182: Migrate workspace.toml to pyproject.toml
I think we should now remove workspace.toml file since this is now not a monorepo anymore

_Originally posted by @jjmachan in https://github.com/explodinggradients/ragas/pull/2178#discussion_r2280723723_

NirantK:
Moved pyproject.toml to where it belongs. This led to some surprises, which are also handled in this PR. 

1. format docs and tests. They were excluded so far because of path issues as a silent error. 
2. Similar issues with pyright. 

Important: I've excluded few things from pyright for now, fixed the rest. 
We should remove exclusions. I'll leave that for @anistark 

Fixes #2182


pyproject.toml
```diff
@@ -68,28 +68,58 @@ test = []
...
 
 [tool.setuptools.dynamic]
-readme = {file = ["README.md"], content-type = "text/markdown"}
+readme = {file = ["ragas/README.md"], content-type = "text/markdown"}
 
-# Ruff configuration is inherited from workspace.toml at the monorepo root
+[tool.ruff]
+line-length = 88
+target-version = "py39"
+exclude = ["*.ipynb", "*/_version.py"]  # Exclude Jupyter notebooks and auto-generated version files from linting
```
jjmachan, line 87: - `_version` is never commited right?
- why did we ignore *.ipynb?

pyproject.toml
```diff
@@ -68,28 +68,58 @@ test = []
...
+[tool.ruff.format]
+quote-style = "double"
+indent-style = "space"
+skip-magic-trailing-comma = false
+docstring-code-format = false
+preview = false
+
+[tool.pyright]
+include = ["ragas/src/ragas"]
+exclude = ["ragas/src/ragas/experimental", "ragas/src/ragas/_analytics.py", "ragas/src/ragas/testset", "ragas/src/ragas/optimizers", "ragas/src/ragas/metrics"]
```
jjmachan, line 108: this ignores a lot of the core modules right? are we sure we need this

pyproject.toml
```diff
@@ -68,28 +68,58 @@ test = []
...
-# Point to the repository root (one level up)
-root = ".."
+version_file = "ragas/src/ragas/_version.py"
 
 [tool.pytest.ini_options]
 addopts = "-n 0"
 asyncio_default_fixture_loop_scope = "function"
-[pytest]
-testpaths = ["tests"]
+testpaths = ["ragas/tests"]
```
jjmachan, line 130: is this correct?

pyproject.toml
NirantK: If we lint `ipynb` as well, it creates multiple check and format issues. So as a convention, I am excluding it, and that is continued from the previous `pyproject.toml`

pyproject.toml
NirantK: Yes, I asked @anistark to take this up. Don't want this PR to get bigger 😅

pyproject.toml
NirantK: Looks right to me and I cross-checked the CI run as well

pyproject.toml
NirantK: > _version is never commited right?

It breaks local dev often enough to bother me.

PR #2191, opened 2025-08-19
Refactor LLM structure and add InstructorLLM

NirantK:


ragas/src/ragas/llms/__init__.py
```diff
@@ -1,15 +1,23 @@
 from ragas.llms.base import (
     BaseRagasLLM,
+    InstructorBaseRagasLLM,
+    InstructorLLM,
     LangchainLLMWrapper,
     LlamaIndexLLMWrapper,
+    T,
```
jjmachan: maybe make this a more explicit name?

jjmachan (approved):

ragas/src/ragas/llms/__init__.py
NirantK: Good catch. On it.

anistark (approved):
LGTM

PR #2193, opened 2025-08-19
deprecate: add testset deprecation notice

anistark:


anistark:
Closing for now as this needs more discussion.

PR #2199, opened 2025-08-19
fix: setup and local ci

anistark:
- Removing `ragas/Makefile` and will use single `Makefile` going forward.
- Fixed failing local ci
- Added a setup venv for first time setup.


anistark:
> I honestly think there are waay too many commands in our makefile right now.
> 
> we just need these IMO
> 
> ```
> make format         # Format ragas code only
> make type           # Type check ragas code only
> make check          # Quick format + type check
> make test           # Run all tests (core + experimental)
> make run-ci         # Run ragas CI pipeline
> ```

Also, 

- install
- clean (doesn't work right now though)
- test with notebooks
- docs cmds

But we can cleanup in separate PR post discussion. This PR would fix the commit flow for now and make setup clean.

Issue #2185, opened 2025-08-17
`make install` fails due to incompatible dependency configuration


aicodexp:
[x] I have checked the [documentation](https://docs.ragas.io/) and related resources and couldn't resolve my bug.

---

## Description

The current development setup has an incompatibility between the root Makefile and `ragas/pyproject.toml` that prevents contributors from setting up the development environment as documented.

### Current Behavior

Running `make install` fails with:
```bash
$ make install
Installing dependencies...
Installing ragas dependencies (including experimental)...
...
warning: The package `ragas @ file:///home/max/ragas/ragas` does not have an extra named `dev`
Setting up pre-commit hooks...
make: pre-commit: No such file or directory
make: *** [Makefile:21: install] Error 127
```

### Environment
- OS: MX Linux 23
- Python: 3.11.2
- uv version: 0.8.11
- (RAGAS version /) Repository commit: f16676764 (main branch)

### Root Cause

1. **Dependency format mismatch**: The root Makefile attempts to install using extras syntax:
   ```makefile
   $(Q)uv pip install -e "./ragas[dev]"
   ```
   However, `ragas/pyproject.toml` defines dev dependencies using `[dependency-groups]` (PEP 735) instead of `[project.optional-dependencies]`, so there is no `dev` extra available. This means `pre-commit` is never installed, causing the hook setup to fail.

2. **Directory structure issue**: When using `uv pip install --group dev`, uv expects a `pyproject.toml` in the current working directory, but in the ragas repository structure, it's located at `ragas/pyproject.toml`. The root Makefile needs to either use `--directory ragas` or `cd ragas` to access it.

## Steps to Reproduce

1. Clone the repository
2. Install uv: `curl -LsSf https://astral.sh/uv/install.sh | sh` 
3. Create venv: `uv venv`
4. Activate: `source .venv/bin/activate`
5. Run `make install`
6. Observe the error

## Expected Behavior

`make install` should successfully install all necessary development dependencies and set up pre-commit hooks.

### Option 1: Use dependency-groups with correct directory (leads to excessive downloads)
```makefile
$(Q)uv pip install -e "./ragas"
$(Q)cd ragas && uv pip install --group dev
# OR: $(Q)uv pip install --directory ragas --group dev
```

**Problem**: This installs 236+ packages including PyTorch, CUDA libraries (~1.5GB download) because the dev group includes:
```toml
"ragas[all,tracing,gdrive,ai-frameworks]"
```
[...]

PR #2200, opened 2025-08-20
feat: separate minimal and full dev installations

NirantK:
This is a thinking out loud PR about how I am thinking on some changes in `pyproject.toml`
 
## Summary
- Add minimal dev installation option with core development tools only (ruff, pytest, pyright)
- Keep full installation with complete ML stack and experimental features for advanced development
- Update documentation and build commands to support both installation types

Fixes https://github.com/explodinggradients/ragas/issues/2185

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -14,16 +14,47 @@ The repository contains:
 
 ## Development Environment Setup
 
-### Installation
+### Installation Options
+
+Choose the appropriate installation based on your development needs:
+
+#### Minimal Development Setup
+Recommended for:
+- Code formatting, linting, and type checking
+- Running unit tests
+- Basic development tasks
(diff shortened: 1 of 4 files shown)


aicodexp:
Hello @NirantK  - you were much quicker than me! 🚀

I would have done the same. And yes, the reference to the no longer existing "experimental" must also be removed. It's great that you have updated Claude and the Development markdown files as well.

However, the distinction between "normal" and "minimal" does increase the complexity of course. I haven't been around long enough to assess how long or for what "minimal" might be sufficient.

**If** that's a valid use case, though, why not aim for the merge?

Interesting: #2199

NirantK:
Closing this in favour of #2218 which we will review and merge

Issue #2170, opened 2025-08-15
Error importing module


rizasaputra:
[ v ] I have checked the [documentation](https://docs.ragas.io/) and related resources and couldn't resolve my bug.

**Describe the bug**
Error `module 'pyarrow' has no attribute 'PyExtensionType'` directly appeared after installing and importing ragas

Ragas version: 0.3.1
Python version: 3.11

**Code to Reproduce**
In notebook:

!pip install ragas
from ragas import evaluate

**Error trace**
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[7], line 12
      1 # from ragas.metrics import (
      2 #     ContextRelevance,
      3 #     ResponseGroundedness, 
   (...)
     10 #     EvaluationDataset
     11 # )
---> 12 from ragas import evaluate
     13 # from ragas.llms import LangchainLLMWrapper
     14 # import ragas

File [/opt/conda/lib/python3.11/site-packages/ragas/__init__.py:2](https://lt7bqomdzjry51g.studio.us-west-2.sagemaker.aws/opt/conda/lib/python3.11/site-packages/ragas/__init__.py#line=1)
      1 from ragas.cache import CacheInterface, DiskCacheBackend, cacher
----> 2 from ragas.dataset_schema import EvaluationDataset, MultiTurnSample, SingleTurnSample
      3 from ragas.evaluation import evaluate
      4 from ragas.run_config import RunConfig

File [/opt/conda/lib/python3.11/site-packages/ragas/dataset_schema.py:13](https://lt7bqomdzjry51g.studio.us-west-2.sagemaker.aws/opt/conda/lib/python3.11/site-packages/ragas/dataset_schema.py#line=12)
     11 import numpy as np
     12 import requests
---> 13 from datasets import Dataset as HFDataset
     14 from pydantic import BaseModel, field_validator
     16 from ragas._version import __version__

File [/opt/conda/lib/python3.11/site-packages/datasets/__init__.py:37](https://lt7bqomdzjry51g.studio.us-west-2.sagemaker.aws/opt/conda/lib/python3.11/site-packages/datasets/__init__.py#line=36)
     34 del pyarrow
     35 del version
---> 37 from .arrow_dataset import Dataset, concatenate_datasets
     38 from .arrow_reader import ReadInstruction
     39 from .builder import ArrowBasedBuilder, BeamBasedBuilder, BuilderConfig, DatasetBuilder, GeneratorBasedBuilder

[...]

PR #2201, opened 2025-08-20
Closes Issue #2170: Address issue 2170 by specifying 'datasets' core dependency version to >=4.0.0

mnedelko:
The fix includes specifying the 'datasets' core dependency to use the latest datasets version >=4.0.0. 

This fixes an issue where pip accidentally resolves to a lower version due to a dependency-resolution chain that causes an unfavourable outcome which leads to ragas breaking at the import step.

Please test ragas comprehensively with this fix in place before merging.

### DIFF
--- a/ragas/pyproject.toml
+++ b/ragas/pyproject.toml
@@ -6,7 +6,7 @@ license = {file = "LICENSE"}
 dependencies = [
     # Core dependencies
     "numpy",
-    "datasets",
+    "datasets>=4.0.0",
     "tiktoken",
     "pydantic>=2.0.0",
     "nest-asyncio",


anistark:
Thanks @mnedelko for the PR 🙌🏼 

Overall looks fine. Might need to fix the tests as well.

Try running: `make test-e2e`

  - E2E tests fail when trying to load
  `load_dataset("explodinggradients/amnesty_qa", "english_v3")`
  - This affects files:
    - `tests/e2e/test_amnesty_in_ci.py`
    - `tests/e2e/test_fullflow.py`
    - `tests/benchmarks/benchmark_eval.py`

mnedelko:
I was unable to run make teste-2e due to the below: Trying to run it in docker instead.

```
error: Distribution `torch==2.8.0 @ registry+https://pypi.org/simple` can't be installed because it doesn't have a source distribution or wheel for the current platform

hint: You're on macOS (`macosx_13_0_x86_64`), but `torch` (v2.8.0) only has wheels for the following platforms: `manylinux_2_28_aarch64`, `manylinux_2_28_x86_64`, `macosx_11_0_arm64`, `win_amd64`; consider adding your platform to `tool.uv.required-environments` to ensure uv resolves to a version with compatible wheels
```

Create a separate issue for this here: https://github.com/explodinggradients/ragas/issues/2208

mnedelko:
The reason the tests fail is because datasets>=4.0.0 removed support for scripts, which affects the load_dataset operation in the following files:

1. tests/e2e/test_amnesty_in_ci.py - Line 18:
  amnesty_qa = load_dataset("explodinggradients/amnesty_qa", "english_v3")["eval"]
  2. tests/e2e/test_fullflow.py - Line 14:
  ds = load_dataset("explodinggradients/amnesty_qa", "english_v3")["eval"]
  3. tests/benchmarks/benchmark_eval.py - Line 19:
  ds = load_dataset("explodinggradients/amnesty_qa", "english_v2")

All three files are trying to load the "explodinggradients/amnesty_qa" dataset, which uses a custom Python script that's no longer supported in datasets>=4.0.0.

 Additionally, there are many documentation files (notebooks and markdown) that also reference these datasets:
  - Various documentation notebooks also use load_dataset("explodinggradients/fiqa", ...)
  - Documentation also references load_dataset("explodinggradients/earning_report_summary", ...)

  These datasets likely also use custom scripts and would fail with datasets>=4.0.0.

The recommended solution is as follows:

**1. Update the Datasets on Hugging Face Hub**

  The dataset maintainers should migrate the datasets to the new format without Python scripts:
  - Convert explodinggradients/amnesty_qa to use Parquet/CSV/JSON files instead of amnesty_qa.py
  - Convert explodinggradients/fiqa similarly
  - This ensures compatibility with datasets>=4.0.0 and future versions
  
  PS: There used to also be a second solution which allowed one to use trusted_remote to true but this option had also been removed from datasets for security reasons.

anistark:
Covered in PR #2222 
includes the commits of this PR.

PR #2229, opened 2025-08-27
fix: add support for both langchain llm and ragas llm

### LINKED ISSUE #2085: Incorrect usage of await llm.generate(...) instead of await llm.agenerate(...) in async context

Summary

In recent versions of Langchain, llm.generate(...) is now a synchronous method, while llm.agenerate(...) is the asynchronous alternative. However, in the current Ragas codebase (e.g., pydantic_prompt.py), I see await llm.generate(...) being used.

This results in a runtime error, because llm.generate() is not a coroutine and cannot be awaited.

Ragas version: 0.2.15
Python version: 3.11
Langchain version: >=0.2.43

**Code to Reproduce**
	1.	Install Ragas 0.2.15 and Langchain 0.2.43+
	2.	Attempt to evaluate using a Langchain LLM
	3.	Ragas internally awaits llm.generate(...) — leads to runtime error

**Error trace**
TypeError: object LLMResult can't be used in 'await' expression

**Expected behavior**
Either 
await llm.agenerate(...)
Or
llm.generate(...)

If compatibility with older Langchain versions is needed, a conditional fallback could be added:
if hasattr(llm, "agenerate"):
    result = await llm.agenerate(prompts)
else:
    result = llm.generate(prompts)

**Additional context**
N/A

anistark:

- Fixes #2085 

## Changes Made

- Handle both LangChain LLMs and Ragas LLMs.


src/ragas/prompt/pydantic_prompt.py
```diff
@@ -87,7 +87,7 @@ def to_string(self, data: t.Optional[InputModel] = None) -> str:
 
     async def generate(
         self,
-        llm: BaseRagasLLM,
+        llm: t.Union[BaseRagasLLM, t.Any],
```
jjmachan: I think this will break the contract. we are only suppose to have BaseRagasLLM here

src/ragas/prompt/pydantic_prompt.py
```diff
@@ -187,13 +187,26 @@ async def generate_multiple(
...
+                callbacks=prompt_cb,
+            )
+        else:
+            resp = await llm.generate(
+                prompt_value,
+                n=n,
+                temperature=temperature,
+                stop=stop,
+                callbacks=prompt_cb,
+            )
```
jjmachan: but this is correct

src/ragas/prompt/pydantic_prompt.py
jjmachan: after reading it carefully I think the question is why are you passing t.Any here

anistark: to fix the typing issue for langchain llm.

src/ragas/prompt/multi_modal_prompt.py
```diff
@@ -101,7 +101,7 @@ def to_prompt_value(self, data: t.Optional[InputModel] = None):
 
     async def generate_multiple(
         self,
-        llm: BaseRagasLLM,
+        llm: t.Union[BaseRagasLLM, t.Any],
```
jjmachan: my concern is this t.Any since it will break type checking.

anistark: Used `from langchain_core.language_models import BaseLanguageModel` instead now.

jjmachan (approved):

PR #2266, opened 2025-09-09
Improved context precision documentation

anupamck:

- Fixes #[issue_number]
- OR describe the issue: What problem does this solve? How can it be replicated

The existing documentation has a mistake, and in some cases, poor style and grammar or is missing key information. 

## Changes Made

- Provided separate example for explaining Context Precision as a metric
- Corrected the definition of the metric LLMContextPrecisionWithoutReference
- Improved Grammar and style
- For NonLLMContextPrecisionWithReference, added packages to be installed and also included an example of the distance algorithm used

---

### DIFF
--- a/docs/concepts/metrics/available_metrics/context_precision.md
+++ b/docs/concepts/metrics/available_metrics/context_precision.md
@@ -13,14 +13,6 @@ $$
 
 Where $K$ is the total number of chunks in `retrieved_contexts` and $v_k \in \{0, 1\}$ is the relevance indicator at rank $k$.
 
-## LLM Based Context Precision
-
-The following metrics uses LLM to identify if a retrieved context is relevant or not.
-
-### Context Precision without reference
-
-`LLMContextPrecisionWithoutReference` metric can be used without the availability of a reference answer, by using the retrieved contexts. To estimate if the retrieved contexts are relevant, this method uses the LLM to compare each chunk in `retrieved_contexts` with the `response`.
-
 #### Example
     
 ```python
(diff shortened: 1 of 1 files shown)


docs/concepts/metrics/available_metrics/context_precision.md
```diff
@@ -87,9 +79,38 @@ Output
...
 
+## LLM Based Context Precision
+
+The following metrics uses LLM to identify if a retrieved context is relevant or not.
+
+### Context Precision without reference
+
+The `LLMContextPrecisionWithoutReference` metric can be used without the availability of a reference answer. To estimate if the retrieved contexts are relevant, this method uses the LLM to compare each chunk in `retrieved_contexts` with the `response`.
+
+#### Example
```
anistark, line 90: There's an example above which is identical. Don't suppose we need another one which shows exact same thing.

docs/concepts/metrics/available_metrics/context_precision.md
```diff
@@ -13,14 +13,6 @@ $$
 
 Where $K$ is the total number of chunks in `retrieved_contexts` and $v_k \in \{0, 1\}$ is the relevance indicator at rank $k$.
 
-## LLM Based Context Precision
```
anistark, line 16: Why to move this down?

anupamck: Well, to structure the document better. 

Right now, the document has an example that explains the metric of context precision itself (and how its value changes if we add irrelevant context chunks in various positions). We then move to the three subcategories of context precision, each with its own example. 

The example with `LLMContextPrecisionWithoutReference` is repeated twice, since the first time I have used it to illustrate Context Precision as a whole, and the second time is an example of this specific sub-category metric. 

Previewing this document [here](https://github.com/anupamck/ragas/blob/main/docs/concepts/metrics/available_metrics/context_precision.md) and reading it from top to bottom make this change clearer.

docs/concepts/metrics/available_metrics/context_precision.md
anupamck: My response to your other comment answers this question.

docs/concepts/metrics/available_metrics/context_precision.md
anistark: Makes sense. Will look at it again. Thanks for the explanation and also for raising this. 🙌🏼

PR #2267, opened 2025-09-09
Knowledge graph/optimize for large corpus

anistark:
contd... #1967


czhiming-maker:
Could you please let me know when this change will be released? I've hit a bottleneck in this area recently. It's taking more than three hours to build a knowledge graph with 1,000 nodes, and it hasn't even generated the scenarios yet.

anistark:
> Could you please let me know when this change will be released? I've hit a bottleneck in this area recently. It's taking more than three hours to build a knowledge graph with 1,000 nodes, and it hasn't even generated the scenarios yet.

Next release is scheduled for tomorrow. But this PR is merged in case you wanna try it out from source.

PR #2283, opened 2025-09-15
feat(metrics): Add Risk-Control Metric Suite

### LINKED ISSUE #2279: feat: Addition of Risk-Control Metrics for Trustworthy RAG Evaluation
### Describe the Feature

This proposal introduces a new, cohesive suite of four interconnected and academically-grounded metrics designed to evaluate a critical dimension currently missing from `ragas`: **a RAG system's ability to control risk by knowing when to refuse to answer.**

Instead of only evaluating the quality of a *generated answer*, these metrics evaluate the quality of the RAG system's preceding *meta-decision*: whether to answer a question ("keep") or to proactively abstain ("discard") when confidence is low or the context is insufficient.

The four proposed metrics are:
*   **Risk:** Measures the probability that an answer the system chose to provide is actually a "risky" one (i.e., from an unanswerable context). A lower score is better, directly quantifying the system's safety.
*   **Carefulness:** Measures the system's ability to correctly identify and discard questions that are unanswerable from the context. This is effectively the recall of the "abstain" decision. A higher score is better.
*   **Alignment:** Measures the overall accuracy of the keep/discard decision-making process, providing a holistic view of the system's judgment.
*   **Coverage:** Measures the proportion of questions the system attempts to answer, quantifying its "helpfulness" or utility.

Together, these metrics provide developers with a "dashboard" to understand and tune the fundamental trade-off between system safety (low Risk) and helpfulness (high Coverage).

### Why is the feature important for you?

The current `ragas` library is excellent for the post-hoc evaluation of RAG answer quality. However, as RAG systems move from research prototypes to **production-grade, enterprise, and safety-critical applications**, the most dangerous failure mode is a confidently incorrect answer (hallucination). A system that safely says "I don't know" is far more valuable and trustworthy than one that guesses.

This feature is critical because it addresses this "evaluation gap" by:

1.  **Shifting the Paradigm:** It moves  from a purely "best-effort" evaluation model (judging the quality of an attempt) to a "safety-first" model that also judges the wisdom of making the attempt at all.
2.  **Enabling Trustworthy AI:** It gives developers the tools to measure and optimize for reliability. One can now set concrete goals like, "Maintain Coverage above 80% while ensuring Risk remains below 5%."
3.  **Aligning with Real-World Needs:** In domains like finance, legal, and medical tech, preventing the propagation of misinformation is paramount. These metrics make `ragas` an even more indispensable tool for developers in these high-stakes fields.

### Additional context

**Academic Grounding**
These metrics are directly derived from the peer-reviewed research paper **"Controlling Risk of Retrieval-augmented Generation: A Counterfactual Prompting Framework"** by Chen et al., presented at EMNLP 2024 Findings. This provides a rigorous, scientifically validated foundation for their inclusion in `ragas`.

*   **Paper Link:** [Controlling Risk of Retrieval-augmented Generation](https://arxiv.org/abs/2409.16146)

AlanPonnachan:

- Fixes #2279
- **Problem:** The `ragas` library currently excels at evaluating the quality of generated answers but lacks metrics to assess a RAG system's trustworthiness and risk-control mechanisms. Specifically, it cannot measure a system's ability to recognize uncertainty and proactively abstain from answering when the retrieved context is insufficient or irrelevant. This is a critical capability for deploying reliable RAG systems in production and safety-critical domains.

## Changes Made

- **Added `_risk_control.py`:** Introduced a new file `src/ragas/metrics/_risk_control.py` which contains the implementation for a new suite of four interconnected metrics:
    - `Risk`: Measures the probability of a "risky" answer (lower is better).
    - `Carefulness`: Measures the ability to correctly discard unanswerable questions.
    - `Alignment`: Measures the overall accuracy of the keep/discard decision.
    - `Coverage`: Measures the proportion of questions the system attempts to answer.
- **Added `risk_control_suite` factory function:** This function efficiently initializes all four metrics, sharing a single calculation pass over the dataset to improve performance.
- **Updated `metrics/__init__.py`:** Exposed the new metrics (`Risk`, `Carefulness`, `Alignment`, `Coverage`) and the `risk_control_suite` factory function to make them accessible to users.

### How to Test
- **Automated:** A new test file `tests/unit/test_risk_control.py` has been added. It includes comprehensive unit tests that verify:
    - Correct calculation of all four metrics on a sample dataset.
    - Correct handling of edge cases (e.g., no "kept" answers, no "unanswerable" questions).
    - Proper error handling for missing required columns.
- **To run the tests:**
  ```bash
  pytest tests/unit/test_risk_control.py

### DIFF
--- a/src/ragas/metrics/_risk_control.py
+++ b/src/ragas/metrics/_risk_control.py
@@ -85,13 +85,13 @@ class _RiskControlCalculator:
         return self._scores
 
 
-@dataclass
+@dataclass(kw_only=True)
 class Risk(SingleTurnMetric):
     """
     Measures the probability that an answer provided by the system is a "risky"
     one (i.e., it should have been discarded). A lower Risk score is better.
-    """    
-    calculator: _RiskControlCalculator = field(repr=False)
+    """
+    calculator: _RiskControlCalculator
     name: str = "risk"
     _required_columns: t.Dict[MetricType, t.Set[str]] = field(default_factory=dict)
 
@@ -102,13 +102,13 @@ class Risk(SingleTurnMetric):
         return self.calculator.get_scores()["risk"]
 
 
-@dataclass
+@dataclass(kw_only=True)
 class Carefulness(SingleTurnMetric):
     """
     Measures the system's ability to correctly identify and discard unanswerable
     questions. It is effectively the recall for the "unanswerable" class.
     """
-    calculator: _RiskControlCalculator = field(repr=False)
+    calculator: _RiskControlCalculator
     name: str = "carefulness"
     _required_columns: t.Dict[MetricType, t.Set[str]] = field(default_factory=dict)
     
@@ -119,13 +119,13 @@ class Carefulness(SingleTurnMetric):
         return self.calculator.get_scores()["carefulness"]
 
 
-@dataclass
+@dataclass(kw_only=True)
 class Alignment(SingleTurnMetric):
     """
     Measures the overall accuracy of the model's decision-making process
     (both its decisions to keep and to discard).
     """
-    calculator: _RiskControlCalculator = field(repr=False)
+    calculator: _RiskControlCalculator
     name: str = "alignment"
     _required_columns: t.Dict[MetricType, t.Set[str]] = field(default_factory=dict)
 
@@ -136,13 +136,13 @@ class Alignment(SingleTurnMetric):
         return self.calculator.get_scores()["alignment"]
 
 
-@dataclass
+@dataclass(kw_only=True)
 class Coverage(SingleTurnMetric):
     """
     Measures the proportion of questions that the system attempts to answer.
     It quantifies the system's "helpfulness" or "utility."
     """
-    calculator: _RiskControlCalculator = field(repr=False)
+    calculator: _RiskControlCalculator
     name: str = "coverage"
     _required_columns: t.Dict[MetricType, t.Set[str]] = field(default_factory=dict)
 
@@ -151,6 +151,7 @@ class Coverage(SingleTurnMetric):
 
     async def _single_turn_ascore(self, sample: SingleTurnSample, callbacks: Callbacks) -> float:
         return self.calculator.get_scores()["coverage"]
+
 
 
 def risk_control_suite(dataset: Dataset) -> list[Metric]:
--- a/src/ragas/metrics/_risk_control.py
+++ b/src/ragas/metrics/_risk_control.py
@@ -1,0 +1,166 @@
+from __future__ import annotations
+
+import typing as t
+from dataclasses import dataclass, field
+
+from ragas.dataset_schema import SingleTurnSample
+from ragas.metrics.base import Metric, MetricType, SingleTurnMetric
+from ragas.run_config import RunConfig
+
+if t.TYPE_CHECKING:
+    from datasets import Dataset
+    from langchain_core.callbacks import Callbacks
+
+
+@dataclass
+class _RiskControlCalculator:
+    """
+    A private helper class to perform the dataset-wide calculations for the risk control suite.
+    This class is instantiated once and shared across all four metrics to ensure the calculation
+    is performed only once.
+    """
+
+    dataset: Dataset
+    _scores: dict[str, float] | None = field(default=None, init=False, repr=False)
+
+    def _calculate(self) -> None:
+        """
+        Iterates through the dataset to count the four outcomes (AK, UK, AD, UD) and
+        computes the four risk-control metrics.
+        """
+        required_columns = {"ground_truth_answerable", "model_decision"}
+        for col in required_columns:
+            if col not in self.dataset.column_names:
+                raise ValueError(
+                    f"Missing required column '{col}' in the dataset for Risk-Control metrics. "
+                    "Please ensure your dataset contains 'ground_truth_answerable' (boolean) and 'model_decision' ('kept'/'discarded') columns."
+                )
+
+        # The four outcomes
+        ak_count, uk_count, ad_count, ud_count = 0, 0, 0, 0
+
+        for row in self.dataset:
+            is_answerable = row["ground_truth_answerable"]
+            decision_is_kept = row["model_decision"].lower() == "kept"
+
+            if is_answerable and decision_is_kept:
+                ak_count += 1
+            elif not is_answerable and decision_is_kept:
+                uk_count += 1
+            elif is_answerable and not decision_is_kept:
+                ad_count += 1
+            elif not is_answerable and not decision_is_kept:
+                ud_count += 1
+
+        total_kept = ak_count + uk_count
+        total_unanswerable = uk_count + ud_count
+        total_decisions = ak_count + uk_count + ad_count + ud_count
+
+        # Risk: Probability that a kept answer is risky. Lower is better.
+        risk = uk_count / total_kept if total_kept > 0 else 0.0
+
+        # Carefulness: Recall for the "unanswerable" class. Higher is better.
+        carefulness = ud_count / total_unanswerable if total_unanswerable > 0 else 0.0
+
+        # Alignment: Overall accuracy of the keep/discard decision. Higher is better.
+        alignment = (ak_count + ud_count) / total_decisions if total_decisions > 0 else 0.0
+
+        # Coverage: Proportion of questions the system attempts to answer. Higher is better.
+        coverage = total_kept / total_decisions if total_decisions > 0 else 0.0
+
+        self._scores = {
+            "risk": risk,
+            "carefulness": carefulness,
+            "alignment": alignment,
+            "coverage": coverage,
+        }
+
+    def get_scores(self) -> dict[str, float]:
+        """
+        Returns the calculated scores. If not already calculated, triggers the calculation.
+        """
+        if self._scores is None:
+            self._calculate()
+        assert self._scores is not None
+        return self._scores
+
+
+@dataclass
+class Risk(SingleTurnMetric):
+    """
+    Measures the probability that an answer provided by the system is a "risky"
+    one (i.e., it should have been discarded). A lower Risk score is better.
+    """    
+    calculator: _RiskControlCalculator = field(repr=False)
+    name: str = "risk"
+    _required_columns: t.Dict[MetricType, t.Set[str]] = field(default_factory=dict)
+
+    def init(self, run_config: RunConfig):
+        pass
+
+    async def _single_turn_ascore(self, sample: SingleTurnSample, callbacks: Callbacks) -> float:
+        return self.calculator.get_scores()["risk"]
+
+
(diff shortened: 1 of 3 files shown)


src/ragas/metrics/_risk_control.py
```diff
@@ -0,0 +1,167 @@
...
+    async def _single_turn_ascore(self, sample: SingleTurnSample, callbacks: Callbacks) -> float:
+        return self.calculator.get_scores()["coverage"]
+
+
+
+def risk_control_suite(dataset: Dataset) -> list[Metric]:
+    """
+    Factory function to create the suite of four risk-control metrics.
+    """
+    calculator = _RiskControlCalculator(dataset)
```
anistark, line 161: Add validation for required columns.

src/ragas/metrics/_risk_control.py
```diff
@@ -0,0 +1,167 @@
...
+        pass
+
+    async def _single_turn_ascore(self, sample: SingleTurnSample, callbacks: Callbacks) -> float:
+        return self.calculator.get_scores()["coverage"]
+
+
+
+def risk_control_suite(dataset: Dataset) -> list[Metric]:
+    """
+    Factory function to create the suite of four risk-control metrics.
```
anistark, line 159: Add more details on usage for docs.

anistark (commented):
Thanks for the PR @AlanPonnachan 

A few questions that comes up:

- What happens if the underlying dataset changes after metric
   creation?
- How does this scale with very large datasets?
- Each `_single_turn_ascore` call recalculates for the
  entire dataset - is this intentional?

Please also update the docs and check out the failing ci. You can check locally by running `make run-ci`.

While at it, please also rebase with main.

AlanPonnachan:
Hi @anistark ,

Thank you so much for the insightful feedback on the initial proposal.  It helped me understand the core `ragas` evaluation architecture much more deeply.

Based on your feedback, I've completely refactored the implementation to address every concern. Before I update the PR, I wanted to propose the new design here to make sure we're aligned.

### Understanding the Challenge: A Corpus-Level Metric

The core challenge, as I now understand it, is that metrics like `Risk` and `Carefulness` are fundamentally **corpus-level** (or dataset-level). The score is an aggregate (`Total UK / Total Kept`) and cannot be calculated from a single row in isolation. This contrasts with row-level metrics like `faithfulness`.

The `evaluate()` function, however, is designed to call `_single_turn_ascore()` on a per-row basis. The design challenge is to bridge this gap in an efficient and stateless way that respects the `ragas` architecture.

### Proposed Solution: Lazy, Cached Corpus Calculation

The proposed solution is to make the metric objects themselves stateless and perform a lazy, one-time calculation for the entire dataset on the first row processed. The results are then cached for all subsequent rows in that evaluation run.

This is achieved by leveraging the internal `__ragas_dataset__` attribute that the `evaluate` function attaches to each sample.

Here’s a snapshot of the new design pattern in `_risk_control.py`:

```python
from weakref import WeakKeyDictionary
# ... other imports

# A module-level cache to store results per dataset object
_calculator_cache: WeakKeyDictionary[Dataset, dict[str, float]] = WeakKeyDictionary()

def _calculate_scores_for_dataset(dataset: Dataset) -> dict[str, float]:
    """
    Performs the one-time, full-dataset calculation and caches the result.
    """
    if dataset in _calculator_cache:
        return _calculator_cache[dataset]
    
    # ... full scan and calculation logic ...
    
    _calculator_cache[dataset] = scores
    return scores


@dataclass
class Risk(SingleTurnMetric):
    name: str = "risk"
    # ...

    async def _single_turn_ascore(self, sample: SingleTurnSample, callbacks: Callbacks) -> float:
        """
        For each row, get a reference to the parent dataset and compute scores.
        The calculation only runs on the first call; subsequent calls are instant cache hits.
        """
        dataset = getattr(sample, "__ragas_dataset__")
        scores = _calculate_scores_for_dataset(dataset)
        return scores["risk"]

# Users now import the singleton instances directly, no factory needed.
risk = Risk()
carefulness = Carefulness()

**Comment by AlanPonnachan** [comment]
Hi @anistark ,

Thank you so much for the insightful feedback on the initial proposal.  It helped me understand the core `ragas` evaluation architecture much more deeply.

Based on your feedback, I've completely refactored the implementation to address every concern. Before I update the PR, I wanted to propose the new design here to make sure we're aligned.

### Understanding the Challenge: A Corpus-Level Metric

The core challenge, as I now understand it, is that metrics like `Risk` and `Carefulness` are fundamentally **corpus-level** (or dataset-level). The score is an aggregate (`Total UK / Total Kept`) and cannot be calculated from a single row in isolation. This contrasts with row-level metrics like `faithfulness`.

The `evaluate()` function, however, is designed to call `_single_turn_ascore()` on a per-row basis. The design challenge is to bridge this gap in an efficient and stateless way that respects the `ragas` architecture.

### Proposed Solution: Lazy, Cached Corpus Calculation

The proposed solution is to make the metric objects themselves stateless and perform a lazy, one-time calculation for the entire dataset on the first row processed. The results are then cached for all subsequent rows in that evaluation run.

This is achieved by leveraging the internal `__ragas_dataset__` attribute that the `evaluate` function attaches to each sample.

Here’s a snapshot of the new design pattern in `_risk_control.py`:

```python
from weakref import WeakKeyDictionary
# ... other imports

# A module-level cache to store results per dataset object
_calculator_cache: WeakKeyDictionary[Dataset, dict[str, float]] = WeakKeyDictionary()

def _calculate_scores_for_dataset(dataset: Dataset) -> dict[str, float]:
    """
    Performs the one-time, full-dataset calculation and caches the result.
    """
    if dataset in _calculator_cache:
        return _calculator_cache[dataset]
    
    # ... full scan and calculation logic ...
    
    _calculator_cache[dataset] = scores
    return scores


@dataclass
class Risk(SingleTurnMetric):
    name: str = "risk"
    # ...

    async def _single_turn_ascore(self, sample: SingleTurnSample, callbacks: Callbacks) -> float:
        """
        For each row, get a reference to the parent dataset and compute scores.
        The calculation only runs on the first call; subsequent calls are instant cache hits.
        """
        dataset = getattr(sample, "__ragas_dataset__")
        scores = _calculate_scores_for_dataset(dataset)
        return scores["risk"]

# Users now import the singleton instances directly, no factory needed.
risk = Risk()
carefulness = Carefulness()

**Comment by anistark** [comment]
@AlanPonnachan This sounds like a good plan.

A few thoughts:

1. We won't be able to use metrics outside `evaluate()` as there'll be no `__ragas_dataset__`. So, also won't be able to test indidual metrics in isolation.
2. Cache not updating with if dataset changes. 

```py
dataset = Dataset.from_list([...])
results1 = evaluate(dataset, [risk])        # Calculates and caches

dataset = dataset.add_row(new_data)  # Dataset changed
results2 = evaluate(dataset, [risk])        # Uses old cached results
```

3. If ragas changes internals, such as renames `__ragas_dataset__` to `__dataset__` this'll break. However this I can think of as something that would be changed alongside. But better to not hardcode.


_How about if you can think of a hybrid approach which combines using factory for control and singleton instances for standard use?_

AlanPonnachan:
@anistark 

 I've designed a new hybrid solution that I believe addresses every point.


The new design isolates the core calculation logic into a **pure, standalone function** that is completely independent of the `ragas` framework. The metric classes then act as thin, thread-safe wrappers around this testable core.

This is managed through a factory function, `risk_control_suite()`, which creates a shared cache for a single `evaluate()` run, ensuring efficiency and safety.

```python
# FILE: src/ragas/metrics/_risk_control.py

import asyncio
from dataclasses import dataclass

def _calculate_scores_for_dataset(dataset: Dataset) -> dict[str, float]:
    """The PURE, TESTABLE core logic. No framework dependencies."""
    # ... (full calculation logic) ...

@dataclass
class _RiskCache:
    scores: dict[str, float] | None = None
    lock: asyncio.Lock = field(default_factory=asyncio.Lock)

def risk_control_suite() -> list[Metric]:
    """Factory to create the suite with a shared, run-specific cache."""
    cache = _RiskCache()
    return [Risk(cache=cache), Carefulness(cache=cache), ...]

@dataclass(kw_only=True)
class Risk(SingleTurnMetric):
    cache: _RiskCache = field(default_factory=_RiskCache)

    async def _single_turn_ascore(self, sample: SingleTurnSample, ...) -> float:
        async with self.cache.lock: # Prevent race conditions
            if self.cache.scores is None:
                dataset = getattr(sample, "__ragas_dataset__")
                self.cache.scores = _calculate_scores_for_dataset(dataset)
        return self.cache.scores["risk"]
```

---

### How This Addresses Your Feedback:

1.  **Testability:** The core logic is now in `_calculate_scores_for_dataset`, which is a pure function. Our unit tests can call this directly, making the metric **fully testable in isolation**.

2.  **Cache Staleness:** The cache is no longer global. The `risk_control_suite()` factory creates a fresh cache for each `evaluate()` run, so state cannot leak between runs. This guarantees correctness.

3.  **Fragile `__ragas_dataset__`:** The dependency is now architecturally contained. The core, testable logic is pure. The fragile `getattr` call is isolated in a thin adapter layer (`_single_turn_ascore`), minimizing the "blast radius" of any future internal changes in `ragas`. (this seems to be a **necessary trade-off** for any corpus-level metric that needs to integrate seamlessly into the existing row-level `evaluate()` loop.)

4.  **Hybrid Approach:** This is achieved perfectly. The `risk_control_suite()` factory provides the controlled, efficient path for users, while the metrics remain robust. I've also added an `asyncio.Lock` to ensure it's safe for parallel execution.

This design feels like a great balance of robustness and simplicity. If this direction looks good to you, I will update the PR with this final implementation.

rajantripathi:
Strong contribution — risk-aware evaluation is exactly what production RAG systems need, and the "Carefulness" metric fills a gap that standard faithfulness/relevancy scores cannot address.

I work on healthcare RAG applications (biomedical question answering, clinical decision support) and this metric suite is directly relevant to that domain. A few observations from that perspective:

**On the Carefulness metric:**
In clinical settings, a RAG system that confidently answers an unanswerable question is significantly more dangerous than one that abstains. The "Carefulness" metric captures this exactly. I would suggest adding a healthcare-specific test case to the test suite, e.g.:

```python
# Example: system should abstain when context is insufficient
test_case = {
    "question": "What is the correct insulin dosage for this patient?",
    "context": "The patient has Type 2 diabetes.",
    "answer": "I cannot determine the correct dosage from the available information.",
    "ground_truth": None
}
# Expected: high Carefulness score (correct abstention)
```

**On the Risk metric:**
The EMNLP 2024 paper framing (counterfactual prompting) is solid. One consideration: in multilingual RAG deployments (English and Uzbek), Risk tends to be higher for low-resource language questions because the model has less calibration data. Worth noting in the documentation as a known limitation.

**Overall:**
This is review-ready from a conceptual standpoint. Happy to help with additional test cases for healthcare or multilingual scenarios if useful for getting this merged.

cc @shahules786

PR #2294, opened 2025-09-19
Fix/asyncio

### LINKED ISSUE #1819: [Feautre Reuqest] [Bug] Async + RAGAS
**Describe the Feature**
As far as I can tell, it seems like there is a lack of async functions in ragas, which leads to a lot of nesting async issues (since sync functions are calling async functions somewhere in the stack). Furthermore, `nest_asyncio.apply()` is automatically called in the import stack, which breaks a lot of applications

```
File "/usr/local/lib/python3.10/site-packages/ragas/executor.py", line 15, in <module>
2025-01-07 23:06:08     nest_asyncio.apply()
```

It would be awesome to have async functions like `aevaluate` instead of `evaluate` -- the approach I might take is make everything async first, and then get/create an event loop and run it there for the sync versions to avoid code duplication.

At the very least, removing the automatic `nest_asyncio.apply()` might be a good idea?

Async code is the backbone of production systems -- without it, ragas will block the event loop of the server

**Additional context**
I THINK the workaround here for now is doing something like

```python
def run_ragas(...):
  from app.evaluation.ragas import generate_dev_set, generate_dev_set_csv
  # other ragas imports

# Then, later on in your code
result = await asyncio.to_thread(run_ragas(...))
```

Which delays the ragas imports and runs it behind an isolated thread, which should be safe. But obviously not ideal

anistark:

- Fixes #1819 

## Changes Made

- Added `aevaluate()` function
- Added `allow_nest_asyncio` parameter in `evaluate()` function
- Added tests
- Added docs
- Updated rest according to the asyncio param usage.

### How to Test


docs/references/evaluate.md
```diff
@@ -1 +1,60 @@
...
+## evaluate()
+
 ::: ragas.evaluation.evaluate
+
+## aevaluate()
+
+::: ragas.evaluation.aevaluate
+
+## Async Usage
+
```
jjmachan, line 12: I think it's better to have it separately listed? 

Image

anistark: Can add separate page for `aevaluate`

jjmachan: yes please do

PR #2302, opened 2025-09-23
fix: improve metric decorators with better validation and error handling

jjmachan:
## Summary
This PR enhances the metric decorator system (`discrete_metric`, `numeric_metric`, `ranking_metric`) with improved input validation, better error messages, and support for both plain value returns and `MetricResult` objects.

### 1. **Automatic MetricResult Wrapping** 
- Metric functions can now return plain values (strings, floats, lists) which are automatically wrapped in `MetricResult` objects
- Functions can still return `MetricResult` objects directly for cases where custom reasons are needed

### 2. **Enhanced Input Validation with Pydantic**
- Added Pydantic-based validation for metric function parameters
- Provides clear, user-friendly error messages for:
  - Type mismatches
  - Missing required parameters
  - Unknown parameters (warnings)
  - Positional arguments (with helpful correction suggestions)

### 3. **Improved Async/Sync Handling**
- Simplified async execution using existing `ragas.async_utils.run`
- Proper detection and handling of event loops
- Both sync and async functions work seamlessly

### 4. **Direct Callable Support**
- Decorated metrics can be called directly like the original function
- `metric()` returns the raw function result
- `metric.score()` returns a `MetricResult` object with validation

### 5. **Better MetricResult Representation**
- Improved `__repr__` to show both value and reason when present
- Cleaner string representation for debugging

## Testing
- Comprehensive test suite added (`test_metric_decorators.py`) covering:
  - Plain value returns vs MetricResult returns
  - Sync and async functions
  - Input validation edge cases
  - Direct callable functionality
  - All three metric types (discrete, numeric, ranking)

## Breaking Changes
None - the changes are backward compatible. Existing code using MetricResult returns will continue to work.

## Benefits
- **Better Developer Experience**: Clear error messages guide users to correct usage
- **More Flexible**: Functions can return simple values without wrapping in MetricResult
- **Type Safety**: Pydantic validation ensures type correctness at runtime
- **Cleaner Code**: Users can write simpler metric functions when custom reasons aren't needed

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/ragas/metrics/decorator.py
+++ b/src/ragas/metrics/decorator.py
@@ -1,6 +1,12 @@
 """decorator factory for creating custom metrics"""
 
-__all__ = ["create_metric_decorator"]
+__all__ = [
+    "create_metric_decorator",
+    "BaseMetricProtocol",
+    "DiscreteMetricProtocol",
+    "NumericMetricProtocol",
+    "RankingMetricProtocol",
+]
 
 import asyncio
 import inspect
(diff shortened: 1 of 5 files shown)


src/ragas/metrics/decorator.py
```diff
@@ -90,47 +87,157 @@ def _validate_result_value(self, result_value):
...
-                            func_args.append(self.prompt)
+                    field_definitions = {}
 
-                        if is_async:
-                            # Async function implementation
-                            if is_async_execution:
-                                # In async context, await the function directly
-                                result = func(*func_args, **func_kwargs)
+                    for name, param in sig.parameters.items():
+                        # Get type hint, default to str if no hint available
```
anistark, line 186: To handle the backwards compatible behaviour where `llm` and `prompt` are handled separately from user-provided parameters, can add something like this:

```py
if name in ['llm', 'prompt']:
    continue
```

PR #2334, opened 2025-10-02
fix: typing

anistark:

- Fixes typing


src/ragas/backends/utils.py
```diff
@@ -1,17 +1,21 @@
...
-def create_nano_id(size=12):
+def create_nano_id(size: int = 12) -> str:
     """Create a short, URL-safe unique identifier."""
     # Define characters to use (alphanumeric)
     alphabet = string.ascii_letters + string.digits
 
     # Generate UUID and convert to int
-    uuid_int = uuid.uuid4().int
+    uuid_obj = uuid.uuid4()
+    uuid_int = t.cast(int, uuid_obj.int)
```
jjmachan: can we join these 2 functions as well?

jjmachan (approved):
just one suggestion and we should be good

PR #2335, opened 2025-10-03
Metrics migration, migrate rouge + answer relevance

rhlbhatnagar:


src/ragas/metrics/v2/base.py
```diff
@@ -0,0 +1,181 @@
...
+
+        return None
+
+    def _validate_llm(self):
+        """Validate that a modern InstructorLLM is provided."""
+        llm = getattr(self, "llm", None)
+        if llm is None:
+            raise ValueError(f"{self.__class__.__name__} requires an llm parameter")
+
+        if type(llm).__name__ != "InstructorLLM":
```
anistark: Better to not use string based checking.

src/ragas/prompt/metrics/answer_relevance.py
```diff
@@ -0,0 +1,45 @@
...
+}}
+Output: {{
+    "question": "What was the groundbreaking feature of the smartphone invented in 2023?",
+    "noncommittal": 1
+}}
+-----------------------------
+
+Now perform the same with the following input
+input: {{
+    "response": "{response}"
```
anistark: If response contains `"`, it breaks JSON

src/ragas/metrics/v2/_answer_relevancy.py
```diff
@@ -0,0 +1,135 @@
...
+    """
+
+    name: str = "answer_relevancy_v2"
+    llm: t.Optional["InstructorBaseRagasLLM"] = field(default=None)
+    embeddings: t.Optional["BaseRagasEmbedding"] = field(default=None)
+    strictness: int = 3
+
+    def __post_init__(self):
+        """Validate that required components are provided."""
+        if self.llm is None:
```
anistark: We shouldn't do manual validation. Better to use Pydantic validators or something similar.

src/ragas/metrics/v2/_answer_relevancy.py
```diff
@@ -0,0 +1,135 @@
+"""Answer Relevancy metric v2 - Class-based implementation with modern components."""
+
+import typing as t
+from dataclasses import dataclass, field
+
+import numpy as np
+from pydantic import BaseModel
+
+from ragas.metrics.result import MetricResult
+from ragas.metrics.v2.base import V2BaseMetric
```
NirantK: @jjmachan do we want this v2 in filename?

src/ragas/prompt/metrics/answer_relevance.py
NirantK: Yeah, let's think through this.

src/ragas/prompt/metrics/answer_relevance.py
```diff
@@ -0,0 +1,45 @@
...
+}}
+
+Example 2
+Input: {{
+    "response": "I don't know about the  groundbreaking feature of the smartphone invented in 2023 as am unaware of information beyond 2022. "
+}}
+Output: {{
+    "question": "What was the groundbreaking feature of the smartphone invented in 2023?",
+    "noncommittal": 1
+}}
```
NirantK:
```suggestion
```
We don't need this anymore with Instructor, it does this for us

src/ragas/metrics/v2/_answer_relevancy.py
jjmachan: no - it already has a module so it's fine

src/ragas/metrics/v2/_answer_relevancy.py
```diff
@@ -0,0 +1,135 @@
...
+        name: The metric name
+        llm: Modern instructor-based LLM for question generation
+        embeddings: Modern embeddings model with embed_text() and embed_texts() methods
+        strictness: Number of questions to generate per answer (3-5 recommended)
+        allowed_values: Score range (0.0 to 1.0)
+    """
+
+    name: str = "answer_relevancy_v2"
+    llm: t.Optional["InstructorBaseRagasLLM"] = field(default=None)
+    embeddings: t.Optional["BaseRagasEmbedding"] = field(default=None)
```
jjmachan: these should be required

src/ragas/metrics/v2/_answer_relevancy.py
```diff
@@ -0,0 +1,135 @@
...
+            raise TypeError(
+                "AnswerRelevancy.__init__() missing required argument: 'llm'"
+            )
+        if self.embeddings is None:
+            raise TypeError(
+                "AnswerRelevancy.__init__() missing required argument: 'embeddings'"
+            )
+
+        # Call parent validation
+        super().__post_init__()
```
jjmachan: if the args are made required then we don't need this check - but something similar should exist in the base class as a passive check

src/ragas/metrics/v2/base.py
```diff
@@ -0,0 +1,181 @@
...
+from dataclasses import dataclass
+
+from ragas.metrics.result import MetricResult
+
+if t.TYPE_CHECKING:
+    pass
+
+
+@dataclass
+class V2BaseMetric(ABC):
```
jjmachan: inherit from the baseclass in metrics.base

src/ragas/metrics/v2/base.py
jjmachan: +1 do change that by doing import checks

src/ragas/metrics/collections/base.py
```diff
@@ -0,0 +1,181 @@
...
+                f"Use: instructor_llm_factory('openai', model='gpt-4o-mini', client=openai_client)"
+            )
+
+    def _validate_embeddings(self):
+        """Validate that modern embeddings are provided."""
+        embeddings = getattr(self, "embeddings", None)
+        if embeddings is None:
+            raise ValueError(
+                f"{self.__class__.__name__} requires an embeddings parameter"
+            )
```
jjmachan: this check is not required

src/ragas/metrics/collections/base.py
```diff
@@ -0,0 +1,181 @@
...
+        if not (min_val <= result_value <= max_val):
+            return f"Metric {self.name} returned {result_value} but expected value in range {self.allowed_values}"
+
+        return None
+
+    def _validate_llm(self):
+        """Validate that a modern InstructorLLM is provided."""
+        llm = getattr(self, "llm", None)
+        if llm is None:
+            raise ValueError(f"{self.__class__.__name__} requires an llm parameter")
```
jjmachan: this check is not required since we are passing we decided to define `llm` and `embedding` arguments as default arguments as required by the metric we are implementing

src/ragas/metrics/v2/example_metric.py
```diff
@@ -0,0 +1,70 @@
...
+    - Async-first design
+
+    Usage:
+        >>> metric = ExampleV2Metric(llm=modern_llm, embeddings=modern_embeddings)
+        >>> result = await metric.ascore(user_input="test", response="test")
+    """
+
+    name: str = "example_v2_metric"
+    llm: t.Optional["InstructorBaseRagasLLM"] = field(default=None)
+    embeddings: t.Optional["BaseRagasEmbedding"] = field(default=None)
```
jjmachan: should remove this pattern from example - `llm` and `embedding` if required by the metric should be marked required and not optional

src/ragas/metrics/v2/example_metric.py
```diff
@@ -0,0 +1,70 @@
...
+            raise TypeError(
+                "ExampleV2Metric.__init__() missing required argument: 'llm'"
+            )
+        if self.embeddings is None:
+            raise TypeError(
+                "ExampleV2Metric.__init__() missing required argument: 'embeddings'"
+            )
+
+        # Call parent validation
+        super().__post_init__()
```
jjmachan: no need for this

src/ragas/prompt/metrics/answer_relevance.py
jjmachan, line 1: move the prompt to the file that we create metric itself

src/ragas/prompt/metrics/answer_relevance.py
rhlbhatnagar: Just did a to string on the pydantic prompt and copied the entire value as is. I think we can use the existing prompt as is at least as part of the migration.

src/ragas/prompt/metrics/answer_relevance.py
rhlbhatnagar: Actually let's revist this as part of the promptMixin we're creating, great catch though.

src/ragas/metrics/collections/base.py
```diff
@@ -43,53 +44,16 @@ def __post_init__(self):
...
-        async_tasks = []
-        for input_dict in inputs:
-            # Process input asynchronously
-            async_tasks.append(self.ascore(**input_dict))
-
-        # Run all tasks concurrently and return results
-        return await asyncio.gather(*async_tasks)
+        return MetricResult(
+            value=0.0, reason="Base metric placeholder - override ascore() in subclass"
+        )
```
jjmachan: it should pass the results right?

src/ragas/prompt/metrics/answer_relevance.py
```diff
@@ -40,6 +40,6 @@ def answer_relevancy_prompt(response: str) -> str:
 
 Now perform the same with the following input
 input: {{
-    "response": "{response}"
+    "response": {safe_response}
 }}
```
jjmachan, line 44: are we still sticking with JSON format?

src/ragas/metrics/collections/example_metric.py
```diff
@@ -0,0 +1,47 @@
...
+    Usage:
+        >>> metric = ExampleMetric()
+        >>> result = await metric.ascore(user_input="test", response="test")
+    """
+
+    name: str = "example_metric"
+
+    async def _ascore_impl(self, user_input: str, response: str) -> MetricResult:
+        """
+        Calculate example score asynchronously.
```
jjmachan: this needs to be updated?

PR #2350, opened 2025-10-09
Simplified Prompt Mixin

rhlbhatnagar:


src/ragas/prompt/simple_mixin.py
```diff
@@ -0,0 +1,161 @@
...
+
+        if prompt_name not in current_prompts:
+            available_names = list(current_prompts.keys())
+            raise ValueError(
+                f"Prompt '{prompt_name}' not found. Available prompts: {available_names}"
+            )
+
+        current_prompt = current_prompts[prompt_name]
+        modified_prompt = current_prompt.copy_with_modifications(
+            instruction=instruction, examples=examples
```
anistark, line 147: Is there a case where both are `None`? Since both are optional.
What's supposed to happen in that case?

src/ragas/metrics/collections/_answer_relevancy_v2.py
anistark, line 1: Why `_v2` ?

anistark:
Closing this in favor of #2428

PR #2354, opened 2025-10-10
feat: add bypass_n option to LangchainLLMWrapper for n-completion control

SimFG:
When using some openai-compatible thinking models at that time, it was found that centain errors would occur.

2025-10-10 11:22:45,625 - ragas.executor - ERROR - Exception raised in Job[3]: BadRequestError(Error code: 400 - {'error': {'code': 'InvalidParameter', 'message': 'Reasoning model does not support n > 1, logit_bias, logprobs, top_logprobs Request id: xxx', 'param': '', 'type': 'BadRequest'}})

### DIFF
--- a/src/ragas/llms/base.py
+++ b/src/ragas/llms/base.py
@@ -159,7 +159,7 @@ class LangchainLLMWrapper(BaseRagasLLM):
         self.is_finished_parser = is_finished_parser
         # Certain LLMs (e.g., OpenAI o1 series) do not support temperature
         self.bypass_temperature = bypass_temperature
-        # Certain Reason LLMs do not support n
+        # Certain reasoning LLMs (e.g., OpenAI o1 series) do not support n parameter for
         self.bypass_n = bypass_n
 
     def is_finished(self, response: LLMResult) -> bool:
--- a/tests/unit/llms/test_llm.py
+++ b/tests/unit/llms/test_llm.py
@@ -9,9 +9,6 @@ from langchain_core.prompt_values import PromptValue
 
 from ragas.llms.base import BaseRagasLLM, LangchainLLMWrapper
 
-if t.TYPE_CHECKING:
-    from langchain_core.prompt_values import PromptValue
-
 
 class FakeTestLLM(BaseRagasLLM):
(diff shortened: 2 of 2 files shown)


src/ragas/llms/base.py
```diff
@@ -225,7 +228,7 @@ def generate_text(
             old_temperature = self.langchain_llm.temperature  # type: ignore
             self.langchain_llm.temperature = temperature  # type: ignore
 
-        if is_multiple_completion_supported(self.langchain_llm):
+        if is_multiple_completion_supported(self.langchain_llm) && and not self.bypass_n:
```
anistark:
```suggestion
        if is_multiple_completion_supported(self.langchain_llm) and not self.bypass_n:
```

SimFG:
@anistark Thanks a lot for your time, I have added some relative unit test cases.

src/ragas/llms/base.py
```diff
@@ -158,6 +159,8 @@ def __init__(
         self.is_finished_parser = is_finished_parser
         # Certain LLMs (e.g., OpenAI o1 series) do not support temperature
         self.bypass_temperature = bypass_temperature
+        # Certain Reason LLMs do not support n
```
anistark:
```suggestion
        # Certain reasoning LLMs (e.g., OpenAI o1 series) do not support n parameter for
   multiple completions
```

tests/unit/llms/test_llm.py
```diff
@@ -1,10 +1,13 @@
...
 
+import pytest
 from langchain_core.outputs import Generation, LLMResult
+from langchain_core.prompt_values import PromptValue
 
-from ragas.llms.base import BaseRagasLLM
+from ragas.llms.base import BaseRagasLLM, LangchainLLMWrapper
 
 if t.TYPE_CHECKING:
     from langchain_core.prompt_values import PromptValue
```
anistark: I guess this is redundant now that you've added in line 8 above. Or was that not needed?

SimFG:
I have updated the pull request, thanks your careful review.

PR #2376, opened 2025-10-20
Migrate SummaryScore

rhlbhatnagar:


src/ragas/metrics/collections/_summary_score.py
```diff
@@ -0,0 +1,189 @@
...
+
+        Returns:
+            MetricResult with summary score (0.0-1.0)
+        """
+        # Step 1: Combine contexts and extract keyphrases
+        text = "\n".join(reference_contexts)
+        keyphrases = await self._extract_keyphrases(text)
+
+        if not keyphrases:
+            # If no keyphrases extracted, return perfect score
```
anistark: Shouldn't we raise error here instead of perfect score?
Or at least log it?

src/ragas/metrics/collections/_summary_score.py
```diff
@@ -0,0 +1,189 @@
...
+            return MetricResult(value=1.0)
+
+        # Step 2: Generate questions from keyphrases
+        questions = await self._generate_questions(text, keyphrases)
+
+        if not questions:
+            # If no questions generated, return perfect score
```
anistark: Same as above, to raise or at least log it.

src/ragas/metrics/collections/_summary_score.py
```diff
@@ -0,0 +1,189 @@
...
+
+        # Validate coefficient
+        if not (0.0 <= coeff <= 1.0):
+            raise ValueError(f"Coefficient must be between 0.0 and 1.0, got {coeff}")
+
+        # Call super() for validation (without passing llm in kwargs)
+        super().__init__(name=name, **kwargs)
+
+    async def ascore(
+        self, reference_contexts: List[str], response: str
```
anistark, line 120: Should add validation here to check if `reference_contexts` or `response` is empty.

PR #2380, opened 2025-10-23
docs: quickstart guide with interactive LLM and project structure

anistark:
### Documentation (docs/getstarted/evals.md):
- Added `quickstart` cmd to follow through guide with an example project.
- Updated "Custom Evaluation with LLMs" to reference DiscreteMetric from
generated code
- Replaced static examples with modern `llm_factory` API
- Changed "Using Pre-Built Metrics" to AspectCritic with modern async/await
syntax
- Updated "Evaluating on a Dataset" to use ragas.Dataset API

### Build Configuration (mkdocs.yml):
- Made social plugin conditional: `enabled: !ENV [MKDOCS_CI, true]`

### Makefile:
- Added explicit `MKDOCS_CI=false` to serve-docs target. This avoids social plugin error in macos if in case `cairosvg` is not found.


sanjeed5:
Thoughts on quickstart:

[experiments_quickstart](https://ragas--2380.org.readthedocs.build/en/2380/getstarted/experiments_quickstart/) this feels a lot more simpler and easier to understand than proposed [quickstart](https://ragas--2380.org.readthedocs.build/en/2380/getstarted/quickstart) according to me. (and Claude and GPT) 
Tutorials - learn by doing according to diataxis too. 

Also regarding the other rag_eval guide, we already have an experiment first rag evals guide, so we can remove the older metrics based ones. This is done in #2394 

Other stuff looks great! That makefile fix is great!

PR #2403, opened 2025-11-05
Response Groundedness

rhlbhatnagar:


src/ragas/prompt/metrics/response_groundedness.py
```diff
@@ -0,0 +1,62 @@
...
+I will provide a rating of 0, 1, or 2, without any additional information.
+
+---
+**Context:**
+[{context}]
+
+**Assertion:**
+[{response}]
+
+Do not explain."""
```
anistark: Did we miss this part after: `"Based on the provided context and response, the Groundedness score is:"` ?

PR #2428, opened 2025-11-17
Prompt class + implementation in Context Precision

rhlbhatnagar:


src/ragas/prompt/metrics/base_prompt.py
```diff
@@ -0,0 +1,152 @@
...
+                else:
+                    # Fallback to original
+                    translated_examples.append((input_ex, output_ex))
+
+            except Exception:
+                # Fallback to original example if translation fails
+                translated_examples.append((input_ex, output_ex))
+
+        new_prompt.examples = translated_examples
+        return new_prompt
```
rhlbhatnagar, line 152: Do we return a new prompt? Or should we replace the existing one.

src/ragas/metrics/collections/context_precision/util.py
```diff
@@ -1,104 +1,67 @@
...
-    # This variant uses "response" as the "answer" parameter in the prompt
-    return context_precision_prompt(question, context, response)
+"""Context Precision prompt - Class-based implementation with structured input/output."""
+
+from pydantic import BaseModel, Field
+
+from .base_prompt import BasePrompt
+
+
+class ContextPrecisionInput(BaseModel):
```
rhlbhatnagar, line 8: Now the input + output classes also need to be defined in the prompt file to remove circular dependencies, should we just turn the metric into a folder and just keep propmt + input + output classes there.

eg 

metrics/collections/context_precision
-- prompt.py
-- classes.py
-- metric.py

?

anistark: Should we discuss on the notion doc first, since we're all there anyway? 
Settle down on the architecture and structure and then move to PR?

anistark (approved):
LGTM.

@rhlbhatnagar The ci is failing. Can you please fix that and we can merge it.

anistark:
Merged in #2433.

PR #2446, opened 2025-11-24
feat: dual adapter support (Instructor + LiteLLM)

### LINKED ISSUE #2416: Gemini supported llm and embedding docs
**Describe the Feature**I
I have access to Gemini for llm_judge and embedding model I was using mixedbread-ai/mxbai-embed-large-v1
I am facing issue in 3 metrics - faithfulness, response_relevancy, context_recall
I am attaching the ss for your reference:

Image

As our organization has access to Gemini key so I want ragas to be compatible with gemini models

**Additional context**
Add any other context about the feature you want to share with us.

anistark:
Add support for multiple structured output adapters (Instructor, LiteLLM) with automatic selection and fallback mechanisms. This enables seamless support for providers like Google Gemini that don't work well with Instructor.

Also fixes #2416

### DIFF
--- a/docs/howtos/integrations/gemini.md
+++ b/docs/howtos/integrations/gemini.md
@@ -195,18 +195,7 @@ if not os.environ.get("GOOGLE_API_KEY"):
 
 ### Rate Limits
 
-Gemini has rate limits. For production:
-
-```python
-from ragas.run_config import RunConfig
-
-# Configure retries and timeouts
-config = RunConfig(
-    timeout=60,
-    max_retries=3,
-)
-llm.set_run_config(config)
-```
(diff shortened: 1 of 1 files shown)


docs/howtos/integrations/gemini.md
```diff
@@ -0,0 +1,288 @@
...
+
+```python
+from ragas.run_config import RunConfig
+
+# Configure retries and timeouts
+config = RunConfig(
+    timeout=60,
+    max_retries=3,
+)
+llm.set_run_config(config)
```
jjmachan: are we still using RunConfig?

anistark: Not in the new one. Will update the docs.

PR #2474, opened 2025-12-07
New input parameter for the TestsetGenerator Class : LLM_CONTEXT 

narabi:

- Fixes #[issue_number]
- OR describe the issue: What problem does this solve? How can it be replicated?

During experimentation across different dataset types (CSV, PDF, etc.), I observed that the TestsetGenerator did not consistently generate queries that reflect what users might want to ask when interacting with the system. Although we could define various personas, the description provided there alone did not resolve the issue.

To address this, I introduced a new parameter in the TestsetGenerator called llm_context. 
This parameter allows us to guide the AI to generate more targeted question types based on the expected user. 
With llm_context, the LLM can be steered to produce specific styles of queries—for example, yes/no questions, use-case questions, calculation/application questions, or any pattern required for evaluation.

#### This improvement results in higher-quality evaluation test datasets.

In the test example provided we will see : 

 - Without llm_context: The TestsetGenerator creates generic, basic questions about the document content - like "What is a Pell Grant?" or simple factual questions.

 - With llm_context: It uses the additional context and generates specific, calculation-based questions that require applying formulas and working with actual numbers from the document - like "A student's SAI is 1,004 and their Pell COA is $6,493. What would be their Scheduled Award?"

## Changes Made

Added `llm_context` parameter to ragas `TestsetGenerator` and subsequently, changed these 7 files:
- `src/ragas/testset/synthesizers/generate.py`
- `src/ragas/testset/synthesizers/base.py` 
- `src/ragas/testset/synthesizers/__init__.py`
- `src/ragas/testset/synthesizers/single_hop/prompts.py`
- `src/ragas/testset/synthesizers/single_hop/base.py`
- `src/ragas/testset/synthesizers/multi_hop/prompts.py`
- `src/ragas/testset/synthesizers/multi_hop/base.py`

I have tested multiple scenarios and these were the required changes. The TestsetGenerator worked fine end to end. 

### How to Test
- [ x] Manual testing steps:

Test Files are available under the tests/unit
TEST_SDG_LLM_CONTEXT.py
TEST_SDG_LLM_CONTEXT.md
TEST_SDG_LLM_CONTEXT/The_Federal_Pell_Grant_Program.pdf

  1.  Create `.env` file with `OPENAI_API_KEY=your_key_here`
  2. Run PYTHONPATH=/Users/[YOUR username]/Documents/ragas/src .venv/bin/python tests/unit/TEST_SDG_LLM_CONTEXT.py
  3. If errored, it can be due to new installations that would need to be added to the  .toml file. The error should guide. 
  4. After each TEST_SDG_LLM_CONTEXT.py run, the csv files below get re-created
- `SDG_without_LLM_CONTEXT.csv` - Generic loan questions
- `SDG_with_LLM_CONTEXT.csv` - Context-guided loan questions
5. The test generates 4 questions each round to compare output quality between with and without `LLM_CONTEXT' parameter.

## References

- Related issues: 
- Documentation: 
- External references: 

---

### DIFF
--- a/tests/unit/test_llm_context.py
+++ b/tests/unit/test_llm_context.py
@@ -1,17 +1,18 @@
 #!/usr/bin/env python3
 """Test llm_context feature with calculation-based Pell Grant questions"""
-import os
+
 from dotenv import load_dotenv
 from langchain_core.documents import Document
 from langchain_openai import ChatOpenAI, OpenAIEmbeddings
 
+from ragas.embeddings import LangchainEmbeddingsWrapper
+from ragas.llms import LangchainLLMWrapper
+from ragas.run_config import RunConfig
 from ragas.testset import TestsetGenerator
 from ragas.testset.persona import Persona
-from ragas.llms import LangchainLLMWrapper
-from ragas.embeddings import LangchainEmbeddingsWrapper
-from ragas.run_config import RunConfig
 
 load_dotenv()
+
 
 def main():
     # Create documents from hardcoded text (no PDF needed!)
@@ -68,12 +69,15 @@ def main():
     
     Example 6: If calculation results in $3,456.78, the disbursement is $3,456.
     """
-    
+
     # Use single document to minimize async complexity
     docs = [
-        Document(page_content=pell_grant_text, metadata={"source": "pell_grant_doc", "page": 1})
+        Document(
+            page_content=pell_grant_text,
+            metadata={"source": "pell_grant_doc", "page": 1},
+        )
     ]
-    
+
     print(f"Created {len(docs)} document from Pell Grant text")
 
     # Setup models
@@ -84,7 +88,7 @@ def main():
     personas = [
         Persona(
             name="Financial Aid Officer",
-            role_description="A financial aid officer who needs to calculate Pell Grant awards accurately using specific formulas and numerical examples"
+            role_description="A financial aid officer who needs to calculate Pell Grant awards accurately using specific formulas and numerical examples",
         )
     ]
 
@@ -124,11 +128,15 @@ Answers should show the calculation steps and final numerical result.
         llm=generator_llm,
         embedding_model=generator_embeddings,
         persona_list=personas,
-        llm_context=llm_context  # 🆕 WITH CONTEXT for calculation questions!
+        llm_context=llm_context,  # 🆕 WITH CONTEXT for calculation questions!
     )
 
     # Minimal transforms (workaround for ragas headline bug)
-    from ragas.testset.transforms import EmbeddingExtractor, CosineSimilarityBuilder, OverlapScoreBuilder
+    from ragas.testset.transforms import (
+        CosineSimilarityBuilder,
+        EmbeddingExtractor,
+        OverlapScoreBuilder,
+    )
     from ragas.testset.transforms.extractors.llm_based import NERExtractor
 
     minimal_transforms = [
@@ -152,14 +160,14 @@ Answers should show the calculation steps and final numerical result.
         docs[:num_docs],
         testset_size=1,  # Generate 1 calculation-based question (minimal to avoid async issues)
         transforms=minimal_transforms,
-        run_config=run_config
+        run_config=run_config,
     )
 
     print(f"\n✅ Generated {len(dataset_with_context)} queries WITH llm_context!")
 
     # Convert to dataframe
     df_with_context = dataset_with_context.to_pandas()
-    
+
     # Display samples
     print("\n" + "=" * 80)
     print("📊 QUESTIONS WITH LLM CONTEXT (calculation-based):")
@@ -171,7 +179,7 @@ Answers should show the calculation steps and final numerical result.
         print(f"Question: {eval_sample.user_input}")
         print(f"Answer: {eval_sample.reference}")
         print("-" * 80)
-    
+
     print("\n📊 DataFrame Columns:", df_with_context.columns.tolist())
     print(f"📊 DataFrame Shape: {df_with_context.shape}")
 
@@ -183,7 +191,7 @@ Answers should show the calculation steps and final numerical result.
     generator_no_context = TestsetGenerator(
         llm=generator_llm,
         embedding_model=generator_embeddings,
-        persona_list=personas
+        persona_list=personas,
         # NO llm_context!
     )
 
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -26,6 +26,9 @@ dependencies = [
     "langchain-core",
     "langchain-community",
     "langchain_openai",
+    
+    # Embeddings
+    "cohere",
```
anistark: Dont suppose this dependency is serving any purpose. Better to remove it.

SDG_with_LLM_CONTEXT.csv
anistark, line 1: These output files don't need to be committed.

src/ragas/testset/synthesizers/generate.py
```diff
@@ -192,7 +200,7 @@ def generate_with_langchain_docs(
         kg = KnowledgeGraph(nodes=nodes)
 
         # apply transforms and update the knowledge graph
-        apply_transforms(kg, transforms)
+        apply_transforms(kg, transforms, run_config=run_config or RunConfig())
```
anistark, line 203: This is irrelevant for this PR. Better to take it up separately.

narabi:
Thank you for the feedback.

A few follow up questions:
- Should I remove all these, or just the cohere line?

     "langchain-core",
     "langchain-community",
     "langchain_openai",
+
+    # Embeddings
+    "cohere”,


- Should I submit another PR for the apply_transforms(kg, transforms,
run_config=run_config or RunConfig()) line ?


Thanks,

Nooshin

anistark:
> - Should I remove all these, or just the cohere line?

I can see just the `cohere` dependency added in this [PR diff](https://github.com/vibrantlabsai/ragas/pull/2474/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711). But, it's not used anywhere so it's kinda dead dependency, right? 

> - Should I submit another PR for the apply_transforms(kg, transforms, run_config=run_config or RunConfig()) line ? 

Yes, you can.

tests/unit/TEST_SDG_LLM_CONTEXT.py
```diff
@@ -0,0 +1,175 @@
...
+from ragas.llms import LangchainLLMWrapper
+from ragas.embeddings import LangchainEmbeddingsWrapper
+from ragas.run_config import RunConfig
+
+load_dotenv()
+
+def main():
+    # Load PDFs
+    path = "TESTS/UNIT/TEST_SDG_LLM_CONTEXT/"
+    loader = DirectoryLoader(path, glob="*.pdf", loader_cls=PyPDFLoader ) # PyMuPDFLoader
```
anistark: It's good to test from a pdf file but this doesn't need to go under unit tests. Unit tests need to be more compact. You can simply write a test which shows passing `llm_context` from a string in same file. Another to not passing. Another to passing but not really a context, and so on.
Also, the test file name doesn't have to be CAPS. Please check out how the rest of the files are written and follow similar naming convention.

pyproject.toml
```diff
@@ -26,6 +26,7 @@ dependencies = [
     "langchain-core",
     "langchain-community",
     "langchain_openai",
+
```
anistark: If there's no change, better to not add new lines or white spaces.

narabi: Thanks for the feedback. should I make any changes for this?

tests/unit/TEST_SDG_LLM_CONTEXT.py
narabi: Thanks for the feedback. 

- I will rename the files. 

- I have a clarifying question on the first part: 
Should I move TEST_SDG_LLM_CONTEXT folder which has the pdf file, outside the unit test folder? where would be best to put it?
If that's okay, I'd like to stick with this example file for my test use case.

- Also, should I remove the TEST_SDG_LLM_CONTEXT.md ?

anistark: Yes, please move it. For unit test, you can write some of that content in the test file directly. It just needs to test the functionality. If we need to test end-2-end, we can add a diff test in e2e.

For you personal testing, you can keep it wherever you want. Just don't commit it, cause then it's on the repo. Personally we use `_experiments` folder which is gitignored for temp tests that uses files that're not needed to be committed.

As a practice, it's better to not add any file that has no meaning for anybody else but you. And we keep the project lean as possible.

pyproject.toml
anistark: If you can, just remove this new line with your next commit.

tests/unit/TEST_SDG_LLM_CONTEXT.py
narabi: There are two reasons that the PDF example could be beneficial:
	1.	If I had used an easier example, I wouldn’t have discovered the issue that led to #2480. I was running into failures during concurrent API calls, which prompted me to dig deeper and realize that the run_config was missing.
	2.	This example also has educational value for people trying to learn Synthetic Data Generation more deeply.

I can move the PDF file outside of the unit test folder. I can proceed and create a data folder (if that doesn't exist) and placing the PDF there; this could be useful for future use cases and contributors as well.

Additionally, the SDG example is a specific scenario, and generating the real output as a CSV as the result of the unit test could be ideal. Those CSV files could also live in the data folder.

anistark: Unit tests are more meant for "units", so to say. What you've is an e2e test. So, if you want to keep this, better to move it to e2e.
Secondly, if you want to import a pdf, better to keep it at a remote path and import from there. If at all needed. Or have it in your local. Can mention something like a warning... "For best results, use a real pdf file like so and so..."

The markdown file also isn't needed. If you need to make a note of something, better to keep it in docstrings or comments in the test file.

If you do intend to move this to e2e, would prefer if you write a unit test file as well which checks individual functions in various cases. It doesn't have to check qualitative aspects. Only functional aspects.

narabi: Got it. I went with the first suggested solution. removed all extra files and kept it under the tests/unit folder with a simpler test usecase (no pdf is required and no csv output gets created)

narabi: I'm not sure why the [CI / code_quality_check] has failed.

anistark: Run `make format` once and commit. 

To test locally, you can run `make run-ci`

narabi: Got it. Done and committed.
It passed the checks

anistark (approved):

PR #2493, opened 2025-12-09
feat: Enable PDF export for documentation

### LINKED ISSUE #2251: Printing documentation to PDF
**Describe the Feature**
An option to print to PDF would help go over all the features of the package.

Moving through pages makes it hard to return to the right place and understand what was already covered.

cgaswin:
- Fixes #2251 

## Changes Made
- Add mkdocs-to-pdf plugin to mkdocs.yml for PDF generation
- Add mkdocs-to-pdf>=0.10.1 to docs dependency group in pyproject.toml
- Configure PDF output to site/pdf/document.pdf
- Exclude community pages from PDF output

### How to Test
- [ ] Manual testing steps:
  1. Install documentation dependencies: `uv pip install -e . --group docs`
  2. Build documentation: `mkdocs build`
  3. Verify PDF file exists: `ls -lh site/pdf/document.pdf`
  4. Verify PDF is valid: `file site/pdf/document.pdf` (should show "PDF document")
  5. Open the generated PDF to verify content is readable and properly formatted
  6. Confirm community pages are excluded from PDF (check PDF table of contents)

Screenshot from 2025-12-10 01-39-42
Screenshot from 2025-12-10 01-35-34

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -161,7 +161,9 @@ Run `make help` to see all targets. Here are the essential commands:
 
 ### Documentation
 - `make build-docs` - Build all documentation
+- `make build-docs-pdf` - Build documentation with PDF export (requires WeasyPrint)
 - `make serve-docs` - Serve documentation locally
+- See `docs/community/pdf_export.md` for PDF export details and limitations
 
 ## Development Workflows
 
--- a/docs/community/pdf_export.md
+++ b/docs/community/pdf_export.md
@@ -1,0 +1,30 @@
+# PDF Export
+
+## Purpose
+The PDF export feature builds the complete Ragas documentation as a single PDF file using MkDocs with the `mkdocs-to-pdf` plugin.
+
+## Usage
+
+The implementation uses two separate MkDocs configurations:
+- `mkdocs.yml` for standard HTML builds (no PDF dependencies required)
+- `mkdocs-pdf.yml` which inherits from the main config and adds the PDF plugin
+
+Build PDF documentation:
+```bash
+make build-docs-pdf
+```
+
+The generated PDF will be available at `site/pdf/document.pdf`.
+
+Build HTML documentation only:
+```bash
+make build-docs
+```
+
+The `make build-docs-pdf` command automatically checks for system dependencies before building.
+
+## Current Limitations
+
+**System Dependencies**: WeasyPrint requires OS-specific system libraries (Pango, Cairo) that must be installed separately. If you encounter issues, refer to the [WeasyPrint setup instructions](https://doc.courtbouillon.org/weasyprint/stable/first_steps.html) and [troubleshooting guide](https://doc.courtbouillon.org/weasyprint/stable/first_steps.html#troubleshooting).
+
+**ReadTheDocs**: PDF generation is not currently enabled in the ReadTheDocs build configuration.


cgaswin:
Hi @anistark 

Updated the PR to follow your suggestion

    - to-pdf is now gated with enabled_if_env: ENABLE_PDF_EXPORT in mkdocs.yml

    - Added build-docs-pdf which runs MKDOCS_CI=false ENABLE_PDF_EXPORT=1 uv run --group docs mkdocs build

Tested locally: make build-docs builds HTML only, make build-docs-pdf also creates site/pdf/document.pdf.

anistark:
@cgaswin It's not working for me.

```sh
WeasyPrint could not import some external libraries. Please carefully follow the installation steps before reporting an issue:
https://doc.courtbouillon.org/weasyprint/stable/first_steps.html#installation
https://doc.courtbouillon.org/weasyprint/stable/first_steps.html#troubleshooting
```

Getting these errors:

```sh
OSError: cannot load library 'gobject-2.0-0': dlopen(gobject-2.0-0, 0x0002): tried: 'gobject-2.0-0' (no such file), '/System/Volumes/Preboot/Cryptexes/OSgobject-2.0-0' (no such file), '/usr/lib/gobject-2.0-0' (no such file, not in dyld cache), 'gobject-2.0-0' (no such file).  Additionally, ctypes.util.find_library() did not manage to locate a library called 'gobject-2.0-0'
```

cgaswin:
Hi @anistark ,
Did you try make build-docs-pdf or make build-docs? Is this the complete log?

I’m not seeing this on my Ubuntu machine , make build-docs-pdf works there with the current setup, but from the traceback and WeasyPrint docs it looks like the “missing library ” issue mentioned in the troubleshoot section of WeasyPrint docs.

Could you check that the WeasyPrint system dependencies are installed on your OS (GTK/GLib, cairo, pango, etc.) and that its troubleshooting steps for missing libraries are applied?

Docs: https://doc.courtbouillon.org/weasyprint/stable/first_steps.html#installation

Troubleshooting: https://doc.courtbouillon.org/weasyprint/stable/first_steps.html#troubleshooting

If you’re on macOS that typically means installing the stack via Homebrew and possibly setting DYLD_FALLBACK_LIBRARY_PATH; 

 I’ll keep digging on my side as well.

anistark:
> Did you try make build-docs-pdf or make build-docs? Is this the complete log?

I ran this:

```sh
make build-docs-pdf
```

> If you’re on macOS that typically means installing the stack via Homebrew and possibly setting DYLD_FALLBACK_LIBRARY_PATH;

I'm using on mac. But any config needed should be added to the makefile itself. Probably a better library that works cross platform. Or we won't know if something would break on windows or some other linux distros as well.

anistark:
Actually now it's breaking `make build-docs` too. You'll need to keep this completely isolated from existing functionality. 

`weasyprint` is the culprit.

cgaswin:
> I'm using on mac. But any config needed should be added to the makefile itself. Probably a better library that works cross platform. Or we won't know if something would break on windows or some other linux distros as well.


Makes sense, I will look into it and maybe change to a better plugin.

>Actually now it's breaking make build-docs too. You'll need to keep this completely isolated from existing functionality.

Shouldnt have happened, I didnt particularly do any changes in build-docs. Let me see why that happened.

cgaswin:
Hi @anistark,

I’ve updated the setup so **PDF export is now fully opt-in**.

- `make build-docs`  
  Uses only the `docs` group, which no longer includes `mkdocs-to-pdf`.  
  This means the regular docs build should no longer run into the WeasyPrint/system-deps issues.

- `make build-docs-pdf`  
  Adds the `docs-pdf` group and sets `ENABLE_PDF_EXPORT=1`, so only this target installs and enables the to-pdf plugin.

This should fix the problems you were seeing with `make build-docs`.  
I’m also looking into alternatives and tweaks around WeasyPrint to reduce the pain further.

cgaswin:
Hi @anistark,

I was checking out a few options for PDF support. Most of them use WeasyPrint, which depends on a bunch of system libraries that may not always be installed. Some others use Playwright/headless browsers/Puppeteer, but those come with their own setup issues too.

I was thinking we could have a Makefile command that checks if WeasyPrint is installed and, if not, just points the user to the install/troubleshooting page. We could run this check when someone does `make build-docs-pdf`.
Does that sound reasonable, or do you think there's a better way to handle this?

I also found this plugin: https://timvink.github.io/mkdocs-print-site-plugin/index.html


It adds a “Download PDF” button in the browser. Not sure if that’s the behaviour we want, and the example on their site didn’t seem to work for me.
Do you think it’s worth looking into?

anistark:
> I was thinking we could have a Makefile command that checks if WeasyPrint is installed and, if not, just points the user to the install/troubleshooting page. We could run this check when someone does `make build-docs-pdf`. Does that sound reasonable, or do you think there's a better way to handle this?

Sounds good. Test it out on all cases. Not-Installed, Installed but didn't work, Worked but crashed in between, worked completely. Any other cases you can think of.

> I also found this plugin: https://timvink.github.io/mkdocs-print-site-plugin/index.html
> 
> It adds a “Download PDF” button in the browser. Not sure if that’s the behaviour we want, and the example on their site didn’t seem to work for me. Do you think it’s worth looking into?

If it didn't work for you on their own site, probably not a good plugin.

cgaswin:
Hey @anistark, I’ve updated the approach slightly based on some testing.

It turns out that `mkdocs` attempts to **import** all plugins listed in the config at startup, even if they are gated by `enabled_if_env`. So, if a user does a minimal install (without PDF deps), the build crashes immediately before checking the environment variable.

To fix this while keeping the "minimal install" promise, I’ve split the configuration:
1.  **`mkdocs.yml`**: Clean, standard HTML build. (Default)
2.  **`mkdocs-pdf.yml`**: Inherits from the main config but adds the PDF plugin.

I’ve also updated the `Makefile` to handle this:
*   `make build-docs`: Runs standard HTML build (fast, no heavy deps).
*   `make build-docs-pdf`: Uses the specific PDF config.
*   **Safety Check**: I added a `check-pdf-deps` target that runs before the PDF build. It tries to render a tiny dummy PDF in memory (using `target=None` so no files are created). If the required system libraries (Pango/Cairo) are missing, this step will fail gracefully with setup instructions, rather than letting the build crash with a stack trace.

**One note on ReadTheDocs:**
Currently, `readthedocs.yml` points to the default config, so the PDF **won't** be generated on the live website. To enable it there, we would need to update `readthedocs.yml` to install the system dependencies (`libcairo`, `libpango`, etc.) and switch the build command to use the PDF config.

Do you think we should enable that for the live site now, or keep the PDF generation as an offline/local feature for now? Let me know if you have any suggestions!

anistark:
> Hey @anistark, I’ve updated the approach slightly based on some testing.

> * `make build-docs`: Runs standard HTML build (fast, no heavy deps).
> * `make build-docs-pdf`: Uses the specific PDF config.
> * **Safety Check**: I added a `check-pdf-deps` target that runs before the PDF build. It tries to render a tiny dummy PDF in memory (using `target=None` so no files are created). If the required system libraries (Pango/Cairo) are missing, this step will fail gracefully with setup instructions, rather than letting the build crash with a stack trace.

Makes sense. 

> Do you think we should enable that for the live site now, or keep the PDF generation as an offline/local feature for now? Let me know if you have any suggestions!

Nope, we don't need on live site at all. This is for folks generating locally.

anistark:
@cgaswin Can you also please document this somewhere in the docs?

Purpose, usage and current limitations.

cgaswin:
Sure @anistark , I can add a new pdf_export.md under docs/community/ to document the purpose, usage, and limitations, and then link to it from CONTRIBUTING.md. Would that work?

anistark:
> Sure @anistark , I can add a new pdf_export.md under docs/community/ to document the purpose, usage, and limitations, and then link to it from CONTRIBUTING.md. Would that work?

That sounds good for now. Thanks

cgaswin:
Hi @anistark 

I’ve opened a new, cleaner PR for this change: [#2522](https://github.com/vibrantlabsai/ragas/pull/2522).
After rebasing this branch on top of the latest main, it picked up a lot of unrelated upstream changes and the diff became quite noisy.
To keep the review focused on the PDF export workflow and docs, #2522 contains only the intended Makefile, MkDocs, and documentation updates.
Happy to close this PR once the new one is reviewed.

anistark:
> Hi @anistark
> 
> I’ve opened a new, cleaner PR for this change: [#2522](https://github.com/vibrantlabsai/ragas/pull/2522). After rebasing this branch on top of the latest main, it picked up a lot of unrelated upstream changes and the diff became quite noisy. To keep the review focused on the PDF export workflow and docs, #2522 contains only the intended Makefile, MkDocs, and documentation updates. Happy to close this PR once the new one is reviewed.

Thanks. Closing this in favor of #2522

PR #2513, opened 2025-12-15
fix: allow unannotated parameters to accept any type

### LINKED ISSUE #2512: text2sql examples runtime error [predicted_result: Input should be a valid string]
[X ] I have checked the [documentation](https://docs.ragas.io/) and related resources and couldn't resolve my bug.

**Describe the bug**
Error when calling `text2sql_experiment.arun`
```
Warning: Task failed with error: 
❌ Type validation errors for execution_accuracy:

   - predicted_result: Input should be a valid string (got:    Balance
0      0.0)
```

Ragas version: 0.0.4
Python version: 3.12

**Code to Reproduce**
```python
import asyncio
import os

import ragas_examples
from ragas_examples.text2sql.evals import text2sql_experiment, load_dataset

# Set your OpenAI API key
os.environ["OPENAI_API_KEY"] = "sk-3xxx"
os.environ["MODEL_NAME"] = "model-xx"

ragas_examples.text2sql.evals .__file__ = os.path.realpath(__file__)


async def run_evaluation():
    """Run text-to-SQL evaluation with direct code approach."""
    # Load dataset
    dataset = load_dataset()
    print(f"Dataset loaded with {len(dataset)} samples")

    prompt_file_name = f"{os.environ["MODEL_NAME"]}-prompt-v1"

    # Run the experiment
    results = await text2sql_experiment.arun(
        dataset,
        name=prompt_file_name,
        model=os.environ["MODEL_NAME"],
        prompt_file=None,
    )

    # Report results
    print(f"✅ {prompt_file_name}: {len(results)} cases evaluated")

    # Calculate and display accuracy
    accuracy_rate = sum(1 for r in results if r["execution_accuracy"] == "correct") / max(1, len(results))
    print(f"{prompt_file_name} Execution Accuracy: {accuracy_rate:.2%}")


# Run the evaluation
asyncio.run(run_evaluation())
```

**Error trace**
No stack print.

**Expected behavior**
No errors.

**Additional context**

The third param of method [`execution_accuracy`](https://github.com/vibrantlabsai/ragas/blob/84c038d9d479bccd91c9a0afd21861431535a837/examples/ragas_examples/text2sql/evals.py#L32) has no type defined. But the discrecate_metric decorator has a customer validation rule where validate param without specified type as a `str` one, this is the source code that throw `Validation error`
<br/>Image

dhyaneesh:
- Fixes #2512 

## Changes Made

- Changed default type hint from `str` to `t.Any` for unannotated parameters in `_create_pydantic_model()` method (`src/ragas/metrics/decorator.py` line 173)

- Updated comment to reflect that unannotated parameters default to `Any` instead of `str` (line 167)

- This allows unannotated parameters to accept any type (DataFrames, dicts, lists, None, strings, etc.) without validation errors, which is appropriate since the Pydantic model already has `arbitrary_types_allowed=True` configured

## Testing
The changes are too simple for testing , although I have reproduced and tested this locally.

## Steps to Reproduce

1. Create a metric function with an unannotated parameter (no type annotation)

2. Decorate it with `@discrete_metric` decorator

3. Call the metric's `score()` method with a non-string value (e.g., pandas DataFrame) for the unannotated parameter

4. Observe the validation error indicating the parameter should be a string

### DIFF
--- a/src/ragas/metrics/decorator.py
+++ b/src/ragas/metrics/decorator.py
@@ -164,13 +164,13 @@ def create_metric_decorator():
                     field_definitions = {}
 
                     for name, param in sig.parameters.items():
-                        # Get type hint, default to str if no hint available
+                        # Get type hint, default to Any if no hint available
                         type_hint = type_hints.get(name, param.annotation)
                         if type_hint == inspect.Parameter.empty:
                             if param.default != inspect.Parameter.empty:
                                 type_hint = type(param.default)
                             else:
-                                type_hint = str
+                                type_hint = t.Any
 
                         # Get default value
                         if param.default != inspect.Parameter.empty:


anistark:
Thanks for the fix @dhyaneesh 🎉

PR #2568, opened 2026-01-22
fix: fix testset generation bugs

kraftn:

The PR fixes the following bugs related to test set generation.
1. `MultiHopSpecificQuerySynthesizer` often generates samples based on only a single reference context. The root cause is that only one context is selected during scenario preparation when nodes have non-identical themes. 
For example, consider two nodes in a knowledge graph that are connected because they have overlapping keywords. The first node has the keyword "space", and the second one has "spaces". In this case, only one node is selected during scenario generation because independent keywords are used, rather than keyword pairs.
2. When using `Parallel`, its transformations are applied sequently instead of asynchronously.
3. The pandas DataFrame of samples generated using `MultiHopSpecificQuerySynthesizer` and `MultiHopAbstractQuerySynthesizer` doesn't contain information about the personas, query styles and query lengths used during generation.

## Changes Made

1. To fix the first issue, groups of themes are now considered instead of independent themes in `MultiHopSpecificQuerySynthesizer`.
2. The recursive call to `apply_transforms` for a list of transformations of a `Parallel` instance has been replaced with a call to `generate_execution_plan` on the `Parallel` instance.
3. Information about personas, query styles and query lengths has been added to `MultiHopQuerySynthesizer`.
4. Relationships created with `OverlapScoreBuilder` are now marked with `bidirectional = True`. This doesn't affect behavior, but is semantically more correct.

### How to Test
- [ ] Automated tests added/updated
- [x] Manual testing steps:
  1. Apply transformations to a knowledge graph using `Parallel`. Verify that the transofrmations are executed asynchronously. 
  2. Generate a test set using `MultiHopSpecificQuerySynthesizer` and `MultiHopAbstractQuerySynthesizer`, then convert the result to a pandas DataFrame. Verify that the number of contexts is greater than one in all the cases, and DataFrame rows include persona names, query styles and query lengths.

---

### DIFF
--- a/src/ragas/testset/transforms/relationship_builders/traditional.py
+++ b/src/ragas/testset/transforms/relationship_builders/traditional.py
@@ -172,6 +172,7 @@ class OverlapScoreBuilder(RelationshipBuilder):
                                 f"{self.property_name}_{self.new_property_name}": similarity,
                                 "overlapped_items": overlapped_items,
                             },
+                            bidirectional=True,
                         )
                     )
 


anistark (approved):
Thanks for the fix @kraftn 🙌🏼

PR #2612, opened 2026-02-26
feat: support DO_NOT_TRACK and DISABLE_TELEMETRY env vars

### LINKED ISSUE #2603: Support DO_NOT_TRACK and DISABLE_TELEMETRY env vars
**Describe the Feature**
People seem to be standardizing on the DO_NOT_TRACK=1 and/or DISABLE_TELEMETRY=1
Would be nice respect these environment variables and disable tracking if either has a non-empty value.

Reduces mental overhead of managing each project's specific do not track env var (currently `RAGAS_DO_NOT_TRACK`)

**Additional context**
see also:
https://github.com/huggingface/huggingface_hub/issues/1917
https://git.eeqj.de/sneak/consoledonottrack.com/src/branch/master/index.markdown

toroleapinc:
## Summary

Adds support for the standard `DO_NOT_TRACK` and `DISABLE_TELEMETRY` environment variables to disable telemetry, in addition to the existing `RAGAS_DO_NOT_TRACK`.

Closes #2603

## What changed

**`src/ragas/_analytics.py`:**
- Added a `_is_truthy()` helper that treats non-empty values (except `0`, `false`, `no`) as truthy, following the [Console Do Not Track](https://consoledonottrack.com/) standard
- Extended `do_not_track()` to check `DO_NOT_TRACK` and `DISABLE_TELEMETRY` env vars in addition to the existing `RAGAS_DO_NOT_TRACK=true` check
- Fully backward-compatible: existing `RAGAS_DO_NOT_TRACK=true` behavior is preserved

**`tests/unit/test_analytics.py`:**
- Added parametrized tests covering all three env vars with various truthy/falsy values
- Added unit tests for the `_is_truthy()` helper

## How to test

```bash
# Any of these will disable telemetry:
DO_NOT_TRACK=1 python -c "from ragas._analytics import do_not_track; print(do_not_track())"  # True
DISABLE_TELEMETRY=1 python -c "from ragas._analytics import do_not_track; print(do_not_track())"  # True
RAGAS_DO_NOT_TRACK=true python -c "from ragas._analytics import do_not_track; print(do_not_track())"  # True

# Run the tests:
pytest tests/unit/test_analytics.py -v
```


toroleapinc:
The Python 3.9 unit test failure isn't from this PR — it's a `TypeError: unsupported operand type(s) for |: 'ModelMetaclass' and 'NoneType'` coming out of `llama_index/core/indices/prompt_helper.py:558`, which is using PEP 604 union syntax that's not valid in 3.9. So it's an upstream incompat that surfaces during import.

This PR only adds env-var handling for telemetry opt-out, so it shouldn't touch that path. Happy to dig further if you'd like, otherwise the failure should clear once llama_index pins or the 3.9 matrix entry gets dropped.

PR #2630, opened 2026-03-09
fix: close file handle in get_userid() to resolve ResourceWarning

tinspj1997:
open() was used without a context manager, leaving the file handle unclosed after json.load(). Replaced with `with` statement to ensure the file is closed immediately after reading.


tancnle:
👋🏼 @anistark appreciate your thoughts on this PR? We also observed a number of warnings on our build caused by this error.

PR #2638, opened 2026-03-17
fix: allow evaluate() to accept modern v0.4.x collection metrics

### LINKED ISSUE #2624: [Bug]: evaluate() throws TypeError when using the recommended ragas.metrics.collections and llm_factory

**Describe the bug**
In Ragas 0.4.x, users receive `DeprecationWarning`s urging them to migrate from legacy metrics to the modern `ragas.metrics.collections` and to replace `LangchainLLMWrapper` with `llm_factory`. 

However, following these official warnings leads to a hard crash. Passing the newly initialized collection metrics to the standard `evaluate()` function fails. It appears that the `evaluate()` / `aevaluate()` engine still strictly checks if the metrics are instances of the legacy `Metric` base class, which fails for the new collection metrics.

Ragas version: 0.4.3
Python version: 3.11.4

**Code to Reproduce**
```python
from ragas import evaluate, EvaluationDataset
from openai import OpenAI

# 1. Use the officially recommended modern modules
from ragas.metrics.collections import AnswerCorrectness

dataset = EvaluationDataset.from_list([{
    "user_input": "test query",
    "retrieved_contexts": ["test context"],
    "response": "test response",
    "reference": "test reference"
}])

client = OpenAI(api_key="dummy_key")
judge_llm = llm_factory("gpt-4o-mini", client=client)

# 2. Initialize the metric using the modern factory LLM
metric = AnswerCorrectness(llm=judge_llm)

# 3. Trigger the evaluation
# This will crash because evaluate() does not recognize the new metric type
result = evaluate(dataset=dataset, metrics=[metric])
```

**Error trace**
```
  ...
  File ".../site-packages/ragas/evaluation.py", line 457, in _async_wrapper
    return await aevaluate(
  File ".../site-packages/ragas/evaluation.py", line 133, in aevaluate
    raise TypeError(
TypeError: All metrics must be initialised metric objects, e.g: metrics=[BleuScore(), AspectCritic()]
```

**Expected behavior**
If users are strictly instructed by deprecation warnings to use ragas.metrics.collections, the standard evaluate() function should be updated to recognize and accept these new metric instances (instead of strictly checking against the legacy Metric base class).

**Additional context**
The issue stems from the type-check loop in evaluation.py (around line 133):
if isinstance(metrics, list) and any(not isinstance(m, Metric) for m in metrics):
Since the new collection metrics no longer inherit from the legacy Metric class, this check raises a TypeError.

**🌟 Proposed Fix / PR Offer:**
I'd love to submit a PR to fix this type-check mismatch. Could you confirm if I should simply update the isinstance check to include the new collection base classes, or if there's a different approach you prefer for the 0.4.x transition?

tzlwn1:
Fixes #2624 

### Description
In Ragas 0.4.x, users are encouraged to use the modern `ragas.metrics.collections`. However, passing these new metrics to the standard `evaluate()` engine results in a crash because the type-check strictly checks against the legacy `Metric` base class.

This PR updates the `isinstance` validation in `evaluation.py`. It uses a broader check to safely allow both legacy metrics and the new collection-based metrics (which inherit from `SingleTurnMetric` or `MultiTurnMetric`) to pass through.

This provides a smoother transition for users adopting the v0.4.x architecture.


tzlwn1:
Withdrawing this PR — no longer pursuing this change.

PR #2650, opened 2026-03-28
feat: propagate document metadata id as reference_context_ids in synt…

99-NinetyNine:
## Changes Made
-  I have added the attribute reference_context_ids in SingleTurnSample. This is because, if we already have chunks and we need to compute context recall/precision without LLM, then those samples should have the reference chunk id. 
-  To do that, application shall send document_metadata->id as one of node->properties.


99-NinetyNine:
image
Some credit error!

PR #2685, opened 2026-04-16
fix: add ensure_ascii=False to json.dumps during prompt creation

mertsaru:
utf-8 characters such as ü is read as ASCII format \u001f and that breaks the llm by making llm generate tokens indefinetly until fills the token limit

## Changes Made
- add ensure_ascii=False to each json.dumps()  

### How to Test
  0. Use Azure AI with litellm (optional, may not be needed for the test, however during my test, i used Azure AI)
  1. Create a dataset with reference, which the reference have the character 'ü' in it
  2. Use NoiseSensitivity metric in ragas.metrics.collections
  3. Use reference dataset that has the character 'ü' in NoiseSensitivity metric

---


R-behera (commented):
Nice catch. Preserving non-ASCII characters here seems especially important for multilingual evaluation inputs, where escaped JSON can make prompt debugging harder and can subtly change what a human reviewer sees. A small regression test with accented or non-Latin text would make this fix much safer against future prompt-formatting changes.

PR #2687, opened 2026-04-17
fix: drop deprecated top_p for anthropic

### LINKED ISSUE #2674: `InstructorLLM` sends deprecated `top_p` to Anthropic, causing HTTP 400 on current-gen models

**Describe the bug**
When using `llm_factory` with `provider="anthropic"`, Ragas always sends `top_p=0.1` alongside `temperature`. Anthropic has deprecated `top_p` — models released after Claude Opus 4.6 reject any value below `0.99` with a `400 Bad Request` error. This breaks all Ragas evaluations on current Anthropic models out of the box.

Ragas version: 0.4.3
Python version: 3.10.19 (not version-specific — affects any Python 3.9+ since the bug is in Ragas parameter handling)

**Root cause**

`InstructorModelArgs` defines `top_p` as a non-optional field with a hardcoded default:

```python
class InstructorModelArgs(BaseModel):
    temperature: float = 0.01
    top_p: float = 0.1  # always serialized by model_dump()
    max_tokens: int = 1024
```

Because it's non-optional, `model_dump()` always includes it. `_map_provider_params` for Anthropic falls through to the generic pass-through with no filtering:

```python
def _map_provider_params(self) -> t.Dict[str, t.Any]:
    provider_lower = self.provider.lower()

    if provider_lower == "google":
        return self._map_google_params()
    elif provider_lower in ("openai", "azure"):
        return self._map_openai_params()
    else:
        # Anthropic, LiteLLM, and other providers - pass through unchanged
        return self.model_args.copy()
```

So `top_p=0.1` is always forwarded to the Anthropic API. This can be verified directly:

```python
from anthropic import AsyncAnthropic

llm = llm_factory(
    model="claude-sonnet-4-6",
    provider="anthropic",
    client=AsyncAnthropic(),
    temperature=0,
)

print(llm.model_args)
# {'temperature': 0, 'top_p': 0.1, 'max_tokens': 1024}

print(llm._map_provider_params())
# {'temperature': 0, 'top_p': 0.1, 'max_tokens': 1024}
```

**Anthropic deprecation**

From the [Anthropic Python SDK docs](https://platform.claude.com/docs/en/api/python/messages/create):

> **Deprecated.** Models released after Claude Opus 4.6 do not support setting `top_p`. A value >= 0.99 will be accepted for backwards compatibility, all other values will be rejected with a 400 error. You should either alter `temperature` or `top_p`, but not both.

**Code to Reproduce**
```python
import asyncio
from anthropic import AsyncAnthropic

async def test():
    client = AsyncAnthropic()
    await client.messages.create(
        model="claude-sonnet-4-6",
        max_tokens=10,
        temperature=0,
        top_p=0.1,
        messages=[{"role": "user", "content": "hi"}],
    )

asyncio.run(test())
# BadRequestError: Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error',
# 'message': '`temperature` and `top_p` cannot both be specified for this model. Please use only one.'}}
```

**Suggested fix**

Add an Anthropic-specific branch to `_map_provider_params`, following the existing pattern for Google and OpenAI:

```python
def _map_provider_params(self) -> t.Dict[str, t.Any]:
    provider_lower = self.provider.lower()

    if provider_lower == "google":
        return self._map_google_params()
    elif provider_lower in ("openai", "azure"):
        return self._map_openai_params()
    elif provider_lower == "anthropic":
        return self._map_anthropic_params()
    else:
        # LiteLLM and other providers - pass through unchanged
        return self.model_args.copy()

def _map_anthropic_params(self) -> t.Dict[str, t.Any]:
    """Map parameters for Anthropic models.

    top_p is deprecated for models released after Claude Opus 4.6.
    Values below 0.99 are rejected with a 400 error, so it must be removed.
    See: https://platform.claude.com/docs/en/api/python/messages/create
    """
    mapped_args = self.model_args.copy()
    mapped_args.pop("top_p", None)
    return mapped_args
```

**Expected behavior**
`top_p` is not sent to the Anthropic API unless explicitly set by the caller.

**Additional context**
Workaround until fixed: `llm.model_args.pop("top_p", None)` after calling `llm_factory`

hyl64:
## Summary
This PR stops passing the deprecated `top_p` parameter to Anthropic provider requests from InstructorLLM, avoiding current-model HTTP 400 failures.

## Issues
- fixes #2674

## Verification
- py_compile passed
- source-level runtime verification confirmed that the Anthropic provider params no longer include `top_p`


R-behera (commented):
This looks like a sensible provider-specific fix, and the targeted test helps. One edge case worth thinking about is whether older Anthropic-compatible routes still expect `top_p` to be forwarded. If you intentionally want this behavior only for current Claude generations, a short note in the mapping logic or a model-name-specific test could make that contract clearer for future maintainers.

PR #2697, opened 2026-04-30
fix(llms): drop deprecated Anthropic top_p default in InstructorLLM

GopalGB:
## Summary

- Anthropic deprecated `top_p` for models released after Claude Opus 4.6; values < 0.99 get rejected with HTTP 400.
- `InstructorModelArgs.top_p = 0.1` (the library default) is therefore unusable on current Anthropic models — every Ragas evaluation 400s out of the box.
- Add an Anthropic-specific branch in `_map_provider_params` that strips the hardcoded default while preserving values users explicitly opt into.

## Why this fix

- The default top_p is forwarded unconditionally because `InstructorModelArgs.model_dump()` always serialises every field.
- Stripping the value at the provider-mapping layer (rather than making the field optional) keeps the existing `model_args` dict intact for OpenAI/Google/LiteLLM, where the default is still valid.
- Conditional drop (`== default`) means power users who pass `top_p=0.99` still get their value through.

## How to test

```bash
make install-minimal
pytest tests/unit/llms/test_instructor_factory.py -k anthropic -v
```

The two new test cases cover the default-stripped path and the explicit-opt-in path.

Refs: #2674


GopalGB:
Friendly ping — drops Anthropic's deprecated `top_p=1.0` default in `InstructorLLM`. github-actions confirmed no docs update needed (bug fix only). cc maintainers.

PR #2704, opened 2026-05-10
fix: compute factual correctness recall from reference coverage

Raditya0902:
## Summary

Fixes `FactualCorrectness(mode="recall")` in the collections implementation to compute recall from reference claim coverage.

Previously, recall used true positives from response-claim verification together with false negatives from reference-claim verification. This mixed two different claim sets and could produce an incorrect recall score.

## Changes

- Computes recall using `response_reference`, i.e. reference claims verified against the response
- Adds a deterministic unit test for asymmetric claim coverage
- Leaves precision, F1, and the legacy implementation unchanged

## Validation

- `uv run pytest tests/unit/test_factual_correctness_collections.py -v` — 1 passed
- `uv run pytest tests/e2e/metrics_migration/test_factual_correctness_migration.py -v` — 1 passed, 4 skipped


Raditya0902:
Checks/tests are passing locally — happy to address any feedback. Let me know if this needs anything before it can be reviewed.

PR #2707, opened 2026-05-10
fix: use chunk summary when filtering prechunked nodes

### LINKED ISSUE #2680: `CustomNodeFilter` skips pre-chunked `CHUNK` nodes even when `SummaryExtractor` added a chunk summary

**Describe the bug**

When using pre-chunked testset generation, `CustomNodeFilter` skips `CHUNK` nodes that do not have a parent `DOCUMENT`, even if the chunk itself already has a `summary`.

This happens with the default pre-chunked transform path:

```python
TestsetGenerator.generate_with_chunks(...)
```

`default_transforms_for_prechunked()` applies both `SummaryExtractor` and `CustomNodeFilter` to `NodeType.CHUNK` nodes. `SummaryExtractor` writes `summary` directly onto each chunk. However, `CustomNodeFilter.custom_filter()` only looks for a parent document summary when the node type is `CHUNK`.

In pre-chunked generation, chunks are created directly and usually do not have parent document nodes. As a result, `CustomNodeFilter` sets `summary = ""`, logs a warning, and skips filtering.

Observed warning:

```text
WARNING:ragas.testset.transforms.filters:Node <uuid> does not have a summary. Skipping filtering.
```

This warning is misleading because the chunk can have its own `summary`.

Ragas version: `0.4.3`  
Python version: `3.11.9`

**Code to Reproduce**

This minimal reproduction avoids external LLM calls and directly exercises `CustomNodeFilter.custom_filter()`:

```python
import asyncio
import logging

from ragas.llms import BaseRagasLLM
from ragas.testset.graph import KnowledgeGraph, Node, NodeType
from ragas.testset.transforms.filters import (
    CustomNodeFilter,
    QuestionPotentialOutput,
)

logging.basicConfig(level=logging.WARNING)


class DummyLLM(BaseRagasLLM):
    def generate_text(self, *args, **kwargs):
        pass

    async def agenerate_text(self, *args, **kwargs):
        pass

    def is_finished(self, response):
        return True


class RecordingScoringPrompt:
    def __init__(self):
        self.calls = []

    async def generate(self, data, llm):
        self.calls.append(data)
        return QuestionPotentialOutput(score=3)


async def main():
    scoring_prompt = RecordingScoringPrompt()

    node = Node(
        type=NodeType.CHUNK,
        properties={
            "page_content": "This is the chunk content.",
            "summary": "This is the chunk summary.",
        },
    )
    kg = KnowledgeGraph(nodes=[node])

    node_filter = CustomNodeFilter(
        llm=DummyLLM(),
        scoring_prompt=scoring_prompt,
    )

    result = await node_filter.custom_filter(node, kg)

    print("filter result:", result)
    print("scoring prompt calls:", len(scoring_prompt.calls))


asyncio.run(main())
```

Current output on `0.4.3`:

```text
WARNING:ragas.testset.transforms.filters:Node <uuid> does not have a summary. Skipping filtering.
filter result: False
scoring prompt calls: 0
```

The scoring prompt is never called, even though the `CHUNK` node has a `summary`.

**Error trace**

No exception is raised. The issue appears as repeated warnings during `generate_with_chunks()`:

```text
WARNING:ragas.testset.transforms.filters:Node <uuid> does not have a summary. Skipping filtering.
```

**Expected behavior**

For `CHUNK` nodes, `CustomNodeFilter` should:

1. Prefer the parent document summary when a parent document exists and has a non-empty `summary`.
2. Fall back to the chunk node’s own `summary` when there is no parent summary.
3. Keep the existing skip behavior only when neither parent summary nor chunk summary exists.

This would preserve existing non-pre-chunked behavior while making the default pre-chunked transform path work as intended.

**Additional context**

The issue appears to be caused by a mismatch between `default_transforms_for_prechunked()` and `CustomNodeFilter`.

`default_transforms_for_prechunked()` applies:

```python
summary_extractor = SummaryExtractor(llm=llm, filter_nodes=filter_chunks)
node_filter = CustomNodeFilter(llm=llm, filter_nodes=filter_chunks)
```

So summaries are extracted directly onto `CHUNK` nodes.

But `CustomNodeFilter.custom_filter()` currently does this for chunks:

```python
if node.type.name == "CHUNK":
    parent_nodes = get_parent_nodes(node, kg)
    if len(parent_nodes) > 0:
        summary = parent_nodes[0].properties.get("summary", "")
    else:
        summary = ""
else:
    summary = node.properties.get("summary", "")
```

In pre-chunked mode, `parent_nodes` is empty, so the chunk’s own `summary` is ignored.

A minimal fix would be to fall back to `node.properties["summary"]` for `CHUNK` nodes when no non-empty parent summary is available.

Raditya0902:

- Fixes #2680

`CustomNodeFilter` treated all `CHUNK` nodes as split chunks and only looked for summaries on parent document nodes. For pre-chunked inputs, `CHUNK` nodes may not have parent document relationships. Even when `SummaryExtractor` adds a summary directly to the chunk node, the filter ignored it, treated the summary as missing, and skipped filtering.

## Changes Made

- Updated `CustomNodeFilter` to fall back to the chunk node's own `summary` when a `CHUNK` node has no parent
- Preserved existing behavior for parented `CHUNK` nodes, which still use the parent summary
- Added a deterministic unit test for a parentless pre-chunked `CHUNK` node with a summary

### How to Test

- [x] Manual testing steps:
  1. Run `uv run pytest tests/unit/test_prechunked_generation.py -v`
  2. Run `make format`

Validation results:

- `uv run pytest tests/unit/test_prechunked_generation.py -v` — 6 passed, 1 existing collection warning
- `make format` — passed

## References

- Related issues: #2680
- Documentation:
- External references:

Not applicable.


Raditya0902:
Checks/tests are passing locally — happy to address any feedback. Let me know if this needs anything before it can be reviewed.

PR #2720, opened 2026-05-14
fix(examples): guard against IndexError on empty LLM choices list

qizwiz:
## Summary

Five `ragas_examples` scripts access `response.choices[0]` without first checking whether the LLM API returned any choices. When the API returns an empty list (content-policy filter, truncated streaming response, rate-limit reply), this raises a bare `IndexError`.

### Changes

Minimal inline guards at each call site — no new files:

| File | Violation sites | Fix |
|------|-----------------|-----|
| `agent_evals/agent.py` | 1 | `if not response.choices: raise ValueError(...)` before `choices[0]` (caught by outer try) |
| `rag_eval/rag.py` | 1 | `if not response.choices: raise ValueError(...)` before `choices[0]` (caught by outer try) |
| `text2sql/analyze_errors.py` | 1 | Early return with error dict if no choices |
| `prompt_evals/prompt.py` | 2 (shared ternary) | Guard ternary: `if response.choices and response.choices[0].message.content` |
| `workflow_eval/workflow.py` | 6 (3 ternary pairs) | Guard each ternary: `if response.choices and response.choices[0].message.content` |


qizwiz:
Hi — gentle ping on this PR. The fix is minimal (10-line diff, 5 guard checks on  access in example scripts). Happy to answer any questions or make adjustments if needed.

qizwiz:
**Note: discovered a second crash vector**

Testing against live providers revealed that Gemini 2.5 Flash returns HTTP 200 with `choices[0].message = None` (not an empty `choices` list) when content is filtered — `finish_reason: PROHIBITED_CONTENT`. This causes `AttributeError: 'NoneType' object has no attribute 'content'` even when `choices` is non-empty and the first check passes.

The comprehensive guard covering both paths:
```python
if not response.choices or response.choices[0].message is None:
```

If any files in this PR only guard against empty `choices`, I can update them.

PR #2733, opened 2026-05-21
fix: correct recall and f1 TP computation in FactualCorrectness metric

### LINKED ISSUE #2693: FactualCorrectness(mode="recall") returns precision score instead of recall
* RAGAS version: 0.4.3
* Python version: 3.12

# Summary

FactualCorrectness(mode="recall") returns a score that is not aligned with what we could expect from the reference's claim decomposition. I think the bug is in `_single_turn_ascore` in `src/ragas/metrics/_factual_correctness.py`.

# Root cause

The method runs two NLI tasks:
reference_response → decomposes response, verifies against reference (precision direction)
response_reference → decomposes reference, verifies against response (recall direction)

Then:

```
tp = sum(reference_response)   # ← "precision" TP: response claims supported by reference
fn = sum(~response_reference)  # ← recall FN: reference claims NOT supported by response

# mode="recall":
score = tp / (tp + fn)         # ← WRONG: uses "precision" TP in recall formula
```

tp should be sum(response_reference) (reference claims supported by response) when computing recall, in order to make sense regarding the reference's claim decomposition. Using sum(reference_response) (response claims supported by reference) is the "precision" TP, not the "recall" TP.

# Concrete example I got

* Response: 4 claims, 3 supported by reference → precision TP = 3
* Reference: 3 claims, 2 supported by response → recall TP = 2, recall FN = 1
* Expected recall = 2 / (2 + 1) = 0.67
* Actual result with mode="recall": 3 / (3 + 1) = 0.75

# Expected behavior

```
if self.mode == "precision":
    score = tp / (tp + fp + 1e-8)         # tp = sum(reference_response)
elif self.mode == "recall":
    tp_recall = sum(response_reference)    # ← correct TP for recall
    score = tp_recall / (tp_recall + fn + 1e-8)
else:  # f1
    precision = sum(reference_response) / (len(reference_response) + 1e-8)
    recall    = sum(response_reference) / (len(response_reference) + 1e-8)
    score = fbeta_score(...)
```

rajkumar-prog:
## Problem
Fixes #2693

`FactualCorrectness._single_turn_ascore` runs two NLI tasks:

| Variable | What it computes |
|---|---|
| `reference_response` | Response claims verified against reference — **precision direction** |
| `response_reference` | Reference claims verified against response — **recall direction** |

Before this fix, both `recall` and `f1` modes used the same `tp = sum(reference_response)`, which is the **precision TP**. This causes:

- **Recall mode**: inflated score — uses response-claim TP where reference-claim TP is required
- **F1 mode**: recall leg of fbeta_score was also computed with the wrong TP

### Concrete example from the issue

| | Value |
|---|---|
| Response claims supported by reference (precision TP) | 3 of 4 |
| Reference claims supported by response (recall TP) | 2 of 3 |
| **Old recall** (wrong) | `3 / (3+1) = 0.75` |
| **New recall** (correct) | `2 / (2+1) = 0.667` |

## Fix

**`_factual_correctness.py`** — introduce `tp_precision` and `tp_recall` as distinct variables:
```python
tp_precision = sum(reference_response)  # response claims supported by reference
tp_recall    = sum(response_reference)  # reference claims supported by response
fn           = sum(~response_reference)

# recall mode (fixed)
score = tp_recall / (tp_recall + fn + 1e-8)

# f1 mode (fixed)
score = fbeta_score(tp_precision, fp, tp_recall, fn, self.beta)
```

**`metrics/utils.py`** — `fbeta_score` now accepts separate `tp_precision` and `tp_recall` so precision and recall legs are computed independently.

## Tests
Added 9 unit tests in `tests/unit/test_factual_correctness_scores.py`:
- `fbeta_score` correctness for edge cases and asymmetric precision/recall
- Regression test that explicitly catches the old wrong value (`0.75`) for recall
- All three modes (`precision`, `recall`, `f1`) tested for perfect, zero, and mixed scores
- All 9 tests pass ✅


rajkumar-prog:
Closing — this was accidentally submitted to vibrantlabsai/ragas. The correct PR targeting explodinggradients/ragas will be opened separately.

PR #2757, opened 2026-06-01
Replace deprecated asyncio.get_event_loop() with get_running_loop()

kratos0718:
## Summary

`asyncio.get_event_loop()` is deprecated since Python 3.10 when there is no running loop, and emits a `DeprecationWarning` ([docs](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_event_loop)).

This replaces it with `asyncio.get_running_loop()` at three call sites that all run inside `async` functions and only use the loop for `run_in_executor()` — a running loop is guaranteed there, so it's a direct, non-deprecated drop-in:

- `src/ragas/embeddings/base.py` — `BaseRagasEmbeddings.embed_texts` (sync branch)
- `src/ragas/embeddings/utils.py` — `run_sync_in_async`
- `src/ragas/llms/oci_genai_wrapper.py` — `OCIGenAIWrapper.agenerate_text`

No behavior change. Call sites that use the `get_event_loop()` + `loop.is_running()` pattern to *detect* whether a loop is already running are intentionally left unchanged, since `get_running_loop()` raises instead of returning when no loop is active.

## Test plan

- `ruff check` passes on all three files.
- Pure timing/executor-loop swap; behavior of the affected coroutines is unchanged.


kratos0718:
Friendly ping — this has been open since 1 June.

Re-checked against current `main` before pinging: `asyncio.get_event_loop()` is still present in all three files the PR touches (`embeddings/base.py`, `embeddings/utils.py`, `llms/oci_genai_wrapper.py`), so the change still applies and the branch is `MERGEABLE`.

`get_event_loop()` has been deprecated since 3.10 and raises a DeprecationWarning outside a running loop, so this is mostly future-proofing rather than a live bug. Happy to close it if that isn't a direction you want to take.

PR #2759, opened 2026-06-05
fix: make VertexAI imports optional to prevent ImportError on `import ragas`

### LINKED ISSUE #2753: Title: ModuleNotFoundError on import: langchain_community.chat_models.vertexai  removed in newer langchain-community versions
I have checked the [[documentation](https://docs.ragas.io/)](https://docs.ragas.io/) and related resources and couldn't resolve my bug.**

**Describe the bug**

`import ragas` crashes immediately on startup with `ModuleNotFoundError` for users on modern `langchain-community` who are NOT using Google VertexAI. The crash happens before any user code runs — simply importing the library fails.

The root cause is in `src/ragas/llms/base.py` — this line executes unconditionally at import time:

from langchain_community.chat_models.vertexai import ChatVertexAI

`ChatVertexAI` was deprecated in `langchain-community` since version `0.0.12` and fully moved to the separate `langchain-google-vertexai` package. In modern installations of `langchain-community`, this path no longer exists — causing ragas to crash for any user not on Google Cloud, including Groq, OpenAI, Bedrock, and Anthropic users.

**Ragas version:** 0.4.3
**Python version:** 3.11.14
**langchain-community version:** latest

**Code to Reproduce**

uv init test-ragas
cd test-ragas
uv add ragas langchain-community langchain-groq
uv run python -c "import ragas; print('ok')"

**Error trace**

  File "<string>", line 1, in <module>
  File ".venv/lib/python3.11/site-packages/ragas/__init__.py", line 3, in <module>
    from ragas.evaluation import evaluate
  File ".venv/lib/python3.11/site-packages/ragas/evaluation.py", line 28, in <module>
  File ".venv/lib/python3.11/site-packages/ragas/llms/__init__.py", line 1, in <module>
    from ragas.llms.base import (
  File ".venv/lib/python3.11/site-packages/ragas/llms/base.py", line 12, in <module>
    from langchain_community.chat_models.vertexai import ChatVertexAI
ModuleNotFoundError: No module named 'langchain_community.chat_models.vertexai'

**Expected behavior**

`import ragas` should succeed regardless of whether the user has VertexAI installed. VertexAI is an optional integration — it should not be a hard unconditional import that blocks all non-Google-Cloud users from using the library.

**Proposed fix**

In `src/ragas/llms/base.py`, replace the hard import with a layered fallback:

```python
# Current (broken for modern langchain-community):
from langchain_community.chat_models.vertexai import ChatVertexAI

# Proposed fix — three-level fallback:
try:
    from langchain_google_vertexai import ChatVertexAI          # correct modern location
except ImportError:
    try:
        from langchain_community.chat_models.vertexai import ChatVertexAI  # legacy fallback
    except ImportError:
        ChatVertexAI = None  # VertexAI not installed — non-VertexAI users unaffected


This approach:
- Works for users with `langchain-google-vertexai` installed (new correct path)
- Still works for legacy installations with old `langchain-community` (backward compatible)
- Does not break non-VertexAI users (Groq, OpenAI, Bedrock, Anthropic)

**Workaround for users hitting this right now**

Either manually patch `.venv/lib/python3.11/site-packages/ragas/llms/base.py` with the try/except above, or install `langchain-google-vertexai` as an additional dependency.

**Additional context**

The same deprecated import pattern likely applies to `VertexAI` (non-chat LLM) on the next line of `base.py`. Worth auditing the full file for any other `langchain_community` VertexAI imports while this is being addressed.

Also noting: the newer ragas docs now recommend `litellm` as the integration path for VertexAI. If the team is moving in that direction, removing the direct `ChatVertexAI` import entirely in favour of the `litellm` factory would be the cleaner long-term fix. Happy to submit a PR once the preferred direction is confirmed.

### PR DESCRIPTION
**Pull request by Incheonkirin**
## Summary

`import ragas` currently crashes with `ModuleNotFoundError` for any user **not** on Google VertexAI (OpenAI / Anthropic / Groq / Bedrock / …). The crash happens at import time, before any user code runs.

Fixes #2753.

## Root cause

`src/ragas/llms/base.py` imports the VertexAI classes unconditionally at module top:

```python
from langchain_community.chat_models.vertexai import ChatVertexAI
from langchain_community.llms import VertexAI
```

`ChatVertexAI` / `VertexAI` were deprecated in `langchain-community` and moved to the separate, optional `langchain-google-vertexai` package. On modern `langchain-community` installs that path no longer exists, so simply `import ragas` fails.

## Fix

- Guard both imports with `try/except ImportError` (set the names to `None` on failure).
- Build `MULTIPLE_COMPLETION_SUPPORTED` with the VertexAI classes **only when importable**, so behavior is unchanged when `langchain-google-vertexai` is installed.

Minimal and backward-compatible — no new dependency, no behavior change for VertexAI users.

## Testing

- `python -m py_compile src/ragas/llms/base.py` passes.
- Verified both guard branches: VertexAI present → all 6 classes registered; absent → 4 registered and no import-time crash.


Incheonkirin:
Replacement PR is now #2837 from a proper fork. It contains the rebased change and current validation; no reopen is needed.

Issue #2745, opened 2026-05-25
ragas 0.4.3: ChatVertexAI import broken — uses removed langchain_community path


decko:
## Bug

`ragas 0.4.3` fails to import on any installation with modern `langchain-community`:

```
File "ragas/llms/base.py", line 12, in <module>
    from langchain_community.chat_models.vertexai import ChatVertexAI
ModuleNotFoundError: No module named 'langchain_community.chat_models.vertexai'
```

## Root cause

`ChatVertexAI` was moved from `langchain-community` to `langchain-google-vertexai` in the langchain ecosystem. The import in `ragas/llms/base.py:12` still references the old location.

## Fix

```python
# Before (broken)
from langchain_community.chat_models.vertexai import ChatVertexAI

# After (correct)
from langchain_google_vertexai import ChatVertexAI
```

And add `langchain-google-vertexai` to ragas's dependencies.

## Reproduction

```bash
pip install ragas==0.4.3 langchain-community langchain-google-vertexai
python -c "from ragas.llms import llm_factory"  # fails
```

## Workaround

Pin `ragas<0.4` (`ragas==0.3.9` works).

## Environment

- ragas 0.4.3
- langchain-community 0.3.x
- Python 3.12

Sanket2329:
hey @decko 
if it is available i would like to work on it!

decko:
@Sanket2329 I'm not a developer or CODEOWNER on ragas project. Feel free to work on this issue. Yet, check the project's contribution guidelines to understand how to properly work with it.

athuKawale:
this worked for me:
uvx \
  --with "langchain-community<0.4" \
  --with "langchain-google-vertexai" \
  ragas quickstart rag_eval

decko:
> this worked for me: uvx --with "langchain-community<0.4" --with "langchain-google-vertexai" ragas quickstart rag_eval

For me too @athuKawale. But the issue is that under the current situation, we can't use the latest langchain-community version.

Issue #2741, opened 2026-05-24
No module named 'langchain_community.chat_models.vertexai'


liuwencan:
Here's how to solve this problem. 
Error:

(base) (ragas_rel) PS E:\code\python-project\ragas_rel> uv run python main.py          
Traceback (most recent call last):
  File "E:\code\python-project\ragas_rel\main.py", line 5, in <module>
    from ragas.llms import llm_factory
  File "E:\code\python-project\ragas_rel\.venv\Lib\site-packages\ragas\__init__.py", line 5, in <module>
    from ragas.evaluation import aevaluate, evaluate
  File "E:\code\python-project\ragas_rel\.venv\Lib\site-packages\ragas\evaluation.py", line 31, in <module>
    from ragas.llms import llm_factory
  File "E:\code\python-project\ragas_rel\.venv\Lib\site-packages\ragas\llms\__init__.py", line 1, in <module>
    from ragas.llms.base import (
  File "E:\code\python-project\ragas_rel\.venv\Lib\site-packages\ragas\llms\base.py", line 12, in <module>
    from langchain_community.chat_models.vertexai import ChatVertexAI
ModuleNotFoundError: No module named 'langchain_community.chat_models.vertexai'

JavaCS3:
I’ve also just started using this library and ran into the same issues as you. I’m so disappointed with it that I probably won’t use it again in the future.

Once I finish running the documentation, I’m planning to switch to something else. It seems like a project called OpenEval might be better.

anakin87:
Until ragas fixes this issue upstream, a possible workaround is to pin `langchain-community<0.4.2`.

sohamadhikari2000:
#Hi! 👋

`ChatVertexAI` has been deprecated. You can switch to `ChatGoogleGenerativeAI` and `GoogleGenerativeAI` from the `langchain_google_genai` package instead.

I’ve opened a PR that updates these usages and adds the required dependency. Feel free to take a look, and let me know if you have any questions or run into issues 🙂
PR: #2739

GoldSharon:
Hi @liuwencan, I ran into the same issue and found the root cause.

`langchain_community.chat_models.vertexai` was removed in newer versions.
The fix is to use `langchain_google_vertexai` directly instead.

I've submitted PR #2746 that fixes this. You can try it by installing from my fork until it gets merged:

pip install git+https://github.com/GoldSharon/ragas.git

Environment where fix was verified:
- ragas: 0.4.3
[...]

PR #2784, opened 2026-06-23
fix(embeddings): keep usage telemetry from breaking on non-string model names

### LINKED ISSUE #2783: EmbeddingUsageEvent telemetry NaNs a metric when embeddings expose a non-string .model (e.g. fastembed)

**Describe the bug**

Embeddings whose `.model` attribute is not a string (for example `langchain_community`'s `FastEmbedEmbeddings`, which stores the loaded model *object* on `.model`) cause `LangchainEmbeddingsWrapper` to raise a Pydantic `ValidationError` from *usage telemetry*. The error propagates out of the embed call, so during `evaluate()` the affected metric (e.g. `answer_relevancy`, which needs embeddings) silently becomes `NaN`, while LLM-only metrics score normally.

Root cause: in `src/ragas/embeddings/base.py`, the wrapper logs `track(EmbeddingUsageEvent(model=getattr(self.embeddings, "model", None), ...))`. `EmbeddingUsageEvent.model` is typed `Optional[str]`, so a non-string object fails validation. `track()` is decorated `@silent` so telemetry never breaks user code, but the event is constructed as the *argument* to `track()`, so the error is raised before `@silent` can catch it. Setting `RAGAS_DO_NOT_TRACK=true` does not help either, because the opt-out check lives inside `track()`, after the argument has already been built.

Ragas version: 0.4.3 (also present on `main`)
Python version: 3.12

**Code to Reproduce**

```python
from langchain_community.embeddings.fastembed import FastEmbedEmbeddings
from ragas.embeddings import LangchainEmbeddingsWrapper

emb = LangchainEmbeddingsWrapper(FastEmbedEmbeddings())
emb.embed_query("hello")
```

**Error trace**

```
ValidationError: 1 validation error for EmbeddingUsageEvent
model
  Input should be a valid string [type=string_type, input_value=<fastembed.text.text_embedding.TextEmbedding object at 0x...>, input_type=TextEmbedding]
    For further information visit https://errors.pydantic.dev/2.13/v/string_type
```

Raised from `LangchainEmbeddingsWrapper.embed_query` at `src/ragas/embeddings/base.py`:

```python
track(
    EmbeddingUsageEvent(
        provider="langchain",
        model=getattr(self.embeddings, "model", None),  # non-string for fastembed
        embedding_type="legacy",
        num_requests=1,
        is_async=False,
    )
)
```

In a full `evaluate()` run this surfaces as `answer_relevancy` returning `NaN` (the executor logs `Exception raised in Job[...]: ValidationError(... for EmbeddingUsageEvent ...)`).

**Expected behavior**

Usage telemetry must never break an embedding call. A non-string `.model` should degrade to a best-effort string name or `None`, not raise. This restores the best-effort behavior that the `@silent` decorator on `track()` already intends.

**Additional context**

I have a fix + regression test ready and will open a PR referencing this issue. The fix extracts a safe string model name (falling back to `.model_name`, then `None`) at the telemetry call sites in `LangchainEmbeddingsWrapper`.

aneeshv12:
- Fixes #2783

`LangchainEmbeddingsWrapper` logs an `EmbeddingUsageEvent` for usage telemetry whose `model` field is typed `Optional[str]`, via `getattr(self.embeddings, "model", None)`. When the wrapped embeddings store a non-string object on `.model` (e.g. `langchain_community`'s `FastEmbedEmbeddings`, which keeps the loaded `TextEmbedding` object there), constructing the event raises a Pydantic `ValidationError`. Although `track()` is decorated `@silent` so telemetry can never break user code, the event is built as the *argument* to `track()`, so the error is raised before `@silent` can catch it. `RAGAS_DO_NOT_TRACK=true` does not help either, since the opt-out check runs inside `track()` after the argument is already constructed.

Net effect: in a full `evaluate()`, `answer_relevancy` (which needs embeddings) silently returns `NaN` while LLM-only metrics score normally.

## Changes Made
- Add `_extract_string_model_name()` in `src/ragas/embeddings/base.py`: returns a real `str` model name, falling back `.model` -> `.model_name` -> `None`, so the usage event is always valid.
- Use it at the four telemetry call sites in `LangchainEmbeddingsWrapper` (`embed_query`, `embed_documents`, `aembed_query`, `aembed_documents`) in place of `getattr(self.embeddings, "model", None)`.
- This restores the best-effort behavior that the `@silent` decorator on `track()` already intends: telemetry can no longer break an embedding call.

### How to Test
- [ ] Manual testing steps:
  1. `from langchain_community.embeddings.fastembed import FastEmbedEmbeddings`
  2. `from ragas.embeddings import LangchainEmbeddingsWrapper`
  3. `LangchainEmbeddingsWrapper(FastEmbedEmbeddings()).embed_query("hello")` — raises `ValidationError` before this change, returns a vector after.

Added `test_embedding_usage_telemetry_tolerates_nonstring_model` in `tests/unit/test_analytics.py`: it wraps an embeddings stub whose `.model` is a non-string object and asserts `embed_query` / `embed_documents` succeed. The test fails on `main` (ValidationError) and passes with this fix.

Verification:
- `ruff format` / `ruff check`: clean
- `pyright src/ragas/embeddings/base.py`: 0 errors
- `pytest tests/unit/test_analytics.py tests/unit/test_embeddings.py`: passing

## References
- Related issues: #2783 (this bug). Unrelated but worth noting: a fresh install currently fails `import ragas` on the removed `langchain_community...vertexai` path under langchain 1.x (#2741 / #2745 / #2753), so upstream CI may be red until that lands. This change is independent of it.
- Documentation: n/a
- External references: n/a

Before (any embeddings whose `.model` is a non-string object, e.g. fastembed):
```
ValidationError: 1 validation error for EmbeddingUsageEvent
model
  Input should be a valid string [type=string_type, input_value=<...TextEmbedding object...>, input_type=TextEmbedding]
```
After: the embed call returns its vector and `answer_relevancy` scores normally.


aneeshv12:
Heads-up for reviewers: the `check-docs` job failed on its own step, not on this change. The error is:

```
--json-schema was provided but Claude did not return structured_output. Result subtype: success
```

i.e. the documentation-analysis action received a successful model response but not in the structured-output format it expected, so it errored out before evaluating the diff. The Read the Docs build passed, and the changed code (the new `_extract_string_model_name` helper and the regression test) both carry docstrings, so no documentation updates are needed here.

Happy to rebase or adjust if you'd like the docs job re-run. Thanks for taking a look!

PR #2786, opened 2026-06-23
Guard LangChain embedding telemetry model names

### LINKED ISSUE #2783: EmbeddingUsageEvent telemetry NaNs a metric when embeddings expose a non-string .model (e.g. fastembed)

yinli-systems:
## Summary

Fixes #2783 by normalizing the model name used for legacy LangChain embedding telemetry before constructing `EmbeddingUsageEvent`.

Some LangChain embeddings expose a loaded model object on `.model` instead of a string model name. Since `EmbeddingUsageEvent.model` is typed as `str | None`, constructing the telemetry event could raise before `track()` had a chance to silence telemetry errors. The wrapper now uses a telemetry-safe helper that accepts string `.model`, falls back to string `.model_name`, and otherwise records `None`.

## Tests

- `git diff --check`
- `PYTHONPYCACHEPREFIX=/private/tmp/ragas-pycache PYTHONPATH=src /usr/bin/python3 -m py_compile src/ragas/embeddings/base.py tests/unit/test_embeddings.py`
- `UV_CACHE_DIR=/private/tmp/uv-cache-ragas uv run ruff check src/ragas/embeddings/base.py tests/unit/test_embeddings.py`
- `UV_CACHE_DIR=/private/tmp/uv-cache-ragas uv run ruff format --check src/ragas/embeddings/base.py tests/unit/test_embeddings.py`
- `UV_CACHE_DIR=/private/tmp/uv-cache-ragas PYTHONPATH=/private/tmp/ragas-stubs:src uv run --extra test pytest -q tests/unit/test_embeddings.py`

Note: the pytest command uses a temporary local stub for `langchain_community.chat_models.vertexai` because the current test environment hits the existing import issue tracked separately in #2753 before this test module runs. The stub is not part of this PR.

### DIFF
--- a/src/ragas/embeddings/base.py
+++ b/src/ragas/embeddings/base.py
@@ -257,6 +257,15 @@ class BaseRagasEmbeddings(Embeddings, ABC):
         )
 
 
+def _safe_langchain_embedding_model_name(embeddings: Embeddings) -> str | None:
+    """Return a telemetry-safe model name for LangChain embeddings."""
+    for attr in ("model", "model_name"):
+        value = getattr(embeddings, attr, None)
+        if isinstance(value, str):
+            return value
+    return None
+
+
 class LangchainEmbeddingsWrapper(BaseRagasEmbeddings):
     """
     Wrapper for any embeddings from langchain.
@@ -305,7 +314,7 @@ class LangchainEmbeddingsWrapper(BaseRagasEmbeddings):
         track(
             EmbeddingUsageEvent(
                 provider="langchain",
-                model=getattr(self.embeddings, "model", None),
+                model=_safe_langchain_embedding_model_name(self.embeddings),
                 embedding_type="legacy",
                 num_requests=1,
                 is_async=False,
@@ -323,7 +332,7 @@ class LangchainEmbeddingsWrapper(BaseRagasEmbeddings):
         track(
             EmbeddingUsageEvent(
                 provider="langchain",
-                model=getattr(self.embeddings, "model", None),
+                model=_safe_langchain_embedding_model_name(self.embeddings),
                 embedding_type="legacy",
                 num_requests=len(texts),
                 is_async=False,
@@ -341,7 +350,7 @@ class LangchainEmbeddingsWrapper(BaseRagasEmbeddings):
         track(
             EmbeddingUsageEvent(
                 provider="langchain",
-                model=getattr(self.embeddings, "model", None),
+                model=_safe_langchain_embedding_model_name(self.embeddings),
                 embedding_type="legacy",
                 num_requests=1,
                 is_async=True,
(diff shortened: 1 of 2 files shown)


yinli-systems:
The failing `check-docs` job appears to be an infrastructure/action failure rather than a documentation issue from this PR. The job failed with:\n\n`--json-schema was provided but Claude did not return structured_output. Result subtype: success`\n\nReadTheDocs passed, and this PR does not add or change a user-facing API surface; it only normalizes the telemetry-only model name before constructing `EmbeddingUsageEvent`. I also tried rerunning the failed job, but GitHub returned: `run 28027084826 cannot be rerun; its workflow file may be broken`.

yinli-systems:
Closing this in favor of the earlier #2784, which implements the same safe .model / .model_name normalization across the synchronous and asynchronous telemetry call sites. My independent regression checks support that approach, so there is no value in asking maintainers to review a duplicate.

PR #2787, opened 2026-06-23
Harden multimodal image URL validation

### LINKED ISSUE #2754: [Security] SSRF bypass in Ragas
Dear VibrantLabsai development team,

I have discovered an SSRF bypass vulnerability in the _download_validate_and_encode function of the latest version of Ragas.

I understand that according to security.md, the correct vulnerability report should be sent via email, but after sending it, I found that the email address provided by your team seems invalid. I asked about this issue in the issues section a month ago but did not receive any response (https://github.com/vibrantlabsai/ragas/issues/2683), so I have no choice but to submit the report here.

The current `_download_validate_and_encode` function performs a security check on the host portion of the URL extracted by urlparse to prevent SSRF attacks.

However, there are indeed differences in parsing between urlparse and the library that actually sends the request. As you can see, the current execution order is to first extract the host for verification and then use requests.get to send the request.

Image

The core issue: `urlparse()` and `requests` disagree on which host a URL like `http://127.0.0.1:6666\@1.1.1.1` points to:

- `urlparse()` treats `\` as a regular character and `@` as the userinfo-host delimiter, so it extracts hostname as `1.1.1.1` (public)
- `requests` treats `\` as a path character, connecting to `127.0.0.1` (internal)

Below is a test code I wrote following the code.
```
from ragas.prompt.multi_modal_prompt import ImageTextPromptValue

url = "http://127.0.0.1:6666"
# url = "http://127.0.0.1:6666\@1.1.1.1"
ImageTextPromptValue(items=[])._download_validate_and_encode(url)
```
When an attacker uses `http://127.0.0.1:6666/`, the existing detection logic can detect that this is an internal network address and block it. 

Image

However, when an attacker uses `http://127.0.0.1:6666\@1.1.1.1`, the detection logic resolves the host to `1.1.1.1`, which is a public IP address, thus passing the verification. But in the actual request process, this URL is forwarded by requests.get to `http://127.0.0.1:6666/`, bypassing the detection and achieving an SSRF attack.

Image

yinli-systems:
## What

Harden multimodal image URL handling against ambiguous URL parsing before any outbound image fetch. This rejects URLs with userinfo or backslashes in the authority, including the `urlparse`/`requests` mismatch shape reported in #2754, and repeats the syntactic check inside the direct download helper.

The collections multimodal faithfulness image path now also performs the same syntactic guard and blocks hostnames that resolve to internal IP ranges before calling `requests.get`.

Fixes #2754.

## Tests

- `PYTHONPYCACHEPREFIX=/private/tmp/ragas-pycache /usr/bin/python3 -m py_compile src/ragas/prompt/multi_modal_prompt.py src/ragas/metrics/collections/multi_modal_faithfulness/util.py tests/unit/prompt/test_multi_modal_prompt_security.py`
- `UV_CACHE_DIR=/private/tmp/uv-cache-ragas UV_TOOL_DIR=/private/tmp/uv-tools-ragas uvx ruff check src/ragas/prompt/multi_modal_prompt.py src/ragas/metrics/collections/multi_modal_faithfulness/util.py tests/unit/prompt/test_multi_modal_prompt_security.py`
- `UV_CACHE_DIR=/private/tmp/uv-cache-ragas UV_TOOL_DIR=/private/tmp/uv-tools-ragas uvx ruff format --check src/ragas/prompt/multi_modal_prompt.py src/ragas/metrics/collections/multi_modal_faithfulness/util.py tests/unit/prompt/test_multi_modal_prompt_security.py`
- `UV_CACHE_DIR=/private/tmp/uv-cache-ragas UV_PROJECT_ENVIRONMENT=/private/tmp/ragas-uv-env PYTHONPATH=/private/tmp/ragas-site:src uv run --group dev pytest -q tests/unit/prompt/test_multi_modal_prompt_security.py`

Note: the pytest run used a temporary `/private/tmp/ragas-site/sitecustomize.py` shim for the existing main-branch `langchain_community.chat_models.vertexai` import break with current langchain-community. The shim only affects the local test process and is not part of this PR.

### DIFF
--- a/src/ragas/metrics/collections/multi_modal_faithfulness/util.py
+++ b/src/ragas/metrics/collections/multi_modal_faithfulness/util.py
@@ -2,9 +2,11 @@
 
 import base64
 import binascii
+import ipaddress
 import logging
 import os
 import re
+import socket
 import typing as t
 from io import BytesIO
 from urllib.parse import urlparse
@@ -23,6 +25,7 @@ DATA_URI_REGEX = re.compile(
     r"^data:(image\/(?:png|jpeg|gif|webp));base64,([a-zA-Z0-9+/=]+)$"
 )
 COMMON_IMAGE_EXTENSIONS = {".jpg", ".jpeg", ".png", ".gif", ".webp", ".bmp"}
+DISALLOWED_IP_CHECKS = {"is_loopback", "is_private", "is_link_local", "is_reserved"}
 
 
 class MultiModalFaithfulnessInput(BaseModel):
@@ -63,13 +66,15 @@ def is_image_path_or_url(item: str) -> bool:
     # Check for URL
     try:
         parsed = urlparse(item)
-        if parsed.scheme in ALLOWED_URL_SCHEMES:
+        if _is_allowed_image_url(item):
             path_part = parsed.path
             _, ext = os.path.splitext(path_part)
             if ext.lower() in COMMON_IMAGE_EXTENSIONS:
                 return True
             # Could be an image URL without extension
-            return True if parsed.scheme in ALLOWED_URL_SCHEMES else False
+            return True
+        if parsed.scheme:
+            return False
     except ValueError:
         pass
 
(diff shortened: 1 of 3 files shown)


yinli-systems:
The failing `check-docs` job appears to be the same Claude action infrastructure issue seen on nearby PRs, not a documentation failure from this change. The failed step ends with:

`--json-schema was provided but Claude did not return structured_output. Result subtype: success`

This PR changes internal URL validation around multimodal image fetching and includes docstrings for the new helpers; it does not add a new user-facing API or usage pattern. Local validation passed as listed in the PR description.

PR #2788, opened 2026-06-24
Fix InstructorLLM reasoning model parameter mapping

### LINKED ISSUE #2708: `InstructorLLM._map_openai_params()` fails to remap `max_tokens` for OpenAI models with dotted versions
**Problem**

`InstructorLLM._map_openai_params()` in `src/ragas/llms/base.py` is intended to detect OpenAI/Azure reasoning models and remap `max_tokens` to `max_completion_tokens`.

In practice, the detection logic is too brittle and misses valid reasoning model names. When that happens, `max_tokens` is sent unchanged to `client.chat.completions.create(...)`, and OpenAI rejects the request with:

```python
openai.BadRequestError: Error code: 400 - {'error': {'message': "Unsupported parameter: 'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead.", 'type': 'invalid_request_error', 'param': 'max_tokens', 'code': 'unsupported_parameter'}}
```

This makes structured output via `InstructorLLM` fail for some supported OpenAI reasoning models.

**Why this happens**

The current implementation relies on ad hoc string parsing:

- O-series detection only supports `o1` through `o9`
- GPT detection only supports integer versions `gpt-5` through `gpt-19`
- dotted versions like `gpt-5.1` are not handled

https://github.com/vibrantlabsai/ragas/blob/298b68274234c060deacab3cf5fb52aa3a20e885/src/ragas/llms/base.py#L887

As a result, valid reasoning models can bypass the remapping logic entirely, leaving `max_tokens` in the request.

**Impact**

- `llm_factory(..., adapter="instructor")` and `InstructorLLM(InstructorBaseRagasLLM)` and  can fail with OpenAI 400s for valid reasoning models
- Azure users are especially exposed because of deployment names.

**Current problematic code**

In `src/ragas/llms/base.py`, `_map_openai_params()`:

```python
if model_str.startswith("gpt-"):
    version_str = (
        model_str[4:].split("-")[0].split("_")[0]
    )  # Get version number
    try:
        version = int(version_str)
        if 5 <= version <= 19:
            return True
    except ValueError:
        pass
```

That leads directly to the OpenAI error above.

**Proposed resolution**

Urgent short-term fix:
- replace the current reasoning-model detection with more robust matching
- support:
  - `o` models with multi-digit versions, e.g. `o10`
  - GPT models with dotted versions, e.g. `gpt-5.1`
  - suffixed variants like `gpt-5-mini`, `gpt-5.1-mini`, `o3-mini`
- add tests covering the failing model-name patterns

Ragas version: 0.4.3

yinli-systems:
## Summary

Fix `InstructorLLM._map_openai_params()` reasoning-model detection so OpenAI/Azure reasoning model names with dotted versions or multi-digit o-series versions remap `max_tokens` to `max_completion_tokens`.

This covers model names such as:

- `gpt-5.1`
- `gpt-5.1-mini`
- `o10`
- `o10-mini`

while keeping legacy model names such as `gpt-4o` and `gpt-4.1` on the existing `max_tokens` path.

Fixes #2708.

## Root cause

The previous detection parsed only integer GPT versions and only single-digit o-series names. Dotted GPT versions like `gpt-5.1` raised `ValueError` during integer parsing and skipped the reasoning-model parameter mapping entirely.

## Tests

- `UV_CACHE_DIR=/Users/yinxiaogou/Documents/resume/.tmp-uv-cache uv run ruff check src/ragas/llms/base.py tests/unit/llms/test_instructor_factory.py`
- `UV_CACHE_DIR=/Users/yinxiaogou/Documents/resume/.tmp-uv-cache uv run ruff format --check src/ragas/llms/base.py tests/unit/llms/test_instructor_factory.py`
- `UV_CACHE_DIR=/Users/yinxiaogou/Documents/resume/.tmp-uv-cache PYTHONPATH=src uv run --with 'langchain-community==0.3.27' --with 'langchain-core==0.3.75' --with 'langchain-openai==0.3.29' pytest tests/unit/llms/test_instructor_factory.py`

The temporary LangChain pins are only for local test execution. The default local dependency resolution currently hits the existing VertexAI/LangChain import incompatibility tracked separately in #2753.

### DIFF
--- a/src/ragas/llms/base.py
+++ b/src/ragas/llms/base.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import asyncio
 import inspect
 import logging
+import re
 import threading
 import typing as t
 from abc import ABC, abstractmethod
@@ -859,8 +860,8 @@ class InstructorLLM(InstructorBaseRagasLLM):
         - If structured output is truncated, increase max_tokens further
 
         Pattern-based matching for future-proof coverage:
-        - O-series: o1, o2, o3, o4, o5, ... (all reasoning versions)
-        - GPT-5 series: gpt-5, gpt-5-*, gpt-6, gpt-7, ... (all GPT-5+ models)
+        - O-series: o1, o2, o3, o10, ... (all reasoning versions)
+        - GPT-5 series: gpt-5, gpt-5.1, gpt-5-*, gpt-6, ... (all GPT-5+ models)
         - Other: codex-mini
         """
         mapped_args = self.model_args.copy()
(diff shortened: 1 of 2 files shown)


yinli-systems:
The failing `check-docs` job appears to be an infrastructure/action failure rather than a documentation issue in this PR. The action installed Claude Code successfully, then failed with `--json-schema was provided but Claude did not return structured_output. Result subtype: success` before producing a docs decision. ReadTheDocs passed for this PR.

yinli-systems:
Closing this in favor of #2709, which addresses the same issue and has already received approval. I independently verified the dotted GPT-5 and multi-digit o-series cases covered here, but keeping a second implementation open would only split reviewer attention.

PR #2811, opened 2026-07-04
feat(metrics): add CitationFaithfulness collections metric (citation-attribution faithfulness)

rsrijith:
## What

Adds `CitationFaithfulness`, a modern `collections/` metric that scores whether each inline `[N]` citation in a response points to the specific retrieved passage that supports the claim it is attached to.

Unlike existing metrics, it catches **supported-elsewhere misattribution**: a claim cited to passage `[A]` that `[A]` does not support, even though another retrieved passage `[B]` does.

- `Faithfulness` checks whether a claim is supported by the contexts *anywhere* (ignores which passage was cited).
- `QuotedSpansAlignment` checks whether verbatim quoted spans appear *somewhere* in the sources.
- `CitationFaithfulness` checks the *attribution*: does the cited passage support the specific claim it is attached to.

## How

- Implemented in `src/ragas/metrics/collections/citation_faithfulness/` following the `collections/faithfulness/` pattern (`BaseMetric`, `BasePrompt.to_string` + `llm.agenerate`), i.e. the non-deprecated path — not the legacy `ragas.metrics` module.
- Two-step LLM pipeline: (1) extract cited claims with their `[N]` indices; (2) verify each cited passage supports its specific claim. The step-2 prompt explicitly refuses "supported elsewhere" credit.
- Pure, LLM-free helpers (`has_citation_markers`, `parse_citation_markers`, `compute_citation_score`) live in `util.py` and are unit-tested in isolation, mirroring `quoted_spans/util.py`.
- 1-based indexing (`[1]` → `retrieved_contexts[0]`); out-of-range citations count as incorrect and are never sent to the LLM; score = fraction of cited claims whose citations are all correct; `NaN` for the no-citation degenerate case (consistent with `Faithfulness`).
- Registered in `collections/__init__.py`.

## Usage

```python
from ragas.metrics.collections import CitationFaithfulness
from ragas.llms.base import llm_factory
from openai import AsyncOpenAI

metric = CitationFaithfulness(llm=llm_factory("gpt-4o-mini", client=AsyncOpenAI()))
result = await metric.ascore(
    user_input="Tell me about Paris.",
    response="Paris is the capital of France [1]. The Eiffel Tower is in Paris [2].",
    retrieved_contexts=["Paris is the capital of France.", "The Eiffel Tower is located in Paris."],
)
print(result.value)  # 1.0
```

## Tests

`tests/unit/test_citation_faithfulness_collections.py` — 31 offline unit tests (mocked instructor LLM, no API key): pure-helper tests, prompt/model round-trips, correct attribution (1.0), supported-elsewhere misattribution (0.5), out-of-range citation handling, both NaN degenerate paths, and input validation. `ruff check` and `ruff format --check` pass on the new files.


rsrijith:
Checking in on this one — it's been open since early July with no human review yet.

Worth flagging: the full test suite has never actually run on this PR. Only `check-docs` and the Read the Docs build have executed (both pass); the main CI workflows appear to be gated on maintainer approval for outside contributors, so the PR sits at `UNSTABLE` purely because those checks were never triggered rather than because anything failed.

If a maintainer could approve the workflow run, I'm happy to fix whatever it surfaces.

Quick recap of what's here: `CitationFaithfulness` implemented in the non-deprecated `collections/` path, checking whether each `[N]` citation marker points to the passage that actually supports the claim it's attached to — distinct from `Faithfulness` (which checks support anywhere in the context) and from `QuotedSpansAlignment` (which checks quote fidelity, not attribution). The gap it targets is the misattribution case: a claim cited to passage `[A]` that `[A]` doesn't support, even though some other retrieved passage would.

Happy to rebase, split it up, or adjust naming/scope if that helps it land.

PR #2812, opened 2026-07-05
fix: handle swarm assistant messages without tool calls

### LINKED ISSUE #2813: Swarm converter raises KeyError for assistant messages without tool_calls
## Problem

`convert_to_ragas_messages()` raises `KeyError: 'tool_calls'` when converting a normal Swarm assistant message that does not include tool calls.

A plain assistant response without tool calls can contain only `role` and `content`:

```python
from ragas.integrations.swarm import convert_to_ragas_messages

convert_to_ragas_messages([
    {role: assistant, content: hi}
])
```

## Current behavior

The assistant branch accesses `message[tool_calls]` directly, so the conversion fails before a Ragas `AIMessage` can be created:

```text
KeyError: 'tool_calls'
```

## Expected behavior

No-tool assistant messages should convert to a normal `AIMessage` with empty tool calls:

```python
AIMessage(content=hi, tool_calls=[])
```

## Impact

Swarm traces with regular assistant turns that do not call tools cannot be converted unless callers add an explicit empty `tool_calls` field themselves.

A fix is proposed in #2812.

VectorPeak:
﻿## Issue Link / Problem Description
- Fixes #2813

### What Problem This Solves

convert_to_ragas_messages() currently treats 	ool_calls as a required key for every Swarm assistant message:

`python
message["tool_calls"]
`

That makes no-tool assistant responses fail during conversion, even though a plain assistant message with only 
ole and content is a normal Swarm-style message shape when the assistant did not call any tool:

`python
[{"role": "assistant", "content": "hi"}]
`

In that case, the expected behavior is to convert the message into a Ragas AIMessage with empty tool calls. Instead, the current implementation raises KeyError: 'tool_calls' before the message can be converted.

### Evidence

A minimal reproduction is enough to trigger the failure before this PR:

`python
from ragas.integrations.swarm import convert_to_ragas_messages

convert_to_ragas_messages([
    {"role": "assistant", "content": "hi"}
])
`

Observed behavior before this change:

`	ext
KeyError: 'tool_calls'
`

Expected behavior after this change:

`	ext
AIMessage(content="hi", tool_calls=[])
`

The local focused reproduction used the same payload and verified that the converter now returns an AIMessage with content == "hi" and 	ool_calls == [].

### Possible call chain / impact

`	ext
Swarm conversation trace
  -> assistant message without a tool call
    -> convert_to_ragas_messages(...)
      -> handle_assistant_message(...)
        -> direct access to message["tool_calls"]
        -> KeyError before creating Ragas AIMessage
`

This affects Swarm traces where an assistant turn responds normally without invoking a tool. Those traces are valid no-tool assistant outputs, but they currently cannot be converted into Ragas messages unless callers add an explicit empty 	ool_calls field themselves.

This PR only changes the missing/falsy 	ool_calls path. Existing assistant messages that include tool calls still go through the same convert_tool_calls(...) conversion logic.

## Changes Made
- Treat missing or falsy assistant 	ool_calls as an empty list.
- Preserve existing tool-call conversion behavior when 	ool_calls is present.
- Add a focused unit test for an assistant message without tool calls.

### How to Test
- [x] Manual testing steps:
  1. uv run --isolated --no-project --with ruff ruff format --check src/ragas/integrations/swarm.py tests/unit/test_swarm.py --config pyproject.toml
  2. uv run --isolated --no-project --with ruff ruff check src/ragas/integrations/swarm.py tests/unit/test_swarm.py --config pyproject.toml
  3. Minimal importlib reproduction: convert_to_ragas_messages([{"role": "assistant", "content": "hi"}]) now returns AIMessage(content="hi", tool_calls=[]).

Focused pytest tests/unit/test_swarm.py -q was attempted, but the local editable test environment currently fails before this test is collected because importing 
agas hits an existing dependency import issue: ModuleNotFoundError: No module named 'langchain_community.chat_models.vertexai'. The behavior changed by this PR was verified with a targeted importlib reproduction that loads only 
agas.messages and 
agas.integrations.swarm.

## References
- Related issues: #2813
- Documentation:
- External references:

No screenshot is applicable because this is a message-conversion bug. The smallest example is a plain Swarm assistant message without tool calls:

`python
from ragas.integrations.swarm import convert_to_ragas_messages

messages = convert_to_ragas_messages([
    {"role": "assistant", "content": "hi"}
])
`

Before this PR, the converter failed before producing a Ragas message:

`	ext
KeyError: 'tool_calls'
`

After this PR, the same input is converted as a normal assistant message with no tool calls:

`python
AIMessage(content="hi", tool_calls=[])
`

Assistant messages that do include 	ool_calls still use the existing tool-call conversion path.


VectorPeak:
Closing this for now since it has been stale without maintainer signal. Happy to revisit if this is still useful for the project.

Issue #1955, opened 2025-03-07
Enforcing json format


Leahh02:
- [x] I checked the [documentation](https://docs.ragas.io/) and related resources and couldn't find an answer to my question.

**My Question**
I'm getting exception 
```
 File "/my/path/to/ragas/prompt/pydantic_prompt.py", line 423, in parse_output_string
    raise RagasOutputParserException()
ragas.exceptions.RagasOutputParserException: The output parser failed to parse the output including retries.
```
And I'm pretty sure it's happening because the LLM is getting cut off when it's generating the answer in the JSON format. Right now I have the max tokens for my LLM to generate set to 1024. I can increase the max tokens to help avoid that problem. But I'm wondering what happens if the LLM has a longer generation and still runs into that max token count before it finishes the JSON structure. I know the prompt says to follow the JSON schema, so there's an effort to make sure that doesn't happen, but could there be a more concrete way for RAGAS to make sure that JSON format is finished? Or is that just too complicated?

I might also decrease my temperature and tweak my top_p to decrease the likelihood of that over generation. Maybe there could be a reminder/guidance on settings like that somewhere in the documentation. That might give people some trouble shooting ideas and a reminder that tweaking those settings might give them better performance.

**Code Examples**
After line 201 in pydantic_prompt.py I'm printing out the output string to see what the LLM generates. That's where I saw that the JSON schema wasn't getting finished.
Here's a shortened example of what it looks like:
```
 {
    "statements": [
        {
            "statement": "The crash of Yeti Airlines Flight 691 in Pokhara, Nepal was caused by a technical issue with the plane.",
            "reason": "The context explicitly states that the crash was caused by the aircraft's propellers being feathered, which is a technical issue.",
            "verdict": 1
        },
        {
            "statement": "The weather was clear at the time of the crash.",
            "reason": "There is no"
```

PR #2817, opened 2026-07-06
fix(prompt): use non-zero temperature when retrying to fix malformed output

Mayan-kr:
### Problem

When an LLM's output fails to parse, `RagasOutputParser.parse_output_string`
re-prompts the model with `fix_output_format_prompt` to correct it. That retry
call did not pass a `temperature`, so it fell back to the deterministic default
from `BaseRagasLLM.get_temperature(n=1)` → `0.01`.

The consequence: if the first attempt produced malformed output at a low
temperature, the "fix" attempt runs the model at the *same* near-zero
temperature and gets back the *same* malformed output. Every retry is wasted and
evaluation fails with:

    RagasOutputParserException: The output parser failed to parse the output including retries.

This is a recurring pain point with providers/models that are effectively
deterministic at low temperature (e.g. Groq, local LLMs). Related reports:
#1997, #1955.

### Fix

Pass an explicit non-zero temperature (`FIX_OUTPUT_FORMAT_TEMPERATURE = 0.7`) to
the `fix_output_format_prompt.generate(...)` retry so the self-correction attempt
can produce different, valid output. The **initial** generation is unchanged and
stays deterministic — only the fix/retry path becomes non-deterministic, which is
exactly what that path is for.

### Testing

Added `test_prompt_fix_output_uses_nonzero_temperature`, a regression test using
a recording mock LLM that always returns invalid JSON. It asserts the initial
generation uses `0.01` while every fix-output retry uses
`FIX_OUTPUT_FORMAT_TEMPERATURE`. The test fails on `main` and passes with this
change. `ruff format`/`ruff check` clean; adjacent `tests/unit/prompt` and
`tests/unit/llms` still pass (116 tests).

### Notes

No public API change; `FIX_OUTPUT_FORMAT_TEMPERATURE` is a module-level constant.
A natural follow-up would be to *escalate* temperature across successive retries
and include the last raw output in `RagasOutputParserException`.


Mayan-kr:
Quick summary for reviewers: this fixes a wasted-retry loop where the JSON self-correction step reused the deterministic default temperature (0.01), so deterministic models re-emitted the same malformed output on every retry. The fix only changes the retry path; initial generation stays deterministic. Regression test included. Happy to adjust the constant or switch to escalating temperature across retries if you'd prefer.

PR #2852, opened 2026-07-14
fix(analytics): coerce non-string model attributes before telemetry dispatch

### LINKED ISSUE #2851: fix(analytics): centralise safe model-attribute coercion for EmbeddingUsageEvent
### Problem

`EmbeddingUsageEvent` is typed with `model: Optional[str]`, but several call sites pass the value of `getattr(..., "model", None)` where the wrapped object can expose a non-string model — a loaded `SentenceTransformer` or `CrossEncoder` (HuggingFace's `langchain_community.embeddings.huggingface.HuggingFaceEmbeddings`), a `FastEmbedEmbeddings` instance (the most common user case in issue #2783), or a custom embedding wrapper. Pydantic's `ValidationError` then fires inside `track(...)` and is raised *before* the `@silent` decorator can swallow it, so a `LangchainEmbeddingsWrapper.embed_query` / `embed_documents` / `aembed_query` / `aembed_documents` call returns `NaN` to the calling metric while telemetry writes the error to the logs.

Today, four open PRs (#2784, #2786, #2789, #2838) all try to fix this on `LangchainEmbeddingsWrapper` only. They patch the `model=` keyword but each in a slightly different way, and none of them protects the other `track(...)` call sites in the same module — or any future embedding provider. We already saw #2783 in production: a user with `FastEmbedEmbeddings` (which stores the loaded model *object* on `.model`) crashes silently on every embedding call.

The same unsafe pattern likely exists in the four `ModernBaseEmbeddings` providers (`openai_provider.py`, `google_provider.py`, `litellm_provider.py`, `huggingface_provider.py`) wherever they read an inner-model attribute and forward it to `track(...)`.

### Proposed solution

Add a single helper in `src/ragas/_analytics.py`:

```python
def _safe_model_attr(model_obj: t.Any) -> t.Optional[str]:
    """Coerce a possibly-non-string model attribute into a safe string.

    Returns ``None`` when the attribute is missing, ``None`` itself, or
    not string-coercible in any safe way; otherwise returns ``str(model_obj)``.
    Used by every ``track(...)`` site so a wrapped model object (e.g. a loaded
    SentenceTransformer or FastEmbedEmbeddings instance) cannot crash the
    ``EmbeddingUsageEvent`` model validator.
    """
    if model_obj is None or isinstance(model_obj, str):
        return model_obj
    return str(model_obj)
```

Route every embedding call site that currently passes `model=...` to `track(...)` through this helper. The conversion is one line per site; the helper itself is the single coercion point.

### Alternatives considered

- **Defensive coercion at every call site individually.** Duplicated code, easy to forget on new providers, and the central issue (#2783) would re-open every time a new wrapper ships.
- **Pydantic validator on the event itself.** A `field_validator` could coerce the value, but the public model contract is `Optional[str]`; relaxing that to `Any` would change user-facing types and break consumers that introspect the field.
- **Skip telemetry on non-string models.** Drops signal; maintainers lose visibility into the providers users are using.

### Scope

- One helper added to `src/ragas/_analytics.py` (no new module, no public API change).
- One-line call-site rewrites in `src/ragas/embeddings/base.py` (four sites) and any matching sites in the four `*_provider.py` files.
- A regression test in `tests/unit/test_analytics.py` covering string, `None`, missing attribute, int, and a custom object with `__str__`.

### Acceptance criteria

- Calling `track(EmbeddingUsageEvent(model=_safe_model_attr(<SentenceTransformer instance>), ...))` produces a valid event with `model="SentenceTransformer(...)"` rather than a `ValidationError`.
- The four call sites in `LangchainEmbeddingsWrapper` (and any future provider) all use the helper, so adding a new provider that reads a wrapped `.model` object cannot reintroduce the bug.
- Existing `tests/unit/test_analytics.py` continues to pass.

### Backward compatibility

The helper only relaxes non-string values. Callers that pass a string model (the most common case) see byte-for-byte identical behaviour. Callers that previously crashed get a stringified model name in telemetry.

### Risks

- If a wrapped model object returns an unhelpful `__str__`, telemetry will record something like `<TextEmbedding ...>` — but the user's metric is no longer NaN, which is the trade-off.
- Existing PRs (#2784, #2786, #2789, #2838) on issue #2783 may conflict with the call-site rewrites. Mitigation: this PR keeps the public model contract unchanged, so the conflict reduces to literal duplication of the helper call at the same lines; easy to resolve on the maintainer's side when picking one approach.

Harsh23Kashyap:
### Summary

`EmbeddingUsageEvent.model` is typed `Optional[str]`, but several call sites pass the value of `getattr(self.embeddings, "model", None)` where the wrapped object can expose a non-string model — a loaded `SentenceTransformer` / `CrossEncoder`, a `FastEmbedEmbeddings` instance (the most common user case in #2783), or a custom embedding wrapper. Pydantic's `ValidationError` fires inside `track(...)` *before* the `@silent` decorator can swallow it, silently turning embedding calls into NaN.

This PR introduces a single, central coercion point — `_safe_model_attr(model_obj)` in `src/ragas/_analytics.py` — and routes the four `track(...)` call sites in `LangchainEmbeddingsWrapper` through it. Any future provider that reads a wrapped `.model` attribute becomes protected by default.

### What's changing

- `src/ragas/_analytics.py` — new helper `_safe_model_attr(model_obj)` that returns `None` when the attribute is missing or `None`, the value unchanged when it's already a `str`, and `str(model_obj)` otherwise.
- `src/ragas/embeddings/base.py` — four `track(EmbeddingUsageEvent(...))` call sites in `LangchainEmbeddingsWrapper.embed_query` / `embed_documents` / `aembed_query` / `aembed_documents` route the `model` argument through the helper.
- `tests/unit/test_analytics_safe_model_attr.py` — new regression test covering string / `None` / int / float / object / list / dict / missing-attribute behaviour, identity preservation for strings, end-to-end routing through `LangchainEmbeddingsWrapper` for a non-string model attribute, and string-passthrough parity for existing callers.

### Why this is the right shape

The four open PRs on #2783 (#2784, #2786, #2789, #2838) each patch only one site in `LangchainEmbeddingsWrapper`. This helper complements them: once maintainers pick one approach, the call-site rewrites can be consolidated with a one-line change at every site rather than copy-pasted.

### Verification

- `make run-ci-format-check` — clean
- `uv run pyright src/ragas/_analytics.py src/ragas/embeddings/base.py` — only the pre-existing `HuggingfaceEmbeddings.embed_documents` `Tensor` warning (unrelated)
- New tests: `tests/unit/test_analytics_safe_model_attr.py` — 11 cases pass
- Existing `tests/unit/test_analytics.py` — all pass
- `tests/unit/test_embeddings.py` + `tests/unit/test_embeddings_caching.py` — all pass

### Backward compatibility

The helper only relaxes non-string values. Callers that pass a string model (the most common case) see byte-identical behaviour. Callers that previously crashed get a stringified model name in telemetry.

### Out of scope

- The four competing PRs on #2783 (#2784, #2786, #2789, #2838) are intentionally untouched. This PR complements them rather than replacing them; the maintainer's eventual pick can rebase the call-site rewrites onto one common helper.

Fixes #2851


Harsh23Kashyap:
Can someone take a look when you have a chance?

Harsh23Kashyap:
All review threads resolved, CI green, no new maintainer comments. model attr coerce.

Harsh23Kashyap:
All review threads resolved, CI green, no new maintainer comments. model attr coerce. Standing by.

Harsh23Kashyap:
Still 0 behind upstream. model attr coerce. All review threads resolved, CI green. Standing by.

PR #2861, opened 2026-07-18
fix(backends): LocalCSVBackend crashes on rows with different keys

### LINKED ISSUE #2860: LocalCSVBackend.save_dataset crashes with ValueError on rows with different keys
[x] I have checked the documentation and related resources and couldn't resolve my bug.

**Describe the bug**

`LocalCSVBackend.save_dataset()` (and `save_experiment()`) crash with `ValueError: dict contains fields not in fieldnames: '<key>'` when the rows being saved don't all share the exact same set of keys.

`Dataset.append()` accepts arbitrary dicts when the dataset has no `data_model` configured (see `ragas/dataset.py`), so nothing prevents a caller from building up a list of rows where, say, only some rows carry an optional `reason`/`error` field (a very plausible pattern for `@experiment()` results where extra diagnostic fields get attached conditionally). `LocalCSVBackend._save()` picks the CSV header from only the first row:

```python
fieldnames = data[0].keys()
writer = csv.DictWriter(f, fieldnames=fieldnames)
writer.writeheader()
writer.writerows(data)
```

`csv.DictWriter` defaults to `extrasaction="raise"`, so any later row containing a key absent from the first row's keys raises instead of being saved. The dataset save (and the whole experiment run, since `save()` is typically called at the end of the run) fails entirely.

Ragas version: main branch (post v0.4.3, `src/ragas/backends/local_csv.py`)
Python version: 3.12

**Code to Reproduce**

```python
import tempfile
from ragas.backends.local_csv import LocalCSVBackend

with tempfile.TemporaryDirectory() as tmp:
    backend = LocalCSVBackend(root_dir=tmp)
    data = [
        {"question": "capital of France?", "response": "Paris", "score": 1.0},
        {"question": "capital of UK?", "response": "London", "score": 0.0, "reason": "wrong span matched"},
    ]
    backend.save_dataset("demo", data, data_model=None)
```

**Error trace**

```
  ...
  File "ragas/backends/local_csv.py", line 94, in _save
    writer.writerows(data)
  File ".../csv.py", line 167, in writerows
    return self.writer.writerows(map(self._dict_to_list, rowdicts))
  File ".../csv.py", line 159, in _dict_to_list
    raise ValueError("dict contains fields not in fieldnames: "
ValueError: dict contains fields not in fieldnames: 'reason'
```

**Expected behavior**

Rows with a superset of the first row's keys should save successfully, with missing values round-tripping as empty strings (the way `csv.DictWriter`'s `restval` is designed to be used) instead of the whole save call blowing up.

**Additional context**

I have a fix + regression test ready (compute the union of keys across every row instead of just the first row's keys before constructing the `DictWriter`) and will open a PR referencing this issue.

ErenAta16:
- Fixes #2860

`LocalCSVBackend.save_dataset()`/`save_experiment()` crash with `ValueError: dict contains fields not in fieldnames: '<key>'` whenever the rows being saved don't all share the exact same set of keys. `Dataset.append()` accepts arbitrary dicts when no `data_model` is configured, so nothing stops a caller from mixing rows where only some carry an optional field (e.g. an `@experiment()` row that only attaches a `reason`/`error` field on failures).

## Changes Made
- `src/ragas/backends/local_csv.py`: `_save()` picked CSV fieldnames from `data[0].keys()` only. `csv.DictWriter` defaults to `extrasaction="raise"`, so any later row with a key the first row didn't have blew up the entire save. Changed to collect the union of keys across every row (first-seen order preserved) before constructing the `DictWriter`, so rows missing a key round-trip as an empty string via `DictWriter`'s default `restval` instead of crashing.
- `tests/unit/backends/test_local_csv.py`: added `test_heterogeneous_row_keys`.

### How to Test
- [ ] Manual testing steps:
  1. `pytest tests/unit/backends/test_local_csv.py -v`
  2. Confirm `test_heterogeneous_row_keys` passes and all other backend tests are unaffected (65 passed, 21 skipped locally for the whole `tests/unit/backends/` directory on this branch)

I verified this with the actual RED/GREEN cycle: reverted just the source-file change (`git stash`) with the new test present, confirmed `test_heterogeneous_row_keys` fails with the exact `ValueError` above on unmodified code, then restored the fix and confirmed it passes.

## References
- Related issues: #2860

```python
import tempfile
from ragas.backends.local_csv import LocalCSVBackend

with tempfile.TemporaryDirectory() as tmp:
    backend = LocalCSVBackend(root_dir=tmp)
    data = [
        {"question": "capital of France?", "response": "Paris", "score": 1.0},
        {"question": "capital of UK?", "response": "London", "score": 0.0, "reason": "wrong span matched"},
    ]
    backend.save_dataset("demo", data, data_model=None)  # used to raise ValueError, now saves fine
    loaded = backend.load_dataset("demo")
    assert loaded[0]["reason"] == ""
    assert loaded[1]["reason"] == "wrong span matched"
```


ErenAta16:
Worth noting for whoever picks this up: @Harsh23Kashyap hit the same crash independently and wrote a fix in #2915 before finding this PR, then closed theirs. So the diagnosis has been reproduced by two people who arrived at it separately, and both landed on the same fix, the union of keys across all rows in first-seen order rather than `data[0].keys()`.

The one visible difference between the two is that #2915 passed `restval=""` explicitly. That is a no-op, since it is already the default:

```
>>> inspect.signature(csv.DictWriter.__init__)
(self, f, fieldnames, restval='', extrasaction='raise', dialect='excel', *args, **kwds)
```

Writing the same two heterogeneous rows both ways produces byte-identical output:

```
question,response,score,reason
capital of France?,Paris,1.0,
capital of UK?,London,0.0,wrong span matched
```

and round-trips through `csv.DictReader` with `row0["reason"] == ''`. So there is nothing in #2915 that this is missing; no need to compare them line by line.

@Harsh23Kashyap, thanks for stepping back rather than pushing a competing PR, and sorry you spent the effort. If you want your `TestHeterogeneousRows` cases added here I am happy to pull them in with attribution; yours cover a couple of shapes mine does not.

Still `MERGEABLE` against current main.

PR #2867, opened 2026-07-20
fix: align NonLLMContextRecall threshold to >= for consistency with sibling metric

### LINKED ISSUE #2777: NonLLMContextRecall and NonLLMContextPrecisionWithReference threshold relevance inconsistently (> vs >=)

Varshith-Kali:
Fixes #2777

NonLLMContextRecall and NonLLMContextPrecisionWithReference share the same default \	hreshold=0.5\ and the same NonLLMStringSimilarity distance measure, but apply the threshold with different comparison operators:

- \_context_recall.py:219\ — \1 if score > self.threshold\
- \_context_precision.py:234\ — \1 if score >= self.threshold\

A context whose similarity score is exactly at the threshold is counted as relevant by precision but not by recall. \git blame\ shows the two operators were introduced by the same author in two commits ~50 seconds apart, which suggests an inadvertent inconsistency rather than an intended semantic difference.

## What changed

- Changed \NonLLMContextRecall._compute_score\ from \>\ to \>=\
- Added 5 unit tests covering the boundary case, below/above/below threshold, mixed scores, and custom threshold values

## Files

\src/ragas/metrics/_context_recall.py\ — one-character fix
\	ests/unit/metrics/test_nonllm_threshold_consistency.py\ — new test file


Varshith-Kali:
Closing this PR to consolidate with the related #2868. Will resubmit a single combined fix if capacity allows. Thanks.

PR #2868, opened 2026-07-20
fix: use correct TP for recall direction in FactualCorrectness

### LINKED ISSUE #2693: FactualCorrectness(mode="recall") returns precision score instead of recall

Varshith-Kali:
Fixes #2693

FactualCorrectness._single_turn_ascore runs two NLI tasks:
- `reference_response`: decompose RESPONSE, verify against REFERENCE (precision direction)
- `response_reference`: decompose REFERENCE, verify against RESPONSE (recall direction)

The bug: for all three modes (precision/recall/f1), the code used `sum(reference_response)` as TP. In recall mode, this is the **precision** TP (number of response claims supported by reference), not the recall TP (number of reference claims supported by response).

## Example from the issue

- Response: 4 claims, 3 supported by reference -> precision TP=3, FP=1
- Reference: 3 claims, 2 supported by response -> recall TP=2, FN=1
- Expected recall: 2/(2+1) = 0.67
- Actual (bug): 3/(3+1) = 0.75

## What changed

- `_factual_correctness.py`: separated `tp_precision` and `tp_recall`, use `tp_recall` for recall mode, pass both to `fbeta_score` for f1
- `collections/factual_correctness/metric.py`: same fix in the v2 module (had the identical bug)
- `utils.py`: added optional `tp_recall` kwarg to `fbeta_score` - fully backward-compatible, all existing callers work unchanged

## Files

- `src/ragas/metrics/_factual_correctness.py`
- `src/ragas/metrics/collections/factual_correctness/metric.py`
- `src/ragas/metrics/utils.py`


Varshith-Kali:
Closing this PR to manage my open PR count. May revisit with a consolidated fix in the future. Thanks.

PR #2873, opened 2026-07-21
fix(testset): raise clear ValueError for empty documents in default_transforms

### LINKED ISSUE #2871: default_transforms() raises unhandled ZeroDivisionError on an empty documents list
## Describe the bug

`ragas.testset.transforms.default.default_transforms()` computes a document-length histogram and divides each bin count by `len(documents)` before validating that `documents` is non-empty:

```python
bin_ranges = [(0, 100), (101, 500), (501, float("inf"))]
result = count_doc_length_bins(documents, bin_ranges)
result = {k: v / len(documents) for k, v in result.items()}
```

If `documents` is an empty list, this raises a raw `ZeroDivisionError: division by zero` instead of a clear, actionable message.

This is reachable from the public API: `TestsetGenerator.generate_with_langchain_docs()` and `generate_with_llamaindex_docs()` both call `default_transforms(documents=..., ...)` whenever the caller doesn't pass an explicit `transforms` argument, so passing an empty `documents` list (or a list that resolves to empty after upstream filtering) crashes with this opaque error.

Note the function already validates one bad-input case explicitly — "too short" documents raise a clear `ValueError`:
```python
else:
    raise ValueError(
        "Documents appears to be too short (ie 100 tokens or less). Please provide longer documents."
    )
```
The empty-list case just isn't covered by that same validation, since the crash happens earlier, during the bin-count division.

## Ragas version

## Reproduction
```python
from ragas.testset.transforms.default import default_transforms

default_transforms(documents=[], llm=my_llm, embedding_model=my_embeddings)
# ZeroDivisionError: division by zero
```

## Expected behavior
A clear `ValueError` (matching the existing "too short" validation pattern), e.g. "The documents list is empty. Please provide at least one document to generate default transforms for."

mittalpk:
Fixes #2871

## What
`default_transforms()` divided by `len(documents)` before validating that `documents` was non-empty, so an empty list raised an opaque `ZeroDivisionError: division by zero` instead of a clear error.

This is reachable from the public API: `TestsetGenerator.generate_with_langchain_docs()` and `generate_with_llamaindex_docs()` both call `default_transforms(documents=..., ...)` whenever no explicit `transforms` is passed, so an empty (or emptied-by-filtering) `documents` list crashes with this unhelpful error.

## Fix
Added an explicit empty-list check that raises a clear `ValueError`, following the exact pattern the function already uses for its other input-validation case ("documents appears to be too short").

## Tests
New `tests/unit/test_default_transforms.py`:
- `test_default_transforms_empty_documents_raises_clear_error` — confirmed to fail with `ZeroDivisionError` against pre-fix code (`git stash`), passes post-fix with a clear `ValueError`.
- `test_default_transforms_too_short_documents_raises_value_error` — regression check that the existing "too short" validation still works unchanged.
- `test_default_transforms_with_long_documents_returns_transforms` — sanity check that the normal/happy path is untouched.

## Verification
- `uv run ruff check` / `ruff format --check` on changed files: clean
- `uv run pyright` on changed files: 0 errors
- New tests: 3/3 passing
- Pre-commit hook (`monorepo-ci` format check): passed


mittalpk:
Checking in on this — let me know if it needs any changes to get reviewed.

mittalpk:
Same check-docs failure as #2874 -- it's a pull_request_target workflow and GitHub's own actions/checkout refuses to fetch fork PR code in that context by default (a security guard against "pwn request" attacks) unless allow-unsafe-pr-checkout: true is set in the workflow file. That's a maintainer-side config change, not anything in this PR's content.

anishpulsay:
Hi! I'd like to work on this issue as my first contribution to Ragas. Could you please assign it to me? I'd be happy to submit a PR with the fix and regression tests. Thanks!

PR #2874, opened 2026-07-21
fix(prompt): top_k=0 must return no examples in SimpleInMemoryExampleStore

### LINKED ISSUE #2872: DynamicFewShotPrompt: max_similar_examples=0 (top_k=0) returns all examples instead of none
## Describe the bug

`SimpleInMemoryExampleStore.get_examples(data, top_k=0, ...)` is meant to return at most `top_k` examples, but `top_k=0` returns **every stored example** instead of none, in both the recency-based path (no embedding model) and the cosine-similarity path (with an embedding model).

Root cause: Python's negative-slice quirk where `-0 == 0`, so `list[-top_k:]` with `top_k=0` becomes `list[-0:]` == `list[0:]` == the full list, not an empty slice.

Two call sites hit this:

```python
# recency-based path, no embedding model
if not self.embedding_model or not self._embeddings_list:
    return self._examples[-top_k:]   # top_k=0 -> returns ALL examples
```

```python
# cosine-similarity path, inside _get_nearest_examples
if len(valid_indices) > 0:
    top_indices = valid_indices[
        np.argsort(similarities[valid_indices])[-top_k:]   # same issue
    ]
```

(The other fallback branch, `list(range(max(0, len(embeddings) - top_k), len(embeddings)))`, happens to compute the empty range correctly for `top_k=0` — it's specifically the two `[-top_k:]` slices that are wrong.)

This is reachable from the public `DynamicFewShotPrompt` API: `DynamicFewShotPrompt(..., max_similar_examples=0)` is a reasonable way to opt out of few-shot examples entirely, but instead every stored example gets silently included in the formatted prompt.

## Ragas version

## Reproduction
```python
from ragas.prompt.dynamic_few_shot import SimpleInMemoryExampleStore

store = SimpleInMemoryExampleStore()
for i in range(5):
    store.add_example({"q": f"question {i}"}, {"a": f"answer {i}"})

result = store.get_examples({"q": "new question"}, top_k=0)
print(len(result))  # prints 5, expected 0
```

## Expected behavior
`top_k=0` returns an empty list in both code paths.

mittalpk:
Fixes #2872

## What
`SimpleInMemoryExampleStore.get_examples(data, top_k=0, ...)` returns **all** stored examples instead of none, in both the recency-based path (no embedding model) and the cosine-similarity path (with an embedding model). Root cause: `list[-top_k:]` with `top_k=0` becomes `list[-0:]` == `list[0:]` == the full list, since Python treats `-0` as `0`.

This is reachable from the public `DynamicFewShotPrompt` API -- `DynamicFewShotPrompt(..., max_similar_examples=0)` is a reasonable way to opt out of few-shot examples, but every stored example gets silently included in the formatted prompt instead.

## Fix
Added an explicit `top_k <= 0` early-return in both affected code paths (the no-embedding-model fallback in `get_examples`, and `_get_nearest_examples` for the embedding-similarity path). The third path -- `list(range(max(0, len(embeddings) - top_k), len(embeddings)))`, the "nothing met the threshold" fallback -- already handled `top_k=0` correctly and needed no change.

## Tests
New `tests/unit/prompt/test_simple_in_memory_example_store.py`:
- `test_get_examples_top_k_zero_returns_none_without_embedding_model` / `..._with_embedding_model` -- confirmed to fail (returning all 5 stored examples) against pre-fix code, pass post-fix.
- `test_get_examples_top_k_positive_respects_limit_without_embedding_model` / `..._with_embedding_model` -- regression check that normal `top_k>0` behavior is unaffected.

## Verification
- `uv run ruff check` / `ruff format --check` on changed files: clean
- `uv run pyright` on changed file: 0 errors
- New tests: 4/4 passing
- Existing `tests/unit/prompt/test_dynamic_few_shot_prompt.py` (16 tests): still passing, no regressions
- Pre-commit hook (`monorepo-ci` format check): passed


mittalpk:
Bumping this one as well, same as #2873 — happy to adjust if needed.

mittalpk:
The failing check-docs job isn't something I can fix from here -- it's a pull_request_target workflow, and GitHub's own actions/checkout refuses to fetch fork PR code in that context by default (a security guard against "pwn request" attacks) unless allow-unsafe-pr-checkout: true is set in the workflow file itself. That's a maintainer-side config change, not anything in this PR's content.

Issue #2760, opened 2026-06-06
Add EvaluationResult summary and threshold checks


vipul674:
## Problem

Ragas EvaluationResult already computes aggregate metric values internally, but users do not have a clear public method for retrieving aggregate scores. This makes CI evaluation gates and regression checks awkward.

## Proposal

Add public helpers to EvaluationResult:

- summary()
- check_thresholds()
- assert_thresholds()

These helpers expose aggregate scores and make CI threshold validation straightforward.

## Proposed API

```python
summary = result.summary()

report = result.check_thresholds(
    min_scores={"faithfulness": 0.8},
    max_scores={"error_rate": 0.2},
    ranges={"answer_relevancy": (0.75, 1.0)},
)

result.assert_thresholds(
    min_scores={"faithfulness": 0.8}
)
```

## Expected Behavior

* summary() returns aggregate metric scores as a plain dictionary.
* check_thresholds() returns a structured pass/fail report without raising on normal threshold failures.
* assert_thresholds() raises AssertionError when threshold checks fail.
* Missing metrics, NaN scores, and non-numeric scores fail clearly.
* Invalid threshold configuration raises ValueError.

## Why This Helps

This makes Ragas easier to use in CI, benchmark scripts, regression tests, and recurring evaluation jobs.

## Backward Compatibility

This is fully additive. Existing EvaluationResult behavior remains unchanged.

## Testing Plan

Add unit tests for summary output, threshold pass/fail behavior, max thresholds, range thresholds, missing metrics, NaN values, invalid inputs, and assert_thresholds().

## Documentation

Update the CI documentation to use summary(), check_thresholds(), and assert_thresholds() instead of stale evaluate(..., in_ci=True) guidance.

ferhimedamine:
EvaluationResult summary and threshold checks are a natural addition — right now you have to iterate over metrics manually to determine pass/fail, which gets tedious for large evaluation suites.

One design suggestion: support both absolute thresholds (faithfulness >= 0.8) and relative thresholds (faithfulness within 5% of baseline). The latter is often more useful for catching regressions without having to maintain an up-to-date absolute target.

For RAG evaluation pipelines where the corpus or memory layer is also changing over time (not just the LLM), RAGAS results should ideally be stored with version tags so you can correlate metric changes with document corpus or agent memory updates. dakera-deploy's session memory can store evaluation snapshots alongside the retrieval context that was used, which helps with this kind of longitudinal tracking.

[...]

PR #2894, opened 2026-07-26
feat(dataset_schema): add EvaluationResult.summary() returning per-metric mean/count/missing

### LINKED ISSUE #2893: dataset_schema: add EvaluationResult.summary() returning per-metric mean/count/missing
## Problem

`src/ragas/dataset_schema.py:412` (`EvaluationResult`) computes per-metric aggregate values internally for `__repr__` (lines 449-454 of `__post_init__`): it runs `safe_nanmean` over each metric column. But it doesn't expose that aggregate to callers. The only public way to read the per-metric mean today is `repr(result)`, which returns a string like `"{'faithfulness': 0.8920, 'answer_relevancy': 0.7650}"` — not machine-readable.

For CI gates and regression tracking, callers want a structured per-metric summary (mean, count, missing) that they can assert against:

```python
# Today
mean = float(repr(result).split("'faithfulness': ")[1].split(",")[0])  # ugly

# What they want
summary = result.summary()  # {"faithfulness": {"mean": 0.89, "count": 30, "missing": 0}, ...}
assert summary["faithfulness"]["mean"] >= 0.80
```

## Why it matters

A `summary()` method is the natural building block for any CI/regression workflow that ingests ragas scores. Without it, callers either (a) parse the `__repr__` string (fragile), (b) call `to_pandas().mean()` (heavier, requires pandas), or (c) hand-roll the loop in `__post_init__` (duplicates the existing `safe_nanmean` path, and gets the NaN handling wrong — `sum()/n` returns NaN, not 0, when any value is NaN).

## Proposed solution

Add a `summary()` method on `EvaluationResult` that returns a structured per-metric aggregate. Reuse the existing `safe_nanmean` helper so the NaN semantics match the rest of the file. Return a `Dict[str, Dict[str, float | int]]` so callers can both `summary[name]["mean"]` and `summary[name]["count"]` for downstream assertions.

```python
def summary(self) -> t.Dict[str, t.Dict[str, t.Any]]:
    """Return per-metric aggregate statistics as a structured dict.

    Returns
    -------
    dict of str to dict
        A dict mapping each metric name to a dict with keys ``mean``,
        ``count`` (number of non-null scores), and ``missing`` (number
        of null scores). ``mean`` is ``float("nan")`` when every score
        for the metric is null.
    """
    return {
        name: {
            "mean": safe_nanmean(values),
            "count": sum(1 for v in values if v is not None),
            "missing": sum(1 for v in values if v is None),
        }
        for name, values in self._scores_dict.items()
    }
```

The method reuses the cached `_scores_dict` from `__post_init__`, so it's O(N metrics) at lookup time, not O(N rows).

## Acceptance criteria

- New `summary()` method on `EvaluationResult` returns `dict[str, dict[str, float | int]]`.
- Each per-metric dict has exactly the keys `mean`, `count`, `missing`.
- `mean` uses `safe_nanmean` so a single `np.nan` in the column does not poison the aggregate.
- `count` and `missing` sum to the total number of scores for the metric.
- `mean` is `nan` when `count == 0`.
- Binary columns (e.g. AspectCritic verdicts) work the same as continuous columns — `mean` of a 0/1 column is the fraction of 1s, which is what most callers want.
- New unit tests cover: 3 metrics, NaN-in-column, all-NaN column, all-present column, mixed metric types.
- `ruff check`, `ruff format --check`, `pyright` clean on the touched file.
- Backward compatible: no public API change (additive only).

## Out of scope

- Threshold checks (`check_threshold`): that's a separate feature with a different shape (caller-supplied thresholds, raise on fail, etc.). The user's issue #2760 mixes the two; this PR keeps them separate so the diff stays focused.
- Per-model cost breakdown: requires a `cost_cb` and is a different layer.
- JSON serialization of the summary: callers can `json.dumps(result.summary())` themselves; no need to add a `summary_json()` method.

## Environment

- ragas `main` (post 0.4.3, commit `298b682`)
- Python 3.12

Harsh23Kashyap:
## Summary

`EvaluationResult` already computes per-metric aggregate values internally for `__repr__` via `safe_nanmean`, but doesn't expose them. Callers that need a structured per-metric summary (CI gates, regression tracking, test assertions) either parse the repr string or hand-roll a loop — duplicating the existing `safe_nanmean` path and getting the NaN semantics wrong.

This PR adds a `summary()` method that returns a structured dict:

```python
result.summary()
# {'faithfulness': {'mean': 0.89, 'count': 30, 'missing': 0},
#  'answer_relevancy': {'mean': 0.77, 'count': 28, 'missing': 2}}
```

Fixes #2893

## Why it matters

Any CI / regression workflow that ingests ragas scores currently has three options, all bad:

1. Parse the `__repr__` string: `mean = float(repr(result).split("'faithfulness': ")[1].split(",")[0])` — fragile and breaks if the repr format changes.
2. Call `to_pandas().mean()`: heavier, requires pandas, doesn't expose `count` / `missing`.
3. Hand-roll the loop in `__post_init__`: duplicates the `safe_nanmean` path and gets NaN semantics wrong (a single `np.nan` poisons `sum()/n`).

`summary()` is the natural building block for any of those workflows, and the existing internal `_scores_dict` already has the per-metric lists built, so the cost is essentially zero.

## Implementation

```python
def summary(self) -> t.Dict[str, t.Dict[str, t.Any]]:
    summary = {}
    for name, values in self._scores_dict.items():
        non_null = [v for v in values if v is not None and not _is_nan(v)]
        count = len(non_null)
        if count == 0:
            mean = float("nan")
        else:
            mean = safe_nanmean(non_null)
        summary[name] = {
            "mean": mean,
            "count": count,
            "missing": len(values) - count,
        }
    return summary
```

Two small guard rails around `safe_nanmean`:

- All-null column: filters to the non-null slice first; if the slice is empty, returns `float("nan")` for `mean` rather than calling `safe_nanmean` (which `TypeError`s on all-`None` input).
- NaN values: counted as missing. The `v is not None` check alone misses `np.nan` because `nan is None` is `False`. The `_is_nan` helper handles only the float NaN case without coercing strings or other types.

The method reuses the cached `_scores_dict` from `__post_init__`, so it's O(N metrics) per call regardless of how many rows the evaluation produced.

## Files changed

- `src/ragas/dataset_schema.py` — `+59` (additive; new `summary()` method + small `_is_nan` helper)
- `tests/unit/test_evaluation_result_summary.py` — new file, `+114`

## Verification

- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run python /tmp/pytest_with_stub.py tests/unit/test_evaluation_result_summary.py -p asyncio -p xdist -v` → 6 passed, 0 failed.
- `uv run ruff check src/ragas/dataset_schema.py tests/unit/test_evaluation_result_summary.py` → All checks passed.
- `uv run ruff format --check ...` → already formatted.
- `uv run pyright src/ragas/dataset_schema.py tests/unit/test_evaluation_result_summary.py` → 0 errors, 0 warnings.

The host venv is missing `langchain_community.chat_models.vertexai` (a pre-existing local-env quirk noted in the top-level `HANDOFF.md`). The PR was tested using a local `/tmp/pytest_with_stub.py` that pre-installs a minimal stub for that module. CI on Ubuntu is unaffected.

## Backward compatibility

- No public API change.
- Additive only: new `summary()` method on `EvaluationResult`. `__init__`, `__post_init__`, `__repr__`, `to_pandas`, `total_tokens`, `total_cost` all unchanged.
- Same NaN semantics as the existing `__repr__` (uses the same `safe_nanmean`).
- The `_is_nan` helper is module-private (leading underscore).

## Out of scope

- Threshold checks (`check_threshold(name, threshold)`) — separate feature with a different shape (caller-supplied thresholds, raise on fail, etc.). The user's issue #2760 mixes the two; this PR keeps them separate so the diff stays focused. A follow-up can add the threshold helper if the maintainers want it.
- Per-model cost breakdown — different layer (`cost_cb` is opt-in), separate PR.
- A `summary()` accessor on the new `metrics.collections` API — the collections metrics already have structured result types, so the value-add there is smaller.
- Fixing the underlying `safe_nanmean` to handle all-`None` arrays directly — that's a pre-existing latent bug (`np.isnan(None)` raises `TypeError`) and would touch every caller. Filtered-then-call is sufficient and contained for this PR.


Harsh23Kashyap:
Opened #2894 to add `EvaluationResult.summary()` returning per-metric mean/count/missing. The shape is the comparison-friendly one folks ask for most when looking across runs. Two small questions before a maintainer pass: (1) if two metrics in the result share a name (e.g. a custom metric overriding `faithfulness`), should `summary()` average them or take the last one? (2) worth a one-line example in the docstring? Happy to push either or both as a follow-up commit.

PR #2897, opened 2026-07-29
fix(dataset): train_test_split should reload from backend when _data is empty

### LINKED ISSUE #2896: fix(dataset): train_test_split returns empty datasets when DataTable._data is empty
## Problem

`DataTable.train_test_split` (`src/ragas/dataset.py:383-453`) calls `self.load(self.name, self.backend, self.data_model)` to "reload from backend when `_data` is empty", but `load` is a `@classmethod` that returns a *new* `DataTable` instance. The return value is discarded, so `self._data` stays empty. The method then proceeds to slice an empty list and returns two empty datasets, with no error.

The intended behavior is to populate `self._data` from the backend when the instance was constructed without `data=` (e.g., `Dataset("name", "local/csv", root_dir=".")`), so that the split reflects the persisted rows.

## Reproduction

```python
import tempfile
from ragas.dataset import Dataset
from ragas.backends.local_csv import LocalCSVBackend

with tempfile.TemporaryDirectory() as tmp_dir:
    backend = LocalCSVBackend(tmp_dir)
    data = [{"name": f"row-{i}", "value": i} for i in range(5)]

    # save 5 rows
    ds = Dataset("my_data", backend, data=data)
    ds.save()

    # new instance with same name + backend, no data=
    ds2 = Dataset("my_data", backend)
    assert len(ds2) == 0                       # _data is empty

    train, test = ds2.train_test_split(test_size=0.4, random_state=42)
    print(len(train), len(test))               # → 0 0  (expected: 3 2)
```

Confirmed locally: `train` and `test` are both empty.

## Why it happens

`load` is declared as a classmethod (`src/ragas/dataset.py:155-199`):

```python
@classmethod
def load(cls, name, backend, data_model=None, **kwargs) -> Self:
    ...
    return cls(name, resolved_backend, None, dict_data)
```

The classmethod returns a new instance. The body of `train_test_split` calls it but throws the return value away, so `self._data` is never updated. The "rehydrate from backend when empty" intent only works if the caller does the swap themselves.

A correct in-place rehydrate already exists on the same class: `DataTable.reload` (`src/ragas/dataset.py:275-288`) is the in-place counterpart and is used by `metrics/base.py:1372, 1412`.

## Proposed fix

Replace line 395:

```python
if not self._data:
    self.load(self.name, self.backend, self.data_model)
```

with:

```python
if not self._data:
    self.reload()
```

`reload()` already handles the `DATATABLE_TYPE == "Experiment"` branch, validates against `self.data_model` when set, and assigns to `self._data` in place. Behavior matches the current intent.

## Scope

- 1-line code change in `src/ragas/dataset.py`.
- Add a regression test in `tests/unit/backends/test_inmemory.py` that:
  1. Saves a dataset to a `LocalCSVBackend`.
  2. Constructs a fresh `Dataset` with the same name and backend (no `data=`).
  3. Calls `train_test_split`.
  4. Asserts the rows are loaded and the split sizes match `test_size`.

## Backward compatibility

Pure bug fix. `reload()` already exists and is part of the public API. No signature changes, no deprecations.

## Acceptance criteria

- [ ] `train_test_split` on a `Dataset` constructed without `data=` returns the expected split when the backend has the named dataset.
- [ ] The same fix also covers `Experiment.train_test_split` (inherits from `DataTable`).
- [ ] The existing `test_dataset_train_test_split_uses_inmemory` and `test_train_test_split_comprehensive` tests still pass.
- [ ] A new test covers the empty-`_data` rehydrate path.

## Out of scope

- Re-doing the broader `Dataset` / `DataTable` API surface (e.g., making `reload()` automatic, deprecating the `data=` kwarg).
- A separate bug for `load_experiment` / `load_dataset` returning a new instance — that's the documented behavior of the public `load()` classmethod and matches the user's mental model in `Dataset.load("name", backend)`.

Harsh23Kashyap:
Fixes #2896

## What

`DataTable.train_test_split` discarded the return value of `self.load(...)`. `load` is a `@classmethod` that returns a *new* `DataTable` instance, so the call did not actually populate `self._data`. The result: a `Dataset` constructed without `data=` (just name + backend) silently produced two empty train/test datasets instead of rehydrating from the backend.

Switched the call to `self.reload()`, which is the in-place counterpart already used by `metrics/base.py:1372,1412`.

## Repro (before)

```python
import tempfile
from ragas.dataset import Dataset
from ragas.backends.local_csv import LocalCSVBackend

with tempfile.TemporaryDirectory() as tmp_dir:
    backend = LocalCSVBackend(tmp_dir)
    data = [{"name": f"row-{i}", "value": i} for i in range(5)]

    Dataset("my_data", backend, data=data).save()
    fresh = Dataset("my_data", backend)              # _data is empty

    train, test = fresh.train_test_split(test_size=0.4, random_state=42)
    # → 0 0  (expected: 3 2)
```

After: `train=3, test=2` as expected.

## Files

- `src/ragas/dataset.py` — 1-line change at `train_test_split` (line 395): `self.load(...)` → `self.reload()`.
- `tests/unit/backends/test_inmemory.py` — new regression test `test_train_test_split_rehydrates_from_backend_when_data_empty` covering the previously-untested empty-`_data` path.

## Verification

- `uv run ruff check src/ragas/dataset.py tests/unit/backends/test_inmemory.py` → All checks passed
- `uv run ruff format --check` → 2 files already formatted
- `uv run pyright src/ragas/dataset.py` → 0 errors, 0 warnings
- `tests/unit/backends/test_inmemory.py` → 80 passed, 6 skipped
- `tests/unit/test_datatable_inheritance.py` → 16 passed
- `tests/unit/test_experiment.py` → 22 passed

## Risk

Pure bug fix; `reload()` is the documented in-place rehydrate method and was already part of the public API. No signature changes, no deprecations. The two existing `train_test_split` tests still pass (they both pass `data=` so the `if not self._data` branch was never exercised).


ErenAta16 (commented):
Diagnosis confirmed against the source. `load` really is a classmethod that builds a fresh object:

```python
@classmethod
def load(cls: t.Type[Self], name, backend, data_model=None, **kwargs) -> Self:
    ...
    return cls(name, resolved_backend, None, dict_data)
```

so `self.load(self.name, self.backend, self.data_model)` constructed a populated `DataTable`, threw it away, and left `self._data` exactly as empty as it was. `reload` is the in-place counterpart:

```python
def reload(self) -> None:
    ...
```

which is what the call site wanted. One-line fix, right method.

The failure mode is worth spelling out in the description because it is quiet: the split does not raise, it returns two empty datasets. A caller who checks `len(train) + len(test)` gets `0` and can reasonably conclude their source data was empty rather than that the reload silently no-opped.

The test is well-targeted. Constructing a fresh `Dataset(same_name, same_backend)` without `data=` and then splitting is precisely the path that was broken, and asserting the split reflects the persisted rows rather than asserting `reload` was called keeps it a behaviour test. The Given/When/Then docstring makes it obvious what regressed if it ever fails again.

One thing worth a glance while you are in here: `self.load(...)` with a discarded return is the kind of mistake that repeats, since the classmethod and the instance method read almost identically at a call site. If there are other `self.load(` occurrences in `dataset.py` or its siblings, they are worth checking for the same shape. I did not sweep for it, so this is a suggestion rather than a finding.

PR #2899, opened 2026-07-30
fix(prompt): top_k=0 returns all examples instead of empty list

### LINKED ISSUE #2872: DynamicFewShotPrompt: max_similar_examples=0 (top_k=0) returns all examples instead of none

### LINKED ISSUE #2877: InMemoryExampleStore.get_nearest_examples returns all examples when top_k=0
Passing `top_k=0` to `InMemoryExampleStore.get_nearest_examples()` returns every example above the threshold instead of returning nothing.

The culprit is the negative slice in `src/ragas/prompt/few_shot_pydantic_prompt.py`:

```python
top_indices = valid_indices[np.argsort(similarities[valid_indices])[-top_k:]]
```

With `top_k=0`, `[-top_k:]` becomes `[-0:]`, and numpy reads that as `[0:]`, the whole array. So asking for zero examples quietly hands you all of them.

This is the same bug that was already fixed for `SimpleInMemoryExampleStore` in #2872. That fix lives in a different file and doesn't touch `InMemoryExampleStore`, so this store still has it.

Quick repro:

```python
from ragas.prompt.few_shot_pydantic_prompt import InMemoryExampleStore

query = [1.0, 0.0]
embeddings = [[1.0, 0.0], [0.9, 0.1], [0.8, 0.2]]

InMemoryExampleStore.get_nearest_examples(query, embeddings, top_k=0, threshold=0.0)
# returns [2, 1, 0], expected []
```

`get_examples(..., top_k=0)` hits the same path since it just delegates here.

Expected: any non-positive `top_k` should return an empty list.

Version: main (0.4.3).

feiiiiii5:
## Root Cause

Python's negative-slice quirk where -0 == 0 causes list[-top_k:] with top_k=0 to become list[0:] == the full list, not an empty slice.

This affects three code paths:
1. SimpleInMemoryExampleStore.get_examples (recency path, no embedding model)
2. SimpleInMemoryExampleStore._get_nearest_examples (cosine similarity path)
3. InMemoryExampleStore.get_nearest_examples (pydantic prompt variant)

## Fix

Add early-return guard for top_k <= 0 in all three locations.

Fixes #2872
Fixes #2877

AI disclosure: This PR was authored with assistance from an AI coding assistant. All changes were reviewed and verified by a human contributor.


feiiiiii5:
The Claude Docs Check failure appears unrelated to this PR's changes (which modify dynamic_few_shot.py, few_shot_pydantic_prompt.py, and test files - no documentation files touched). This seems to be a pre-existing check issue. Could a maintainer confirm?

feiiiiii5:
Closing per the stale-PR discipline: this PR has had no human review in 7 days, and the remaining CI failure (`check-docs`) is a repository-level policy for external-fork PRs that cannot be fixed from the fork side. The fix itself is ready with regression tests — happy to re-open if a maintainer prefers.

PR #1880, opened 2025-01-25
Changed the parse_run_traces to include last 4 letters of run_id

Vidit-Ostwal:
#1871 
Changed the parse_run_traces function, to include the last 4 letter of the run_id of the trace, result of which is a unique key for every call.

Issue #2692, opened 2026-04-22
[Security] Agentic Workflow Injection in Claude Docs Check


xinyi-hou:
## Summary

The `.github/workflows/claude-docs-check.yml` workflow runs on `pull_request_target`, checks out untrusted pull request code, passes attacker-controlled PR metadata and diff content into `anthropics/claude-code-action`, and later interpolates the model-generated `reason` field directly into shell commands that comment on or label the PR.

A malicious PR author who modifies `src/**/*.py` can attempt to influence Claude through the PR title, body, or changed source code. If the model returns shell-special content in `reason`, that content may be interpreted by Bash when the workflow posts PR comments.

## Affected Product

- Repository: `vibrantlabsai/ragas`
- Workflow: `.github/workflows/claude-docs-check.yml`
- Type of Issue: Agentic Workflow Injection
- Affected Versions: Workflow present on `main` as of commit `298b68274234c060deacab3cf5fb52aa3a20e885`; any branch, release, fork, or deployment containing equivalent workflow logic should be considered affected.
- Affected Files: `.github/workflows/claude-docs-check.yml`
- Environment: GitHub-hosted `ubuntu-latest` runner using Bash, GitHub CLI, `jq`, `actions/checkout@v4`, and `anthropics/claude-code-action@v1`

## Details

Relevant workflow behaviors include:

- `.github/workflows/claude-docs-check.yml:4` — `pull_request_target`
- `.github/workflows/claude-docs-check.yml:14` — `pull-requests: write`
- `.github/workflows/claude-docs-check.yml:15` — `issues: write`
- `.github/workflows/claude-docs-check.yml:22` — `${{ github.event.pull_request.head.repo.full_name }}`
- `.github/workflows/claude-docs-check.yml:23` — `${{ github.event.pull_request.head.sha }}`
- `.github/workflows/claude-docs-check.yml:33` — `allowed_non_write_users: "*"`
- `.github/workflows/claude-docs-check.yml:38` — `${{ github.event.pull_request.title }}`
- `.github/workflows/claude-docs-check.yml:76` — JSON schema allows a free-form `reason` string
- `.github/workflows/claude-docs-check.yml:106` — `${{ steps.parse.outputs.reason }}`
- `.github/workflows/claude-docs-check.yml:117` — `${{ steps.parse.outputs.reason }}`

[...]

PR #2900, opened 2026-07-30
fix(tracing): preserve repeated prompt invocations

### LINKED ISSUE #1871: Traces in evaluation result are only saving the last prompt per metric name

**Describe the bug**
Evaluation result's `traces` property only keeps one trace per trace.name, inconsistent with `ragas_traces`. So in FactualCorrectness there are multiple calls to decompose_claims for example

Ragas version: 0.2.11
Python version: 3.11

**Code to Reproduce**
```
 from ragas.metrics._factual_correctness import FactualCorrectness
 from ragas import evaluate
from datasets import Dataset

dataset = Dataset.from_dict({"response": ["Eifel tower is in Paris"], "reference": ["Paris, France is a city where Eifel tower is located"]})
fc = FactualCorrectness()
score = evaluate(
    dataset, metrics=[fc]
)
print(score.scores)
# [{'factual_correctness': 0.67}]
print(score.traces[0]['factual_correctness'])
```

```json
{'claim_decomposition_prompt': {'input': ClaimDecompositionInput(response='Paris, France is a city where Eifel tower is located'),
                                'output': ClaimDecompositionOutput(claims=['Paris is a city in France.', 'The Eiffel Tower is located in Paris.'])},
 'n_l_i_statement_prompt': {'input': NLIStatementInput(context='Eifel tower is in Paris', statements=['Paris is a city in France.', 'The Eiffel Tower is located in Paris.']),
                            'output': NLIStatementOutput(statements=[StatementFaithfulnessAnswer(statement='Paris is a city in France.', reason='The statement about Paris being a city in France is a well-known fact and can be inferred from general knowledge, but it is not directly stated in the context provided.', verdict=0), StatementFaithfulnessAnswer(statement='The Eiffel Tower is located in Paris.', reason='The context explicitly states that the Eiffel Tower is in Paris, making this statement directly inferable.', verdict=1)])}}
```
As you can see in the example traces for that metric include only one prompt per prompt name, instead of 2 like it should, since in the code we call these metrics twice, this can be proven by checking score.ragas_traces

```
In [34]: [score.metadata for score in score.ragas_traces.values()]
Out[34]: 
[{'type': <ChainType.EVALUATION: 'evaluation'>},
 {'type': <ChainType.ROW: 'row'>, 'row_index': 0},
 {'type': <ChainType.METRIC: 'metric'>},
 {'type': <ChainType.RAGAS_PROMPT: 'ragas_prompt'>},
 {'type': <ChainType.RAGAS_PROMPT: 'ragas_prompt'>},
 {'type': <ChainType.RAGAS_PROMPT: 'ragas_prompt'>},
 {'type': <ChainType.RAGAS_PROMPT: 'ragas_prompt'>}]
```

**Error trace**

**Expected behavior**
result.traces has ALL traces per prompt, including the metrics with the same name

**Additional context**
I think the problem is in ragas/callbacks.py::parse_run_traces() line 158, it assigns by metric name so if there are multiple metrics in the same call it will only save the last trace.

I couldnt find anythin in the open issues to correlate with my issue so I thought it was a new one, plus that does seem like unexpected behaviour, please let me know otherwise

tandixit95:
## Summary

Fixes #1871 by preserving every prompt invocation in `EvaluationResult.traces` when the same prompt is called multiple times inside a metric.

- detects duplicate prompt names within a metric trace
- uses ordered `prompt_1`, `prompt_2`, ... keys only when names would collide
- stores the original prompt name in each numbered trace entry
- preserves the existing name-keyed trace shape when prompt names are unique
- updates `GeneticOptimizer` to consume both the ordered and legacy trace shapes
- preserves the optimizer's previous last-invocation behavior when a prompt repeats
- adds regression tests for duplicate traces, legacy shape compatibility, and optimizer lookup

## Context

This follows the design discussion in #1871 and the earlier work in #1880.

The earlier PR established the ordered prompt numbering approach, but review identified two concerns:

1. changing the trace shape globally was broader than necessary
2. `GeneticOptimizer` also needed to be updated and tested

This version addresses both. Unique prompt names retain the existing public structure; numbering is only introduced for metric traces that would otherwise silently overwrite data.

## Example

Before, repeated prompt names were overwritten:

```python
{
    "claim_decomposition_prompt": {...},
    "n_l_i_statement_prompt": {...},
}
```

Even if `claim_decomposition_prompt` ran twice, only the final call survived.

For a trace containing duplicates, the result is now ordered and lossless:

```python
{
    "prompt_1": {"name": "claim_decomposition_prompt", ...},
    "prompt_2": {"name": "n_l_i_statement_prompt", ...},
    "prompt_3": {"name": "claim_decomposition_prompt", ...},
}
```

Metrics whose prompt names are all unique keep the existing name-keyed representation unchanged.

## Validation

- `ruff format --check` on changed files
- `ruff check` on changed files
- 45 relevant unit tests passing, including callbacks, optimizer trace compatibility, dataset schema, cache, and executor coverage

## CI note

The `Claude Docs Check` currently fails during `actions/checkout`, before repository code is executed, because the `pull_request_target` workflow attempts to check out fork code without `allow-unsafe-pr-checkout`. The regular CI workflow is currently `action_required` pending maintainer approval for the fork run. This is unrelated to the changes in this PR.


tandixit95:
The failing `check-docs` status is coming from the base repository workflow rather than this patch. `actions/checkout@v4` is refusing to check out the fork head from the `pull_request_target` workflow because it uses the base-repo token for untrusted fork code. ReadTheDocs is green.

This overlaps the existing workflow-security problem tracked in #2692; there are already workflow-fix PRs in flight, so I’m not opening a duplicate CI PR or opting into unsafe checkout. The tracing change itself does not modify docs/CI files.

tandixit95:
@jjmachan this implements the ordered-list trace shape discussed in #1871, preserving repeated prompt invocations without name collisions and updating the genetic optimizer consumer. Focused validation on the current head: 4 regression tests passed; Ruff lint/format and `git diff --check` passed. The remaining `check-docs` failure is the existing fork-checkout workflow issue noted above. When you have a chance, could you review the data-shape change?

PR #2908, opened 2026-08-01
fix: BleuScore silently scores only the first sentence of the response

hassaanch23:

`BleuScore` silently scores **only the first sentence** of the response and discards everything after it. No exception is raised — a plausible-looking float comes back, so the bug is invisible in normal use.

### Root cause

sacrebleu's `corpus_bleu(hypotheses, references)` takes `references` as a list of reference **streams**. Each stream is one *alternative* reference holding one segment per hypothesis, so the shape is `[n_alternative_refs][n_segments]`.

Both implementations split on `". "` and then built:

```python
reference_sentences = reference.split(". ")     # ["A is one", "B is two", "C is three."]
reference = [[r] for r in reference_sentences]  # [["A is one"], ["B is two"], ["C is three."]]
response  = response.split(". ")                # 3 hypotheses
corpus_bleu(response, reference)
```

That passes **3 reference streams of 1 segment each** alongside **3 hypotheses**. sacrebleu zips the hypotheses against the reference streams, and `zip` stops at the shortest — length 1. So:

1. Only `response_sentences[0]` is ever scored; the rest of the response is dropped.
2. Reference sentences 2..N are reinterpreted as *alternative references* for that first sentence.

### How to replicate

```python
import asyncio
from ragas.metrics import BleuScore
from sacrebleu import corpus_bleu

metric = BleuScore()

async def ragas(ref, resp):
    return await metric._single_turn_ascore(
        SingleTurnSample(reference=ref, response=resp), None
    )

def sacrebleu_truth(ref, resp):
    return corpus_bleu([resp], [[ref]]).score / 100

ref  = "A is one. B is two. C is three."
print(asyncio.run(ragas(ref, ref)))       # 0.0   <- identical text!
print(sacrebleu_truth(ref, ref))          # 1.0
```

| # | Reference | Response | Before | sacrebleu | After |
|---|-----------|----------|-------:|----------:|------:|
| A | `A is one. B is two. C is three.` | *identical* | **0.0000** | 1.0000 | 1.0000 |
| B | `The cat sat on the mat quietly. The dog barked at the mailman.` | `The cat sat on the mat quietly. zzz qqq wombat banana telephone.` | **1.0000** | 0.4931 | 0.4931 |
| C | `…quietly. …mailman. Birds flew over the house.` | `The cat sat on the mat quietly.` | **0.8694** | 0.1969 | 0.1969 |
| D | `The cat sat on the mat quietly. Alpha beta gamma delta epsilon.` | `The cat sat on the mat quietly. Completely different words entirely here.` | **1.0000** | 0.5296 | 0.5296 |

The scores are wrong in **both** directions:

- **Case A** — a *perfect* response scores **0.0**. The first hypothesis (`"A is one"`) is 3 tokens, so it has no 4-grams, 4-gram precision is 0, and the score collapses. Any multi-sentence text whose first sentence is shorter than four tokens scores 0 regardless of quality.
- **Cases B/D** — a response whose entire second half is **garbage** scores a perfect **1.0**, because only the correct first sentence was ever compared.

Multi-sentence text is the normal case for RAG answers, so this affects most real usage.

## Changes Made

- **`src/ragas/metrics/_bleu_score.py`** — score the sample as a single segment: `corpus_bleu([response], [[reference]], **self.kwargs)`.
- **`src/ragas/metrics/collections/_bleu_score.py`** — same fix; the v2 metric had the identical bug.
- **`tests/unit/test_bleu_score.py`** — new; BLEU previously had no unit tests.

### Why this fix, and not a corrected sentence split

- It matches how **`ChrfScore` in this repo already calls sacrebleu** correctly (`_chrf_score.py`: `references = [[reference]]`, `hypotheses = [response]`), so BLEU now follows the existing in-repo precedent.
- Keeping the split but passing a single stream — `corpus_bleu(resp_sents, [ref_sents])` — is **also silently wrong** whenever the sentence counts differ: `zip` truncates to the shorter list and drops the remainder with no error. Verified with a 3-sentence reference against a 2-sentence response: it returns 0.69 with `BP = 1.000`, ignoring the dropped sentence. Single-segment scoring instead applies the brevity penalty correctly (`BP = 0.607`).
- Splitting on `". "` was never robust regardless: it mishandles abbreviations (`"Dr. Smith"`), decimals (`"3. 14"`), and any terminator other than a period.

### Behaviour change

Single-sentence inputs are **unaffected**. The example in `docs/concepts/metrics/available_metrics/traditional.md` still returns `0.7071067811865478` exactly — there's a test asserting it, so no docs change is needed. Multi-sentence inputs change, because they were previously wrong.

### How to Test
- [x] Automated tests added/updated — `tests/unit/test_bleu_score.py`, 26 tests
- [x] Manual testing steps:
  1. `uv run pytest tests/unit/test_bleu_score.py` → 26 passed
  2. Revert the two source files and rerun → **12 of the 26 fail**, confirming they're genuine regression tests
  3. `uv run pytest tests/unit` → no new failures
  4. `uv run pytest tests/e2e/metrics_migration/test_bleu_migration.py` → 5 passed (legacy and v2 stay in agreement)

The tests cover the four cases above, differential tests against sacrebleu called directly, legacy↔v2 agreement, empty-string edge cases, `kwargs` pass-through, and a lock on the documented example value.

`ruff format`, `ruff check`, and `pyright` are all clean on the changed files.

## References

- sacrebleu `corpus_bleu` signature — [`sacrebleu/compat.py`](https://github.com/mjpost/sacrebleu/blob/master/sacrebleu/compat.py): `references` is documented as "A sequence of reference documents with document being defined as a sequence of reference strings"
- In-repo precedent for correct usage: `src/ragas/metrics/_chrf_score.py`
- Sentence splitting was introduced in #1826 (`fix: remove pysbd and sentence segmenting`), on top of the sacrebleu switch in #1744

---


ErenAta16 (commented):
Reproduced this against real `sacrebleu==2.6.0`. The fix is correct, all four hardcoded expectations in the new test match to four decimal places, and the bug is meaningfully worse than "scores only the first sentence".

**`main` does not just truncate, it inverts.** Same four cases, both implementations:

```
case                             main      this PR   PR expects
identical multi-sentence text   0.0000     1.0000      1.0
garbage after first sentence    1.0000     0.4931      0.4931
content missing after first     0.8694     0.1969      0.1969
mismatched trailing sentence    1.0000     0.5296      0.5296
```

A response identical to the reference scores **0.0**. A response whose second sentence is `zzz qqq wombat banana telephone` scores **1.0**. Those are the two most common shapes in a real eval set and they come out backwards, so this is not a precision nit, it is the metric pointing the wrong way.

**Why, since the mechanism explains both halves.** Passing `[[r1], [r2], [r3]]` gives sacrebleu three reference *streams* of one segment each against three hypothesis segments. It does not raise on the length mismatch and it does not warn:

```python
corpus_bleu(["a b c", "d e f", "g h i"], [["a b c"]])   # -> 0.0, no warning
```

What survives is the first hypothesis sentence scored against the reference sentences treated as *alternative* references for it. So the trailing content is not weighted less, it is not present at all, which is why the tail can be anything:

```
reference: "The cat sat on the mat quietly. The dog barked at the mailman."
  tail "The dog barked at the mailman."     main=1.0000
  tail "zzz qqq wombat banana telephone."   main=1.0000
  tail "totally unrelated nonsense here."   main=1.0000
  no tail at all                            main=0.8694
```

And the 0.0 on identical text is the other half of the same mechanism: `"A is one. B is two. C is three."` splits to a 3-token first sentence, BLEU-4 finds no 4-gram, and the geometric mean is zero. So under `main` the score is largely a function of how long the first sentence happens to be and whether a second one exists, which is a good way to describe a metric that is not measuring anything.

**One thing worth deciding rather than leaving implicit.** `reference_bleu()` in the test calls `corpus_bleu([response], [[reference]])`, which is the same call the implementation now makes, so any test using it passes by construction. The four hardcoded floats in `MULTI_SENTENCE_CASES` are the real anchor, and they are the right choice for that reason. The tradeoff is that they are tied to sacrebleu's default tokenizer: they reproduce exactly on `2.6.0`, but a future default change to `13a` tokenization or `effective_order` would move them and the failure would read as a regression in ragas rather than a change in sacrebleu. A version assertion in the module, or a note next to the constants saying where they came from, would make that diagnosable in one look. Non-blocking.

**Both call sites fixed, which matters.** `_bleu_score.py` and `collections/_bleu_score.py` had the same split independently, so fixing one would have left the two implementations of the same metric disagreeing, and the collections one is the newer path people are being moved onto.

The comment you left in place of the split is the part I would keep verbatim. "references is a list of reference streams, each holding one segment per hypothesis" is exactly the sentence whose absence caused this, and it is not obvious from sacrebleu's signature.

hassaanch23:
Thanks @ErenAta16 — that's a genuinely thorough review, and the inversion framing is better than mine. I'd described it as truncation; "the metric points the wrong way" is the accurate summary, and your table makes the case far more directly than my commit message did. The identical-text → `0.0` case is the one that should worry people most, since that's the shape a passing eval is supposed to have.

Acted on the version concern in `a88f40a`.

I went with pinning the **defaults** rather than `sacrebleu.__version__`. A version assertion would fail on every routine upgrade even when scoring is untouched, which trains people to bump it without checking — exactly the reflex that would hide a real change. So the guard asserts the four settings the constants actually depend on:

```python
SACREBLEU_BLEU_DEFAULTS = {
    "tokenizer": "13a",
    "smooth_method": "exp",
    "effective_order": False,
    "max_ngram_order": 4,
}
```

with a note above the constants recording that they're exact `corpus_bleu([response], [[reference]])` outputs on sacrebleu 2.6.0 rather than hand-derived numbers. If a release moves any of them the guard fails with:

> sacrebleu 2.6.0 changed its BLEU defaults: expected {…}, got {…}. The hardcoded scores in MULTI_SENTENCE_CASES and the documented-example value were generated under the previous defaults and need regenerating. This is a sacrebleu behaviour change, not a ragas scoring regression.

Verified it actually bites by simulating a default change to `intl` — it fails there, and the four score assertions stay green, so the diagnosis lands in one place instead of four. `27 passed` on 2.6.0 (was 26).

Small correction to one detail: `13a` is already sacrebleu's default rather than a future move to it, so the live risk is a shift *away* from it — which is what the guard now covers.

On `reference_bleu()` — agreed it passes by construction, and that's deliberate: it's there to pin legacy↔v2 agreement and the sacrebleu-correct baseline for the differential cases, not to independently validate the four floats. Those stand on the hardcoded values alone. Happy to add a docstring line saying so if you think it reads as stronger evidence than it is.

Also worth flagging for whoever picks this up: `check-docs` is red for reasons unrelated to this PR — the workflow uses `pull_request_target` with `actions/checkout` on the fork ref, which GitHub now refuses (`Refusing to check out fork pull request code from a 'pull_request_target' workflow`). It fails for every fork PR, and #2606 doesn't appear to have fully resolved it.

ErenAta16 (commented):
Pinning the defaults rather than the version is the better call, and your reason for it is the right one. A `sacrebleu.__version__` assertion fails on every routine bump regardless of whether scoring moved, so the reflex it trains is "bump the number", which is exactly the moment a real tokenizer change would slip through. Pinning `tokenize`, `smooth_method` and the rest means the constants stay reproducible because the inputs to them are fixed, and a genuine default change surfaces as a scoring diff rather than as a version mismatch nobody reads.

That also makes the four hardcoded floats defensible in a way they were not before. They are now the output of a fully specified configuration rather than of whatever sacrebleu happened to default to on the machine that generated them.

Nothing further from me on this one. For the record, the numbers I measured against `sacrebleu==2.6.0` still line up with the constants in the PR to four decimals, and `main` still produces `0.0000` for identical multi-sentence text and `1.0000` when the second sentence is replaced with unrelated tokens.

Issue #1981, opened 2025-03-26
Misleading documentation for "LLM Based Context Precision without reference"


muhammadyaseen:
[x] I have checked the [documentation](https://docs.ragas.io/) and related resources and couldn't resolve my bug.

**Describe the bug**

I think the documentation for LLM Based Context Precision without reference is wrong. The doc is available at: https://docs.ragas.io/en/latest/concepts/metrics/available_metrics/context_precision/#context-precision-without-reference

It says:
> `LLMContextPrecisionWithoutReference` metric can be used when you have both retrieved contexts and also _reference contexts_ associated with a `user_input`.

I believe the emphasized part is wrong, since this evaluation works without reference context, as the name and heading suggests.

Looking at the code also confirms this:

```python
@dataclass
class LLMContextPrecisionWithoutReference(LLMContextPrecisionWithReference):
    name: str = "llm_context_precision_without_reference"
    _required_columns: t.Dict[MetricType, t.Set[str]] = field(
        default_factory=lambda: {
            MetricType.SINGLE_TURN: {"user_input", "response", "retrieved_contexts"}
        }
    )

    def _get_row_attributes(self, row: t.Dict) -> t.Tuple[str, t.List[str], t.Any]:
        return row["user_input"], row["retrieved_contexts"], row["response"]

```
... it only requires: `user_input`, `response` `retrieved_contexts`.

Ragas version:
Python version:

**Code to Reproduce**
Share code to reproduce the issue

**Error trace**

**Expected behavior**

Updated documentation, for ex:
> `LLMContextPrecisionWithoutReference` metric can be used when you have retrieved contexts (`retrieved_contexts`) associated with a `user_input`.

**Additional context**

Rahul2512Chauhan:
Hi @muhammadyaseen, I’d like to work on this and fix the misleading documentation for LLMContextPrecisionWithoutReference. I’ll submit a PR soon. 🙂

hashim-farooq:
Hey @Rahul2512Chauhan , I believe this issue was resolved. Maybe it should be closed.

GagandeepSingh20:
Hi, I checked the current documentation and it looks like this has been addressed in #2266, which corrected the description of `LLMContextPrecisionWithoutReference`.

Since the current docs now mention that this metric can be used without a reference answer, I think this issue may be resolved. Could a maintainer confirm and close this if appropriate?

Bhargavi-Kambhampaty:
[...]

PR #2938, opened 2026-08-11
fix(metrics): ContextPrecision can return exactly 1.0

### LINKED ISSUE #2909: ContextPrecision never returns exactly 1.0: unconditional 1e-10 in the average-precision denominator
## Describe the bug

`ContextPrecision` can never return exactly `1.0`. A perfectly ranked retrieval scores `0.9999999999`, so any threshold check or test asserting `score == 1.0` fails.

The `1e-10` guard in `_calculate_average_precision` is added to the denominator unconditionally rather than only when it would divide by zero:

```python
# src/ragas/metrics/_context_precision.py
cumsum = 0
numerator = 0.0
for i, ver in enumerate(verifications):
    v = 1 if ver.verdict else 0
    cumsum += v
    if v:
        numerator += cumsum / (i + 1)

denominator = cumsum + 1e-10
score = numerator / denominator
```

so every non-zero score is scaled by `cumsum / (cumsum + 1e-10)`.

## Ragas version

`main`

## Code to Reproduce

```python
def ragas_ap(verdicts):
    cumsum, numerator = 0, 0.0
    for i, v in enumerate(verdicts):
        cumsum += v
        if v:
            numerator += cumsum / (i + 1)
    return numerator / (cumsum + 1e-10)

def textbook_ap(verdicts):
    cumsum, num = 0, 0.0
    for i, v in enumerate(verdicts):
        cumsum += v
        if v:
            num += cumsum / (i + 1)
    return num / cumsum if cumsum else 0.0
```

## Error trace

No error; the value is silently off.

```
case                    ragas                 correct     equal
all relevant [1,1,1]    0.9999999999666667    1.000000    False
single relevant [1]     0.9999999999          1.000000    False
[1,0,1]                 0.8333333332916666    0.833333    False
[0,1,1]                 0.5833333333041666    0.583333    False
20 relevant             0.9999999999950001    1.000000    False
no relevant [0,0]       0.0                   0.000000    True
empty []                0.0                   0.000000    True
```

The asymmetry is the part that bites: the zero case *is* exactly `0.0`, so a user reasonably expects the perfect case to be exactly `1.0`, and it never is. It also means the metric is not exactly reproducible against any external average-precision implementation, which makes cross-checking a pipeline harder than it should be.

The size of the error is negligible for ranking, so this is a correctness and testability issue rather than a scoring one.

## Expected behavior

`1.0` for a perfectly ranked retrieval, and exact agreement with the standard average-precision definition.

Guarding only the zero case preserves the division-by-zero protection without touching any other result:

```python
score = numerator / cumsum if cumsum else 0.0
```

`cumsum` is an integer count of relevant items, so `cumsum == 0` is exact and there is no floating-point reason for the epsilon.

## Additional context

Two things I checked while looking at this, so they do not get re-investigated:

**The average-precision formula itself is correct.** I went in expecting the `if v:` guard to be missing, which is the usual way this gets written wrong, and it is present. Precision is accumulated only at relevant positions and divided by the number of relevant items, matching the standard definition. Only the denominator epsilon is off.

**The NaN branch below is unreachable.**

```python
score = numerator / denominator
if np.isnan(score):
    logger.warning(
        "Invalid response format. Expected a list of dictionaries with keys 'verdict'"
    )
```

`denominator` is at least `1e-10` and `numerator` is a finite sum of floats, so `score` is never NaN. The warning names a malformed-input case that cannot reach this line, since `verifications` is already typed `List[Verification]` and anything malformed fails earlier. Whatever it was meant to catch, it will not, and the message would misdirect anyone debugging a real problem. Worth either removing it or moving the check to where the responses are actually parsed.

Happy to open a PR for both if the direction looks right.

panshularora:
Fixes #2909
Also clarifies docs for without-reference precision (#1981) and updates example scores that showed 0.9999999999.

## Summary
Average precision used `cumsum + 1e-10` in the denominator unconditionally, so perfect rankings never scored exactly 1.0. Only guard the zero-relevant case. Applied in legacy + collections implementations; docs examples updated.

## Test plan
- [ ] Perfect ranking yields 1.0
- [ ] Empty/all-irrelevant yields 0.0

### DIFF
--- a/src/ragas/metrics/_context_precision.py
+++ b/src/ragas/metrics/_context_precision.py
@@ -123,8 +123,9 @@ class LLMContextPrecisionWithReference(MetricWithLLM, SingleTurnMetric):
             if v:
                 numerator += cumsum / (i + 1)
 
-        denominator = cumsum + 1e-10
-        score = numerator / denominator
+        # Guard zero-relevant only; unconditional 1e-10 made perfect rankings
+        # score 0.9999999999 instead of 1.0.
+        score = numerator / cumsum if cumsum else 0.0
         if np.isnan(score):
             logger.warning(
                 "Invalid response format. Expected a list of dictionaries with keys 'verdict'"
@@ -242,9 +243,8 @@ class NonLLMContextPrecisionWithReference(SingleTurnMetric):
             if v:
                 numerator += cumsum / (i + 1)
 
-        denominator = cumsum + 1e-10
-        score = numerator / denominator
-        return score
+        # Guard zero-relevant only so a perfect ranking can score exactly 1.0.
+        return numerator / cumsum if cumsum else 0.0
 
 
 @dataclass
--- a/docs/concepts/metrics/available_metrics/context_precision.md
+++ b/docs/concepts/metrics/available_metrics/context_precision.md
@@ -46,7 +46,7 @@ print(f"Context Precision Score: {result.value}")
 
 Output:
 ```
-Context Precision Score: 0.9999999999
+Context Precision Score: 1.0
 ```
 
 !!! note "Synchronous Usage"
@@ -90,7 +90,7 @@ print(f"Context Utilization Score: {result.value}")
 
 Output:
 ```
-Context Utilization Score: 0.9999999999
+Context Utilization Score: 1.0
 ```
 
 Note that even if an irrelevant chunk is present at the second position in the array, context precision remains the same. However, if this irrelevant chunk is placed at the first position, context precision reduces:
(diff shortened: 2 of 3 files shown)


src/ragas/metrics/_context_precision.py
```diff
@@ -123,8 +123,9 @@ def _calculate_average_precision(
             if v:
                 numerator += cumsum / (i + 1)
 
-        denominator = cumsum + 1e-10
-        score = numerator / denominator
+        # Guard zero-relevant only; unconditional 1e-10 made perfect rankings
+        # score 0.9999999999 instead of 1.0.
+        score = numerator / cumsum if cumsum else 0.0
         if np.isnan(score):
```
rajeshl8, line 129: Building on Varun's feedback about synchronizing the comments across the two mirrored implementations — since this exact numerator / cumsum if cumsum else 0.0 guard now appears in 4 places (_context_precision.py L128 & L247, collections/metric.py L129 & L249), could this be pulled into one shared helper, e.g. _safe_average_precision(numerator, cumsum) -> float? That would remove the duplication entirely rather than just keeping four copies of a comment in sync, and it'd prevent a repeat of exactly the kind of drift this PR is fixing (see also #2977, which hit the same "two near-identical implementations silently diverged" pattern for context recall).

Also +1 on adding a property-based test (e.g. via hypothesis) asserting monotonicity of average precision over random verdict rankings, in addition to the fixed ==1.0/==0.0 cases — would catch regressions in the general case, not just the two boundary values.

PR #2956, opened 2026-08-16
fix(llms): make VertexAI imports optional for langchain-community >= 0.4

sanyaden:
## Problem

A fresh `pip install ragas` today resolves `langchain-community` to the 0.4.x line (the release generation that accompanies langchain 1.x), where the legacy VertexAI integrations were removed. As a result, **`import ragas` crashes immediately**:

```
File "ragas/llms/base.py", line 12, in <module>
    from langchain_community.chat_models.vertexai import ChatVertexAI
ModuleNotFoundError: No module named 'langchain_community.chat_models.vertexai'
```

Reproduce: `pip install ragas==0.4.3 'langchain-community>=0.4'` → `python -c "import ragas"`.

## Root cause

`ChatVertexAI` / `VertexAI` are imported at module top level in `src/ragas/llms/base.py`, but they are only used for the `isinstance` capability check in `MULTIPLE_COMPLETION_SUPPORTED` / `is_multiple_completion_supported()`. `pyproject.toml` declares `langchain-community` unpinned, so new installs pull 0.4.x and the whole import chain (`ragas → evaluation → llms → base`) dies.

## Fix

Make the VertexAI imports optional, preferring the dedicated package these classes actually live in nowadays:

1. try `langchain_google_vertexai` (current home of both classes);
2. fall back to the legacy `langchain_community` locations (still present in 0.3.x);
3. if neither is available, skip the VertexAI entries in the capability list — the check simply returns `False` for those models, which matches the behaviour for any other non-listed LLM.

No behaviour change for existing users of langchain-community 0.3.x; users on `langchain-google-vertexai` now get correct n-completion detection as a bonus.

## Testing

- Repro venv (`ragas==0.4.3` + `langchain-community==0.4.2`): `import ragas` fails before, works after installing this branch; `is_multiple_completion_supported(ChatOpenAI(...))` still returns `True`.
- Verified the legacy import path still resolves on `langchain-community==0.3.30`.
- `pytest tests/unit/llms tests/unit/test_import.py`: **60 passed** with this branch on the langchain 1.x stack (on clean main the same selection can't even collect — it dies on the vertexai import).
- `ruff check` / `ruff format --check` clean on the touched file.


linhongyu510 (approved):
Independent verification on head 96285f2: approved with no blocking findings. I exercised the import/capability contract in three isolated dependency matrices: (1) langchain-community 0.4.2 without langchain-google-vertexai: import ragas succeeds and the Vertex types are correctly absent; (2) langchain-community 0.4.2 plus langchain-google-vertexai 3.2.3: both ChatVertexAI and VertexAI are retained in MULTIPLE_COMPLETION_SUPPORTED; (3) legacy langchain-community 0.3.31: both legacy Vertex types are retained. On the modern stack, tests/unit/llms plus tests/unit/test_import.py pass 60/60; Ruff check, Ruff format --check, and git diff --check also pass. The separate legacy imports in opt-in integration modules are a valid follow-up but are outside the default import chain and do not block this focused fix. AI assistance is disclosed in the PR body.

PR #2977, opened 2026-08-28
fix(metrics): apply NonLLM context threshold consistently (>= in recall)

### LINKED ISSUE #2777: NonLLMContextRecall and NonLLMContextPrecisionWithReference threshold relevance inconsistently (> vs >=)

linhongyu510:
## Summary

Fixes #2777.

`NonLLMContextRecall` and `NonLLMContextPrecisionWithReference` both turn a per-context string-similarity score into a binary relevance judgement using a shared `threshold` (default `0.5`) and the same `NonLLMStringSimilarity` distance measure, but they applied the threshold with **different** comparison operators:

- `NonLLMContextRecall._compute_score` used strict `>`:
  ```python
  ```
- `NonLLMContextPrecisionWithReference` used `>=`:
  ```python
  ```

As a result, a context whose similarity is **exactly** at the threshold was counted as relevant by precision but not by recall — an inconsistency between two metrics that are meant to be used together.

## Fix

Align recall to `>=` so a boundary score is treated as relevant by both metrics. This is the more intuitive "meets the threshold" semantics and matches the existing precision behavior.

Behavior for scores strictly above/below the threshold is unchanged.

## Tests

Added `tests/unit/test_nonllm_context_threshold.py` (pure logic, no LLM/network):
- a score exactly at the threshold is now relevant for recall;
- a score below the threshold stays irrelevant;
- recall and precision agree on a boundary score;
- an empty verdict list still returns `nan`.

```
$ python -m pytest tests/unit/test_nonllm_context_threshold.py -q
4 passed
```

Before the fix, `NonLLMContextRecall(threshold=0.5)._compute_score([0.5])` returned `0.0`; it now returns `1.0`, matching precision.


src/ragas/metrics/_context_recall.py
```diff
@@ -216,7 +216,12 @@ async def _ascore(self, row: t.Dict, callbacks: Callbacks) -> float:
         return await self._single_turn_ascore(SingleTurnSample(**row), callbacks)
 
     def _compute_score(self, verdict_list: t.List[float]) -> float:
-        response = [1 if score > self.threshold else 0 for score in verdict_list]
+        # Use `>=` so a context whose similarity is exactly at the threshold
+        # is treated as relevant, matching NonLLMContextPrecisionWithReference
+        # (which also uses `>=`). Both metrics share the same default
+        # threshold (0.5) and NonLLMStringSimilarity distance measure, so they
+        # must apply the threshold consistently (see issue #2777).
+        response = [1 if score >= self.threshold else 0 for score in verdict_list]
```
rajeshl8, line 224: Good catch — this exact "two independently-implemented metrics silently disagree at the boundary" pattern is also what's being fixed in #2938/#2959/#2953 for ContextPrecision's epsilon handling. Given that pattern is recurring, would it make sense to centralize the threshold comparison (e.g. a small meets_threshold(score, threshold) util) so any future LLM-based or string-similarity-based variant reuses the same >= semantics instead of re-implementing it ad hoc? Right now nothing enforces that NonLLMContextRecall and NonLLMContextPrecisionWithReference stay in sync going forward, only that they happen to match after this fix.

linhongyu510: Agreed, and implemented in `a369f92` — you were pointing at the part that actually matters. Changing recall to `>=` fixed today's mismatch but left nothing preventing the next one; "they happen to match after this fix" was exactly the right description of the state I left it in.

`meets_threshold(score, threshold)` now lives in `ragas/metrics/utils.py` (alongside `fbeta_score`) as the single definition, and **both** `NonLLMContextRecall` and `NonLLMContextPrecisionWithReference` route through it. Precision is included even though it was already correct — leaving it with its own inline `>=` would have kept two implementations and defeated the point.

The interesting part was making the guard test the invariant rather than the outputs. Asserting both metrics return the same thing at the boundary is what the existing tests already do, and that passes right up until someone edits one of them. So there are three:

1. **Identity** — both modules bind the same function object, so neither can be holding a private copy.
2. **Late binding** — patching the rule changes recall's result, proving it reads the shared definition at call time rather than having inlined it.
3. **Structural** — neither module contains a literal `> self.threshold`.

That third one is the one that catches the regression class you described. Reverting recall to an inline `>` fails it **immediately**, before the behavioural tests even matter:

```
FAILED ... test_recall_counts_score_at_threshold_as_relevant
FAILED ... test_recall_and_precision_agree_at_threshold
FAILED ... test_no_metric_reimplements_the_boundary
3 failed, 5 passed
```

One implementation detail worth recording: the identity check has to read `sys.modules["ragas.metrics._context_recall"]`, because `ragas.metrics._context_recall` resolves to the re-exported *class*, not the module — my first attempt failed with `'ContextRecall' object has no attribute 'meets_threshold'`.

Verification: `tests/unit/test_nonllm_context_threshold.py` **8 passed**; `ruff check` across `src/ragas/metrics/` and `ruff format --check` clean.

On #2938/#2959/#2953 — I deliberately did not touch ContextPrecision's epsilon handling here, since that is a different comparison with its own semantics and those PRs are already in flight. If it would help, `meets_threshold` is a natural place for an epsilon-aware variant to land once that work settles, rather than a fourth ad hoc comparison.

PR #3017, opened 2026-09-18
fix: guard the ChatVertexAI import that broke `import ragas` outright

pcbeingused333:
## What broke

`ragas/llms/base.py` imports `langchain_community.chat_models.vertexai.ChatVertexAI` unconditionally at module load time, only to list it in `MULTIPLE_COMPLETION_SUPPORTED` for an `isinstance()` check:

```python
from langchain_community.chat_models.vertexai import ChatVertexAI
...
MULTIPLE_COMPLETION_SUPPORTED = [OpenAI, ChatOpenAI, AzureOpenAI, AzureChatOpenAI, ChatVertexAI, VertexAI]
```

That submodule has been removed from `langchain-community` — Vertex AI chat support moved to the standalone `langchain-google-vertexai` package (langchain-community now prints its own sunset/deprecation notice on import). `langchain-community` is an unpinned dependency in `pyproject.toml`, so any environment that resolves a current `langchain-community` (0.4.2 as of this PR) hits:

```
ModuleNotFoundError: No module named 'langchain_community.chat_models.vertexai'
```

...the moment anything does `import ragas`. Reproduced with a fresh `make install-minimal`.

## Fix

Guard the import the same way this codebase already handles other optional integrations (see `llms/oci_genai_wrapper.py`'s `try/except` around `import oci`), and filter it out of `MULTIPLE_COMPLETION_SUPPORTED` instead of leaving a bare `None` in the list (which would make the `isinstance()` check raise `TypeError`). `langchain_community.llms.VertexAI` — the other import on that line — is untouched; it's still present in this version.

## Testing

- Verified red: with the import unguarded, `import ragas` fails in this environment, and the new test can't even collect.
- Verified green: `import ragas` succeeds; `is_multiple_completion_supported()` still works for the providers that are actually importable.
- Full `tests/unit` suite: 776 passed, 105 skipped (pre-existing failures in this environment are unrelated missing optional deps — `sacrebleu`, `rapidfuzz`, `llama_index`, `git` — from using `make install-minimal`, not this change).
- `ruff check`/`ruff format --check`/`pyright` clean on the changed files.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3018, opened 2026-09-18
fix: SummarizationScore._compute_qa_score raises ZeroDivisionError on empty answers

pcbeingused333:
## The bug

Both the legacy `SummarizationScore` (`metrics/_summarization.py`) and the v2 `SummaryScore` (`metrics/collections/summary_score/metric.py`) compute the QA sub-score as:

```python
correct = sum([1 for a in answers if a.lower() == "1"])
return correct / len(answers)
```

`answers` can legitimately be empty. `_extract_keyphrases` and `_get_questions`/`_generate_questions` already handle the LLM returning nothing by logging an error and continuing with `[]` — that's existing, deliberate behavior in this same file. An empty question list can just as easily produce an empty answer list from the answer-generation step, and when it does, `_compute_qa_score` raises `ZeroDivisionError` instead of returning a degraded score.

The v2 implementation even documents the bug in a comment rather than fixing it:

```python
def _compute_qa_score(self, answers: List[str]) -> float:
    """Compute QA score as ratio of correct answers. Matches legacy behavior exactly."""
    correct = sum([1 for a in answers if a.lower() == "1"])
    return correct / len(
        answers
    )  # Will raise ZeroDivisionError if answers is empty (legacy behavior)
```

## The fix

Return `0.0` for empty `answers` in both implementations, matching the "log and degrade" convention the surrounding code already uses for empty keyphrases/questions, instead of crashing on an LLM step that produced no signal.

## Testing

Neither implementation had any unit-level test coverage at all — the only existing test (`tests/e2e/metrics_migration/test_summary_score_migration.py`) requires live API keys and is skipped without them, so this path was completely unverified.

Added `tests/unit/test_summarization.py` and `tests/unit/test_summary_score_collections.py`, testing `_compute_qa_score` directly (it needs no LLM — it's pure arithmetic over an already-generated list), including the empty-answers regression case. Verified red against the pre-fix code: both raise `ZeroDivisionError` on `_compute_qa_score([])`.

`ruff check`/`ruff format --check` clean on the changed files.


pcbeingused333:
CI is sitting in \`action_required\` (fork-contributor approval gate) — could someone press "Approve and run workflows"?

Local verification while it's waiting:
\`\`\`
$ .venv/bin/python -m pytest tests/unit/test_summarization.py tests/unit/test_summary_score_collections.py -q
8 passed
$ .venv/bin/python -m pytest tests/unit -q
776 passed, 105 skipped  (pre-existing gaps are unrelated missing optional deps from make install-minimal: sacrebleu, rapidfuzz, llama_index, git)
$ .venv/bin/python -m ruff check src/ragas/metrics/_summarization.py src/ragas/metrics/collections/summary_score/metric.py tests/unit/test_summarization.py tests/unit/test_summary_score_collections.py
All checks passed!
\`\`\`

Same "Claude Docs Check" failure as on #3017 — its \`pull_request_target\` workflow refuses to check out fork PR code, unrelated to this diff.

